PR #6, opened 2020-11-04
defineCustomElement

littlespex:
- Moved custom element definition code into utility function
- Updated gulp build script to use ES6 modules to match the `package.json` configuration.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,16 +1,18 @@
-const { src, dest, watch, series } = require('gulp');
-const livereload = require('gulp-livereload');
-
-const pkg = require('./package.json');
-const del = require('del');
-
+import commonjs from '@rollup/plugin-commonjs';
+import resolve from '@rollup/plugin-node-resolve';
+import del from 'del';
+import fs from 'fs';
+import gulp from 'gulp';
 // rollup
-const { rollup } = require('rollup');
-const resolve = require('@rollup/plugin-node-resolve');
-const commonjs = require('@rollup/plugin-commonjs');
-const { terser } = require("rollup-plugin-terser");
+import * as rollupPkg from 'rollup';
+import rollupPluginTerser from 'rollup-plugin-terser';
 
-async function build() {
+const { series, watch } = gulp;
+const pkg = JSON.parse(fs.readFileSync('./package.json'));
+const { rollup } = rollupPkg;
+const { terser } = rollupPluginTerser;
+
+export async function build() {
   await del('dist');
 
   const bundle = await rollup({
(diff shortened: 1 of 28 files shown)


gulpfile.js
```diff
@@ -1,16 +1,18 @@
...
-const livereload = require('gulp-livereload');
-
-const pkg = require('./package.json');
-const del = require('del');
-
+import commonjs from '@rollup/plugin-commonjs';
+import resolve from '@rollup/plugin-node-resolve';
+import del from 'del';
+import fs from 'fs';
+import gulp from 'gulp';
```
littlespex, line 5: The `package.json` file has the `"type": "module"` flag set, so I ran into compile errors when trying to run the `build` and `dev` scripts. I converted the file to ES6 style imports. Alternatively, I could revert these changes and remove `"type": "module"` from the package file.

src/js/media-chrome-button.js
```diff
@@ -87,16 +88,13 @@ class MediaChromeButton extends MediaChromeElement {
...
   set icon(svg) {
     this.iconContainer.innerHTML = svg;
   }
 }
 
-if (!window.customElements.get('media-chrome-button')) {
-  window.customElements.define('media-chrome-button', MediaChromeButton);
-  window.MediaChromeButton = MediaChromeButton;
-}
+defineCustomElement('media-chrome-button', MediaChromeButton);
```
littlespex, line 98: This is the bulk of the changes, converting this chunk into a util function in every component file.

src/js/media-chrome-menu-button.js
```diff
@@ -62,12 +63,9 @@ class MediaChromeMenuButton extends MediaChromeButton {
     }
   }
 
-  mediaSetCallback(media) {}
+  mediaSetCallback(media) { }
 }
 
-if (!window.customElements.get('media-chrome-menu-button')) {
-  window.customElements.define('media-chrome-menu-button', MediaChromeMenuButton);
-  window.MediaChrome = MediaChromeMenuButton;
```
littlespex, line 70: This line is one of the lines that initiated the conversion. There are a few components that register `window.{COMPONENT_NAME}` incorrectly.

src/js/utils/defineCustomElement.js
```diff
@@ -0,0 +1,6 @@
+export function defineCustomElement(name, element) {
+  if (!window.customElements.get(name)) {
+    window.customElements.define(name, element);
+    window[element.name] = element;
+  }
+}
```
littlespex, line 6: Here is the util function.

src/js/media-chrome-menu-button.js
heff: Ha! Oops.

gulpfile.js
heff: Ah nice. I'm not running into that for some reason but seems like the right change.

heff (approved):
Good stuff. Thanks Casey!

Issue #3, opened 2020-04-05
didn't work for me :/


tomByrer:
After having to remove the `"type": "module",` line from `package.json` [thanks to Gulp](https://stackoverflow.com/questions/60209196/gulp-throws-error-must-use-import-to-load-es-module), then firing up lite-server, I could not get any of the example files to work but the "youtube-controls".  Some didn't show any sign of a player at all, & this seems odd

![image](https://user-images.githubusercontent.com/1308419/78512855-78337400-776d-11ea-96fd-071e3b15ffc1.png)

Win10, Firefox & Chromium based browser.

heff:
Thanks for giving it a shot! Could you help me understand how you're using gulp where this happens?

tomByrer:
`yarn build` & `yarn dev` won't run unless I remove the `"type": "module",`.

dylanjha:
I think I ran into this same issue:

![nextjs-media-chrome_2020-11-12_13-26-01](https://user-images.githubusercontent.com/764988/98998231-a985d680-24ea-11eb-8c3e-db7e615943ef.png)

heff:
Switched gulp over to use es6 imports, which I think should fix this issue. @dylanjha want to try again? 🙏

heff:
I think the issue is fixed. Please confirm if you can. Re-open if not.

PR #10, opened 2020-11-15
Switch to using just esbuild

mmcc:
Also moves to outputting a module for each file for future importing purposes.


heff:
Was writing this up in another issue but probably better to keep the convo here.

---

We need to figure out the right build script and distribution options for this project. Feel free to weigh in. Related to #3 .

### Goals

Load all the elements at once in a `<script>`
```
<script src="https://unpkg.com/media-chrome">
```
Load single elements individually in a `<script>`
```
<script src="https://unpkg.com/media-chrome/media-play-button">
or
<script src="https://unpkg.com/media-play-button">
```
Load elements in a project using npm
```
npm install media-chrome
```
```
import 'media-chrome'; // auto-registers all custom elements in the DOM
import { MediaPlayButton } from 'media-chrome'; // returns a custom element for extension
```

The project is currently using gulp + rollup to generate a single file that is published.

@mmcc's #10 switches to just rollup and creates individual files for each component file. Those files continue to have `imports` in them that point to other files in the same directory. Does that work in modern tools? 

@coreymux thoughts?

I hate build scripts...

heff:
Consider ESBuild for this? https://github.com/evanw/esbuild

src/js/media-fullscreen-button.js
```diff
@@ -1,19 +1,8 @@
...
-</svg>
-`;
-
-const exitFullscreenIcon = `
-<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24">
-  <path d="M0 0h24v24H0z" fill="none"/>
-  <path class="icon" d="M5 16h3v3h2v-5H5v2zm3-8H5v2h5V5H8v3zm6 11h2v-3h3v-2h-5v5zm2-11V5h-2v5h5V8h-3z"/>
-</svg>
-`;
+import enterFullscreenIcon from '../svgs/fullscreen.svg';
```
mmcc, line 4: I didn't go through and do these for every SVG, wanted to see how you felt about it first.

heff: Does this work without the build script or is that specially supported by esbuild?

heff:
Recapping what we talked about IRL

Targets:
* CDN Hosted
  1. Bundle and minify the full project as the default
  2. Host each element individually (and minify if possible), maintaining module syntax and shared dependencies
* npm
  * Use the same bundled and minified version as the CDN. Rely on tree shaking for handling unused components

build.js
```diff
@@ -0,0 +1,52 @@
...
+async function build(opts = {}) {
+  const start = Date.now();
+
+  await esbuild.build({
+    entryPoints: ['./index.js'],
+    bundle: true,
+    outfile: './dist/index.js',
+  });
+
+  let entryPoints = await glob('./src/js/media-*.js');
```
heff, line 20: We have to export the utils dir too, otherwise all these break.

index.js
```diff
@@ -1,24 +1,53 @@
...
-import './src/js/media-pip-button.js';
-
-import './src/js/media-title-element.js';
-
-import './src/js/media-control-bar.js';
-
-// Main <media-chrome> element
-import './src/js/media-chrome-container.js';
+import MediaChromeButton from './src/js/media-chrome-button.js';
+import MediaChromeContainer from './src/js/media-chrome-container.js';
```
heff, line 2: I changed the class name of MediaChrome to MediaChromeContainer to disambiguate it from the main project, however I left the tag name itself `<media-chrome>`. However with this "load each element" approach, it might throw people off who are trying to load the media-chrome tag. Interested in your thoughts.

build.js
```diff
@@ -0,0 +1,52 @@
...
+
+const args = process.argv;
+const PORT = 3001;
+
+async function build(opts = {}) {
+  const start = Date.now();
+
+  await esbuild.build({
+    entryPoints: ['./index.js'],
+    bundle: true,
```
heff, line 16: Can we minify here too?

heff (commented):
The util thing is the only actual issue, otherwise this looks great! 👍

index.js
mmcc: I think it'd be a bit confusing to not have something match the name. Would it be crazy to just have `<media-chrome>` as an alias for `<media-chrome-container>`?

build.js
mmcc: Done!

index.js
heff: I tried that and it doesn't let you use the same class for two different tag names. A hack might be to subclass the class and use that to register a second tag name.

I like the discoverability that using <media-chrome> as a tag name adds, but think that could also be confusing with the parent project. Should it just be `@media-chrome/media-container`? Probably overthinking this.

Either way, I think this is good to merge without solving this issue.

build.js
```diff
@@ -0,0 +1,58 @@
...
+    outfile: './dist/index.js',
+    ...opts,
+  });
+
+  /*
+  For all the other files, we'll just export them to dist so folks can import the modules individually if they want.
+  */
+  let entryPoints = await glob('./src/js/**/*.js');
```
heff, line 27: Will this create a `dist/utils/` dir or just pull all js files up into the dist folder?

mmcc: This creates a `dist/utils` directory.

index.js
mmcc: I think `media-container` feels pretty good imo.

heff: Yeah, cool. Let's do it. It's a breaking change, but we can do the subclass hack while we deprecate the `<media-chrome>` version. But not needed in this PR.

build.js
heff: Awesome. Is this ready to merge then?

mmcc: I think so! This should probably come with a decent version bump too, it _shouldn't_ change much for existing users, but it could. Once it's in and published to npm/unpkg I'll put together some examples.

build.js
heff: This doesn't appear to be minifying for me. What changed?

heff: Yeah, that was it. At some point we're going to have to figure out how to minify the templates too (not today).

PR #11, opened 2020-11-17
Turn up volume if zero when unmuted

heff:
Currently when the video/audio starts muted with volume zero, and then you click the _unmute_ button, nothing appears to change. It is actually toggling mute/unmute, but because the volume is still zero, the icon doesn't change.

This update bumps the volume to 0.25 when unmuting the volume is zero.

### DIFF
--- a/src/js/media-mute-button.js
+++ b/src/js/media-mute-button.js
@@ -19,6 +19,10 @@ class MediaMuteButton extends MediaChromeButton {
   onClick(e) {
     const media = this.media;
     media.muted = !media.muted;
+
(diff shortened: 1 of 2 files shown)


andzhan (commented):
This change will still cause playback to start in a muted state if no previous volume level exists, correct? Is it possible to make the default starting volume a non-zero value?

heff:
Ah, I found the bug. Just updated the PR to fix it. Not sure if some browsers will still default to muted naturally, but this removes an issue where media chrome would set the video volume to zero.

andzhan (approved):

PR #12, opened 2020-11-18
issue/1

### LINKED ISSUE #1: Support Safari's fullscreen API
Really thought they would have caught up on fullscreen API naming by now...

`el.requestFullscreen` does not work. Safari has `el.webkitRequestFullScreen`.

Which also means the other fullscreen related APIs are going to be different still.

littlespex:
Added support for legacy fullscreen APIs

Resolves #1

### DIFF
--- a/src/js/media-chrome.js
+++ b/src/js/media-chrome.js
@@ -9,13 +9,18 @@ template.innerHTML = `
   <style>
     :host {
       box-sizing: border-box;
-
       position: relative;
       display: flex;
       width: 720px;
       height: 480px;
       background-color: #000;
       flex-direction: column-reverse;
+    }
+
+    :host(:-webkit-full-screen) {
+      /* Needs to use !important otherwise easy to break */
+      width: 100% !important;
+      height: 100% !important;
     }
 
     ::slotted([slot=media]) {
--- a/src/js/media-fullscreen-button.js
+++ b/src/js/media-fullscreen-button.js
@@ -20,29 +20,68 @@ class MediaFullscreenButton extends MediaChromeButton {
     super();
 
     this.icon = enterFullscreenIcon;
+  }
 
-    this.addEventListener("click", e => {
-      const media = this.media;
+  onClick() {
+    const api = this.getFullscreenApi(this.mediaChrome);
 
-      if (this.mediaChrome == document.fullscreenElement) {
-        document.exitFullscreen();
-      } else {
-        if (document.pictureInPictureElement) {
-          // Should be async
-          document.exitPictureInPicture();
-        }
-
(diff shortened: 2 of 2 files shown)


src/js/media-fullscreen-button.js
```diff
@@ -20,30 +20,69 @@ class MediaFullscreenButton extends MediaChromeButton {
...
-      const media = this.media;
-
-      if (this.mediaChrome == document.fullscreenElement) {
-        document.exitFullscreen();
-      } else {
-        if (document.pictureInPictureElement) {
-          // Should be async
-          document.exitPictureInPicture();
-        }
+  onClick() {
```
littlespex, line 25: Moved the button click functionality to the `onClick` method provided by the base class.

src/js/media-fullscreen-button.js
```diff
@@ -20,30 +20,69 @@ class MediaFullscreenButton extends MediaChromeButton {
...
-      if (this.mediaChrome == document.fullscreenElement) {
+      api.enter();
+    }
+  }
+
+  mediaSetCallback() {
+    const api = this.getFullscreenApi(this.mediaChrome);
+    
+    document.addEventListener(api.event, () => {
+      if (this.mediaChrome == document[api.element]) {
```
littlespex, line 44: The `getFullscreenApi` method needs a reference to the `media-chrome` element, so the fullscreen event listener has been moved to the `mediaSetCallback`

src/js/media-fullscreen-button.js
```diff
@@ -20,30 +20,69 @@ class MediaFullscreenButton extends MediaChromeButton {
...
+    else if (container.msRequestFullscreen != null) {
+      api.enter = container.msRequestFullscreen.bind(container);
+      api.exit = document.msExitFullscreen.bind(document);
+      api.event = "msfullscreenchange";
+      api.element = "msFullscreenElement";
+      api.error = "MSFullscreenError";
+    }
+
+    return api;
+  }
```
littlespex, line 85: Hooray for standards...

heff (approved):
Added a fix for this issue

image

src/js/media-chrome.js
```diff
@@ -18,6 +17,12 @@ template.innerHTML = `
       flex-direction: column-reverse;
     }
 
+    :host(:-webkit-full-screen) {
```
heff, line 20: This was also needed for Safari...

src/js/media-fullscreen-button.js
littlespex: If the baseline is "browsers that support custom elements" I think we could remove the MS section as well.

PR #14, opened 2020-11-20
Cleanup the fullscreen api code

littlespex:


src/js/media-fullscreen-button.js
```diff
@@ -15,31 +15,28 @@ const exitFullscreenIcon = `
...
+};
+
+if (!document.fullscreenElement) {
+  api.enter = "webkitRequestFullScreen";
+  api.exit = document.webkitExitFullscreen != null ? "webkitExitFullscreen" : "webkitCancelFullScreen";
+  api.event = "webkitfullscreenchange";
+  api.element = "webkitFullscreenElement";
+  api.error = "webkitfullscreenerror";
+}
+
```
littlespex, line 33: - Moved the api mapping out of the function as it only needs to be run once.
- Removed `MS` and `moz` cases as they don't apply to the list of minimum supported browser versions.

src/js/media-fullscreen-button.js
```diff
@@ -15,31 +15,28 @@ const exitFullscreenIcon = `
...
-      api.enter();
-    }
-  }
 
-  mediaSetCallback() {
-    const api = this.getFullscreenApi(this.mediaChrome);
-    
     document.addEventListener(api.event, () => {
       if (this.mediaChrome == document[api.element]) {
         this.icon = exitFullscreenIcon;
```
littlespex, line 42: Now that the api mapping is static, the event listener can be added in the constructor

src/js/media-fullscreen-button.js
heff: Nice, thanks for the cleanup pass.

src/js/media-fullscreen-button.js
```diff
@@ -15,31 +15,28 @@ const exitFullscreenIcon = `
...
 
+const api = {
+  enter: "requestFullscreen",
+  exit: "exitFullscreen",
+  event: "fullscreenchange",
+  element: "fullscreenElement",
+  error: "fullscreenerror",
+};
+
+if (!document.fullscreenElement) {
```
heff, line 26: This probably needs to be `document.fullscreenElement !== undefined`, otherwise it's usually `null` at this point in Chrome.

littlespex: good catch

littlespex, line 26:
```suggestion
if (document.fullscreenElement !== undefined) {
```

heff (approved):
Nice

PR #19, opened 2020-12-27
Back in the SSR

mmcc:
It felt lame forcing the end-user to get fancy with hooks and things to keep from blowing up in an SSR environment. This adds that defensiveness on our end instead.

This is based off of #10.


mmcc:
hoookay nvm, I was out of date and main has apparently changed quite a bit.

PR #23, opened 2021-03-17
make things SSR friendly

dylanjha:
* created a `browser-env.js` util
* expose `Window` and `Document` constants. In a server env these are empty shims, in a browser env these are the actual browser globals
* expose `isServer()` util that web components can use to selectively skip things on the server
* expose a `createTemplate()` util that doesn't break in a server-env


src/js/utils/defineCustomElement.js
```diff
@@ -1,6 +1,11 @@
+import { isServer, Window } from './browser-env.js';
+
 export function defineCustomElement(name, element) {
-  if (!window.customElements.get(name)) {
-    window.customElements.define(name, element);
-    window[element.name] = element;
+  if (isServer()) {
+    return;
+  }
+  if (!Window.customElements.get(name)) {
```
heff, line 7: We don't really need the Window part here if we've already returned if isServer. Unless you're going for consistency of window usage.

src/js/media-chrome-button.js
```diff
@@ -1,7 +1,9 @@
 import MediaChromeHTMLElement from './media-chrome-html-element.js';
 import { defineCustomElement } from './utils/defineCustomElement.js';
+import { isServer } from './utils/browser-env.js';
+import { createTemplate } from './utils/createTemplate.js';
 
-const template = document.createElement('template');
+const template = createTemplate();
```
heff, line 6: Wondering if we should instead add `createElement` as a mock function on your `Document` shim. My main driver of that is trying to avoid abstracting enough things into functions that this becomes yet another framework rather than "just using HTML/JS". The hole I see myself digging though is having to add a mock function for any window/document function used in the future...

src/js/media-progress-range.js
```diff
@@ -178,11 +180,11 @@ class MediaProgressRange extends MediaChromeRange {
         this.thumbnailPreview.style.left = `${thumbnailLeft}px`;
         this.thumbnailPreview.time = mousePercent * this.media.duration;
       };
-      window.addEventListener('mousemove', mouseMoveHandler, false);
+      Window.addEventListener('mousemove', mouseMoveHandler, false);
```
heff, line 183: If this did actually get called it would error because addEventListener isn't defined on the mock Window, but we're ok because UI events won't get triggered on the server. But then it's another place to question if we need the mock Window just for consistency.

heff (approved):
Good stuff, thanks! I'm approving this if we want to just get this in and done and evolve the approach from here (let me know), but going to throw in some small comments and one big question:

I'm really wishing we could just `if (isServer) return;` at the top level of a module file. Is there a higher level in the project or somewhere in the build process we could do that? Are there reasons why that still wouldn't work, like expected exports from files that would break? Don't stress too hard on that one if there's not a clear opportunity.

src/js/utils/browser-env.js
```diff
@@ -0,0 +1,11 @@
+const windowShim = {};
+const documentShim = {};
+class HTMLElementShim { };
+
+export function isServer () {
```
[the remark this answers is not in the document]
dylanjha: It _can_ be a bool. You're right. That's a bit cleaner

src/js/utils/defineCustomElement.js
dylanjha: >  you're going for consistency of window usage.

The idea was to go for consistency using `Window` everywhere

src/js/media-chrome-button.js
dylanjha: > My main driver of that is trying to avoid abstracting enough things into functions that this becomes yet another framework rather than "just using HTML/JS". The hole I see myself digging though is having to add a mock function for any window/document function used in the future...

Exactly my thoughts. I don't want to build out a whole advanced mock document & window.

The alternative would be updating each file that uses `createTemplate` to look more like this:

```
import { Document, isServer } from './utils/browser-env';

const template = isServer ? Document.createElement('template') : {};

template.innerHTML = ``;
```

It feels bit messy to have to throw in those ternary operators everywhere.

Options:

1. keep `utils/createElement`
2. add `Document.createElement` to our shim, but with the understanding that we're not going to build out a complex shim for every window/document function
3. remove `utils/createElement`, but add conditionals to check that we're not in a server whenever we want to `createElement('template')` (which we do in every file)

I want to avoid (3) but I don't feel too strongly between (1) and (2). Thoughts?

src/js/media-progress-range.js
dylanjha: for consistency, yes

dylanjha:
> I'm really wishing we could just if (isServer) return; at the top level of a module file.
> expected exports from files that would break? Don't stress too hard on that one if there's not a clear opportunity.

I think it comes down to the expected exports. If the module on the server exports `{}` and on the browser it exports: `{ MediaChromeButton, MediaContainer, etc, etc }` I think that could be a problem.

Would have to test it out though and play around with it, in practice since importing the library registers web components, you don't actually need to do anything with the exported object in your code.

src/js/media-chrome-button.js
heff: Yeah, depends on what we're prioritizing most.
* DRY Code -> # 1 `utils/createTemplate`
* Be clear what's happening, help people create their own components -> # 3 add conditionals
* Maybe somewhere in between -> # 2 Document.createElement

I lean towards 3. Also because it makes it really easy to remove when server-side renderers get their act together and support web components.

I might call it something like `./utils/server-safe-globals`, `import { document } from './utils/server-safe-globals'; `

> Exactly my thoughts. I don't want to build out a whole advanced mock document & window.

At the same time, it's not any more work than creating new function wrappers for everything that uses a window/doc property.

heff: Let me know if you want to do this, or not, or if you want me to do it, and we'll get this pulled in.

heff (approved):
High-five ✋

PR #25, opened 2021-03-19
Introduce media-range-selector component

dylanjha:
Note: this PR depends on #23 

The idea here is to create a component that can be used in order to select a range of content from a media object, like this:

range_2021-03-18_23-05-22

This would be really handy to use with the [Mux Asset Clipping API](https://docs.mux.com/guides/video/create-clips-from-your-videos), which requires start and end timestamps from a Mux Asset.

For a full video of the functionality, see here:

https://stream.new/v/32xPDk24hFr3j7VG0001M011lP01V007JJiO3YNngBXQzqUk

This PR is not ready yet, but opening it up for early review.

TODO:

- [x] Rename to `media-range-selector`. Currently the name is `media-trimmer`, but I think `media-range-selector` is better (s/o @mmcc) This component itself does not do the actually "trimming" or "clipping" - it simply allows the selection and emits events when the selection is updated.
- [x] the default styles can be cleaned up, and also make sure the styles are easily over-ridable for customization
- [ ] When moving the start handle, the end handle seems to shift slightly to the right, this is more noticeable on mobile
- [x] QA with shorter video (like 30s)
- [x] QA with a longer video (like an hour or two?)
- [x] QA with an HLS video
- [ ] when trying to move the start handle too far right or the end handle too far left, things don't really work right

Notes:

* This is the first `media-chrome` component that emits events externally. Officially, the API that a developer would use looks like this:

```js
const mediaRangeSelect = document.querySelector('media-range-select');

mediaRangeSelect.addEventListener('update', (evt) => {
  const { startTime, endTime } = evt.detail;
  console.log('range selection change:', startTime, endTime);
});
```

The idea is that a developer would take these `startTime` and `endTime` values from the `update` event and use them in order to clip a Mux Asset, or do whatever they need to with the selected range.


heff:
Gonna need to rebase

src/js/media-trimmer.js
```diff
@@ -0,0 +1,493 @@
...
+    this._clickHandler = this.handleClick.bind(this);
+    this._dragStart = this.dragStart.bind(this);
+    this._dragEnd = this.dragEnd.bind(this);
+    this._drag = this.drag.bind(this);
+    /*
+     * TODO - teardown these handlers later
+     */
+    this.wrapper.addEventListener('click', this._clickHandler, false);
+
+    this.wrapper.addEventListener('touchstart', this._dragStart, false);
```
heff, line 169: You might [not need to handle touch events also](https://developer.mozilla.org/en-US/docs/Web/API/Touch_events/Supporting_both_TouchEvent_and_MouseEvent)? Either that or we need to be careful of double firing handlers.

src/js/media-trimmer.js
```diff
@@ -0,0 +1,493 @@
...
+    }
+    if (this.draggingEl === this.endHandle) {
+      this.initialX = this.getXPositionFromMouse(evt);
+      this.selection.style.width = `${selectionW + xDelta}px`;
+    }
+    this.dispatchUpdated();
+  }
+
+  dispatchUpdated () {
+    const updatedEvent = new CustomEvent('updated', {
```
heff, line 290: I suggest `update` - event names are usually present tense

src/js/media-trimmer.js
```diff
@@ -0,0 +1,493 @@
...
+      return colorsArray;
+    }
+
+    const buffered = media.buffered;
+    const buffPercent = (buffered.end(buffered.length-1) / media.duration) * 100;
+    colorsArray.splice(1, 0, ['var(--media-progress-buffered-color, #777)', buffPercent]);
+    return colorsArray;
+  }
+
+  enableThumbnails() {
```
heff, line 429: In the future we might be able to break this out into a shared add on.

heff (commented):
A few comments inline. Some of the more complex areas we could try to make easier to understand, but in general this is a great direction.

src/js/media-trimmer.js
```diff
@@ -0,0 +1,493 @@
...
+     * TODO - teardown these handlers later
+     */
+    this.wrapper.addEventListener('click', this._clickHandler, false);
+
+    this.wrapper.addEventListener('touchstart', this._dragStart, false);
+    this.wrapper.addEventListener('touchend', this._dragEnd, false);
+    this.wrapper.addEventListener('touchmove', this._drag, false);
+
+    this.wrapper.addEventListener('mousedown', this._dragStart, false);
+    this.wrapper.addEventListener('mouseup', this._dragEnd, false);
```
heff, line 174: Probably need to listen to window mouseup and mousemove. If you mouseup off the component today it breaks stuff.

src/js/media-trimmer.js
dylanjha: agreed

src/js/media-trimmer.js
dylanjha: interestingly:

* on desktop: `mousedown` fires the callback as expected (when clicking *down*)
* on mobile: `mousedown` waits to fire the callback at the end (touch down, then lift)

You're totally right about double-firing callbacks, let me dig into this and figure out the right solution

heff: If it's helpful, here's the same type of functionality in Video.js. Looks like we did end up handling touchstart, but couldn't tell you if that's still the right answer.
https://github.com/videojs/video.js/blob/main/src/js/slider/slider.js#L60

dylanjha: yep, that was helpful

examples/range-selector.html
```diff
@@ -0,0 +1,56 @@
...
+  <meta name="viewport" content="width=device-width">
+  <script type="module" src="../dist/index.js"></script>
+</head>
+<style>
+  /*
+     over-ride the opacity rule so that the media-control-bar is always
+     showing (the default behavior is to hide it when media is playing)
+   */
+  media-control-bar {
+    opacity: 1;
```
heff, line 12: I've thought about a `<control-bar autohide="numberOfSeconds">`. Looks like that might be useful here if we had an option like `-1` to not hide.

src/js/media-control-bar.js
```diff
@@ -31,7 +31,8 @@ template.innerHTML = `
     }
 
     media-progress-range,
-    ::slotted(media-progress-range) {
+    ::slotted(media-progress-range),
```
heff, line 34: We might want to create a formal class here, like ".media-timeline", to make it easier to make an element expand to full width. Not important for this PR.

heff (approved):
The mousemove comment it the main one that I think is worth fixing before merging this, but otherwise this feels pretty good.

dylanjha:
Renamed to `media-clip-selector`. @heff and I like this name

@heff please have a look 👀 there is _a little_ wonkiness where the end handle can jump a few pixels at times, but I think that can be ironed out over time. I'm keen on getting this in folks' hands so we can see if its useful and gather feedback

examples/clip-selector-with-hls.html
```diff
@@ -0,0 +1,68 @@
...
+     showing (the default behavior is to hide it when media is playing)
+   */
+  media-control-bar {
+    opacity: 1;
+  }
+</style>
+<body>
+
+<media-container>
+  <video
```
heff, line 19: Just want to make sure you're aware of the [hls web component](https://www.npmjs.com/package/hls-video-element).

Does the clip selector work differently with HLS than an MP4 some how?

src/js/media-container.js
```diff
@@ -24,7 +24,8 @@ template.innerHTML = `
       flex-direction: column-reverse;
 
       /* Default dimensions */
-      width: 720px;
+      width: 100%;
+      max-width: 720px;
```
heff, line 28: I'm a little worried about the effects of this on bigger videos, but we can leave it in and see.

heff (approved):

PR #26, opened 2021-03-26
Added config attr for autohiding controls

heff:
Configure how many seconds it takes to auto-hide the controls and other UI elements

Default
`<media-container autohide="2">`

No time
`<media-container autohide="0">`

Never hide
`<media-container autohide="-1">`


dylanjha:
`autohide="3"`, `autohide="4"` would be 4 seconds, right?

heff:
Yeah, exactly. Sorry for the cryptic description. :)

dylanjha (approved):
nice touch!

PR #30, opened 2021-04-11
POC Media Controller - re-architecture to use events and media state attrs

heff:
### Major Change Summary
* `<media-container>` becomes `<media-controller>`
* Control elements emit pre-defined events (e.g. `mediaplayrequest`) to trigger media actions. These are caught and handled by the media-controller, calling the desired action on the media element.
* Control elements receive state via `media${stateName}` props, set by the media-controller, rather than hooking into the media element directly. e.g. `<media-play-button media-paused>` and `mediaPlayButton.mediaPaused`

This sets up an MVC relationship, rather than having every UI component directly connect to the 'model'. (Thanks @mihar-22!) 

Control elements can live outside of media-controller, by setting the media-controller attribute with the ID of the controller or by setting the mediaController property with the media controller element. `<media-play-button media-controller="myMediaController">`

See `examples/media-controller.html`.

Todo:
- [x] Make unpkg links version specific, at least major version.
- [x] Alias media-container to media-controller for v1, maybe change in v2


src/js/media-controller.js
```diff
@@ -0,0 +1,309 @@
...
+  </style>
+  <slot name="media"></slot>
+  <div id="container">
+    <slot></slot>
+  </div>
+`;
+
+/*
+  Loop through child nodes and set the media[State] on every child.
+*/
```
heff, line 70: This is what handles passing state down, in general. Would probably be more convenient if it pierces the shadow dom also.

src/js/media-controller.js
```diff
@@ -0,0 +1,309 @@
...
+class MediaController extends window.HTMLElement {
+  constructor() {
+    super();
+
+    // Set up the Shadow DOM
+    const shadow = this.attachShadow({ mode: 'open' });
+    this.shadowRoot.appendChild(template.content.cloneNode(true));
+    this.container = this.shadowRoot.getElementById('container');
+
+    // Watch for child adds/removes and update the media element if necessary
```
heff, line 105: This is mostly carried over from media-container, but handles any changing of the contained media element.

src/js/media-controller.js
```diff
@@ -0,0 +1,309 @@
...
+            });
+          }
+        }
+      }
+    };
+
+    const observer = new MutationObserver(mutationCallback);
+    observer.observe(this, { childList: true, subtree: true });
+
+    // Capture requests from internal control events
```
heff, line 154: Quick POC of handling events that bubble from child elements or dispatched by associated elements.

src/js/media-controller.js
```diff
@@ -0,0 +1,309 @@
...
+    const mediaName = media.nodeName.toLowerCase();
+
+    if (mediaName.includes('-') && !window.customElements.get(mediaName)) {
+      window.customElements.whenDefined(mediaName).then(()=>{
+        this.mediaSetCallback(media);
+      });
+      return;
+    }
+
+    // Listen for state changes and propagate them to children and associated els
```
heff, line 191: POC of handling media element state changes and propagating them.

src/js/media-chrome-html-element.js
```diff
@@ -36,6 +37,11 @@ class MediaChromeHTMLElement extends window.HTMLElement {
       return;
     }
 
+    if (propName == 'mediaController') {
+      this[propName] = document.getElementById(newValue);
+      return;
+    }
```
mihar-22, line 43: If a component is nested inside some other component's shadow DOM this won't work. Haven't a thought of a solution just yet but wanted to point it out for now.

heff: Yeah, that's a good point. I'd guess that'd break in the button/form relationship too.

I updated the state passing function to pierce shadow doms when open, which appears to be the most common use at least. That way as long as the host el is associated with the controller, state should reach shadow dom nodes. 

I also set `composed:true` on the custom events so they should bubble up out of a shadow dom.

That might be enough. Let me know if you still see some holes.

mihar-22:
Loving the addition of the slots! From my perspective it's coming together a lot cleaner, there are obviously some parts of the codebase that can be refactored for the good of humanity but not a super priority. Interested to talk about these changes and how you're thinking about everything next time we jump on a call 😄  I have some thoughts to share with you too.

heff:
Cool 👍  thanks @mihar-22.

> some parts of the codebase that can be refactored 

Let me know if you think any of those areas impacts the API, as I'm working through refactoring everything.

src/js/media-clip-selector.js
```diff
@@ -2,6 +2,7 @@ import MediaChromeHTMLElement from './media-chrome-html-element.js';
 import { defineCustomElement } from './utils/defineCustomElement.js';
```
heff, line 2: @dylanjha you'll want to double check my update to the clip selector.

PR #36, opened 2021-06-24
Update 'isServer' to also check for customElements

KimPaow:
In some environments there are preexisting `window` shims that do not have the `customElements` on the window object. This update will check for those cases.

See: https://github.com/sanity-io/sanity-plugin-mux-input/issues/36#issuecomment-867955123

### DIFF
--- a/src/js/utils/server-safe-globals.js
+++ b/src/js/utils/server-safe-globals.js
@@ -11,8 +11,6 @@ const documentShim = {
   createElement: function() { return {}; }
 };
 
-export const isServer = !!(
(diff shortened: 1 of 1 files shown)


mmcc (approved):

heff (approved):
Seems fine to try and see what happens, but this could break whatever was relying on the other window shim by overriding it with ours, unless the other shim is just a blank object.

If we run into an issue here we could separate shimming the window and shimming custom elements, so that custom elements get added to any existing window shim (or our window shim if needed).

👍

heff:
v0.1.1 - thanks @KimPaow and @dylanjha!

PR #38, opened 2021-07-09
fix files that get published to npm, fix custom element constructor and fix some missing server safe globals

dylanjha:
- [x] currently only dist/index.js get published to npm & unpkg - the syntax in the`files: []` array was incorrectly specifying which files get included:

**before**

```
❯ npx npm-packlist
LICENSE
dist/index.js
package.json
README.md
```

**after** - now the extras and individual elements will be included

```
❯ npx npm-packlist
LICENSE
dist/utils/defineCustomElement.js
dist/utils/fullscreenApi.js
dist/extras/media-clip-selector/index.js
dist/index.js
dist/media-chrome-button.js
dist/media-chrome-html-element.js
dist/media-chrome-range.js
dist/media-container.js
dist/media-control-bar.js
dist/media-controller.js
dist/media-current-time-display.js
dist/media-duration-display.js
dist/media-fullscreen-button.js
dist/media-mute-button.js
dist/media-pip-button.js
dist/media-play-button.js
dist/media-playback-rate-button.js
dist/media-poster.js
dist/media-progress-range.js
dist/media-seek-backward-button.js
dist/media-seek-forward-button.js
dist/media-settings-popup.js
dist/media-text-display.js
dist/media-thumbnail-preview-element.js
dist/media-time-range.js
dist/media-title-element.js
dist/media-ui-events.js
dist/media-volume-range.js
dist/utils/server-safe-globals.js
dist/utils/stringUtils.js
dist/utils/time.js
package.json
README.md
```

- [x] a few places we were using `document` instead of the server safe `document`
- [x] custom element constructor should not manipulate DOM (some browsers are more strict about this than others, react-dom is strict about this)

https://html.spec.whatwg.org/multipage/scripting.html#custom-element-conformance

> In general, work should be deferred to connectedCallback as much as possible—especially work involving fetching resources or rendering. However, note that connectedCallback can be called more than once, so any initialization work that is truly one-time will need a guard to prevent it from running twice.

PR #39, opened 2021-07-10
Basic audio player styling and examples

heff:
* Added an audio example to basic.html

image

* Added a spotify-like theme (more work needed to finish it, but it works)

image

* Added a `<media-time-display>` element to make it easy to show the current time and duration. Can probably replace media-current-time-display.

image

* Separated out video-specific styles in media-controller to allow audio players to be display: inline-flex and not have default height and widths like a video. This might not be the best approach but today you would add an `audio` attribute to media-controller to signal the audio-friendly layout.

* Removed the `#container` from the shadow dom of media-controller. This made it so you couldn't change the media-controller flex box settings, because they were all in the shadow dom instead of :host. 

* Made the `formatTime` function support negative numbers for showing remaining time as a negative.

### DIFF
--- a/examples/themes/netflix-theme.html
+++ b/examples/themes/netflix-theme.html
@@ -61,6 +61,16 @@
     height: 80px;
     padding-bottom: 20px;
     margin: 0 3px;
+
+    --media-button-icon-transform: scale(1.2);
+    --media-button-icon-transition: transform .2s ease;
+  }
+
+  /* For some reason `media-control-bar > *:hover` doesn't work...
+  while `media-control-bar > *:focus-within` does. 
+  And also `media-control-bar > *:not(:hover)`
+  Really annoying. Need to submit a bug.  */
+  media-control-bar > *:not(:hover) {
     --media-button-icon-transform: scale(1);
     --media-button-icon-transition: transform .2s ease;
   }
@@ -82,27 +92,9 @@
     outline: 0;
   }
 
-  /* For some reason `media-control-bar > *:hover` doesn't work...
-  while `media-control-bar > *:focus-within` does. Really annoying.
-  Need to submit a bug. */
-  media-control-bar > *:hover,
-  media-play-button:hover,
-  media-seek-backward-button:hover,
-  media-seek-forward-button:hover,
-  media-mute-button:hover,
-  media-fullscreen-button:hover {
-    --media-button-icon-transform: scale(1.2);
-    --media-button-icon-transition: transform .2s ease;
-  }
-
   media-volume-range {
     width: 100px;
   }
-
-  /* media-play-button:hover
-   {
-    --media-button-icon-transform: scale(1.5);
-  } */
 
   .videoTitle {
     flex-grow: 1;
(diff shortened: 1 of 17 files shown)


examples/themes/netflix-theme.html
```diff
@@ -61,6 +61,16 @@
     height: 80px;
     padding-bottom: 20px;
     margin: 0 3px;
+
+    --media-button-icon-transform: scale(1.2);
+    --media-button-icon-transition: transform .2s ease;
```
dylanjha, line 66: Unrelated to this change: do we currently document all the `--media-*` vars somewhere? It might be time to do that

src/js/media-time-display.js
```diff
@@ -0,0 +1,36 @@
+import MediaTextDisplay from './media-text-display.js';
+import { defineCustomElement } from './utils/defineCustomElement.js';
+import { formatTime } from './utils/time.js';
+import { Document as document } from './utils/server-safe-globals.js';
+// Todo: Use data locals: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleTimeString
+
+class MediaTimeDisplay extends MediaTextDisplay {
+  constructor() {
+    super();
+    this.mediaCurrentTimeSet(0);
```
dylanjha, line 10: Since `mediaCurrentTimeSet` calls `_update` which calls `innerHTML`, I think we will need to move this into the `connectedCallback` for the [reasons specified here](https://github.com/muxinc/media-chrome/pull/38#discussion_r667224184)

src/js/utils/time.js
```diff
@@ -23,5 +31,5 @@ export function formatTime(seconds, guide) {
   // Check if leading zero is need for seconds
   s = s < 10 ? '0' + s : s;
 
-  return h + m + s;
+  return (negative ? '-' : '') + h + m + s;
```
dylanjha, line 34: this diff is begging for test coverage, which is on the short list of things to set up 😄

src/js/media-time-display.js
```diff
@@ -0,0 +1,36 @@
+import MediaTextDisplay from './media-text-display.js';
+import { defineCustomElement } from './utils/defineCustomElement.js';
+import { formatTime } from './utils/time.js';
+import { Document as document } from './utils/server-safe-globals.js';
+// Todo: Use data locals: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleTimeString
+
+class MediaTimeDisplay extends MediaTextDisplay {
```
dylanjha, line 7: need to add `media-time-display` to the README as an available component?

dylanjha (changes requested):
neat example, and I love the `media-time-display` component with the `show-duration` attribute

examples/themes/netflix-theme.html
heff: Yeah, I agree. I was starting to thing through what the right strategy is. These are elements, so I'd love to keep things looking as much like styling a native element, and only use variables in special cases.

The rub being that custom elements don't give you a good way to style many things at once. If I want to change the background of all button controls, I have to name them all specifically, or provide a variable. Or another option would be to force a class name on those elements, but I'm not sure how well that would play with other things.

dylanjha (approved):
LGTM!

PR #42, opened 2021-07-29
Theme architecture proposal

heff:
I was thinking through how we might host and load themes, and I landed on using custom elements to do it. 

Custom elements:
* Have a mechanism for template variables (slots)
* Have a global registry, accessible to the shadow and light doms on the page (customElements.define). This would be hard to recreate otherwise.
* Have a mechanism for knowing when the template is loaded (customElements.whenDefined)

Hoping to get feedback.

I see two ways they might be used.

### 1. Used directly in the HTML

```html
<media-theme-netflix>
  <video slot="media">
</media-theme-netflix>
```

### 2. Access the template string to clone into the shadow dom of a wrapping player

```html
<my-player theme="netflix">
  <video slot="media">
</my-player>
```

Video player constructor
```js

constructor() {
    const shadow = this.attachShadow({ mode: 'open' });

    customElements.whenDefined(themeName).then(()=>{
      const theme = customElements.get(themeName);
      const template = themeEl.template;
      shadow.innerHTML = template;
    });
}

```

### DIFF
--- a/src/js/media-container.js
+++ b/src/js/media-container.js
@@ -143,7 +143,12 @@ class MediaContainer extends MediaChromeHTMLElement {
 
   // First direct child with slot=media, or null
   get media() {
-    return this.querySelector(':scope > [slot=media]');
+    let media = this.querySelector(':scope > [slot=media]');
+
+    // Chaining media slots for media templates
+    if (media.nodeName == 'SLOT') media = media.assignedElements({flatten:true})[0];
+
+    return media;
   }
 
   mediaSetCallback(media) {
--- a/src/js/media-controller.js
+++ b/src/js/media-controller.js
@@ -86,7 +86,10 @@ class MediaController extends MediaContainer {
         super[fullscreenApi.enter]();
       },
       MEDIA_EXIT_FULLSCREEN_REQUEST: () => {
-        this.getRootNode()[fullscreenApi.exit]();
+        document[fullscreenApi.exit]();
+
+        // Shadow root throws an error for this function
+        // this.getRootNode()[fullscreenApi.exit]();
       },
       MEDIA_ENTER_PIP_REQUEST: () => {
         const docOrRoot = this.getRootNode();
@@ -409,7 +412,8 @@ function propagateMediaState(nodeList, stateName, val) {
     }
 
     // Make sure custom els are ready
-    if (childName.includes('-') && !window.customElements.get(childName)) {
+    // Even if customElements.get returns something, it may not be ready!
+    if (childName.includes('-')) {
       window.customElements.whenDefined(childName).then(setAndPropagate);
     } else {
       setAndPropagate();
--- a/themes/example.html
+++ b/themes/example.html
@@ -1,0 +1,31 @@
+<html>
+<head>
+  <script type="module" src="../index.js"></script>
+  <script type="module" src="./media-theme-netflix.js"></script>
+</head>
+<body>
+
+<h2>Example 1: Using the custom element directly</h2>
+
+<media-theme-netflix>
+  <video
+    slot="media"
+    src="https://stream.mux.com/DS00Spx1CV902MCtPj5WknGlR102V5HFkDe/high.mp4"
+  ></video>
+</media-theme-netflix>
+
+<h2>Example 2: Using the template to clone the theme</h2>
+
+<script>
+  customElements.whenDefined('media-theme-netflix').then(()=>{
+    const theme = customElements.get('media-theme-netflix');
+
+    // myElShadow.innerHTML = theme.template;
+  });
+</script>
+
+
+
+
+</body>
+</html>
--- a/themes/example.html
+++ b/themes/example.html
@@ -5,7 +5,7 @@
 </head>
 <body>
 
-<h2>Example 1: Using the custom element directly</h2>
+<h2>Example: Using the custom element directly</h2>
 
 <media-theme-netflix>
   <video
@@ -13,19 +13,6 @@
     src="https://stream.mux.com/DS00Spx1CV902MCtPj5WknGlR102V5HFkDe/high.mp4"
   ></video>
 </media-theme-netflix>
-
-<h2>Example 2: Using the template to clone the theme</h2>
-
-<script>
-  customElements.whenDefined('media-theme-netflix').then(()=>{
-    const theme = customElements.get('media-theme-netflix');
-
-    // myElShadow.innerHTML = theme.template;
-  });
-</script>
-
-
-
 
 </body>
 </html>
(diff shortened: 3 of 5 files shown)


themes/media-theme-netflix.js
```diff
@@ -0,0 +1,164 @@
...
+    display: inline;
+    font-weight: 200;
+    margin-left: 5px;
+  }
+
+</style>
+
+<media-controller>
+
+  <slot name="media" slot="media"></slot>
```
heff, line 128: This is the thing that feels nuckin futs. Chaining slots. But it works!

heff:
@dylanjha @mmcc @cjpillsbury Speak up if you have an opinion on this, otherwise I'm gonna move forward with it.

themes/media-theme-netflix.js
dylanjha: Whoa.

So from the outside, an application does this:

```html
<media-theme-netflix>
  <video slot="media"></video>
</media-theme-netflix>
```

That "media" slot gets passed through into the "media" slot for `media-controller`. Pretty nuckin futs

dylanjha (approved):
I'm in favor of this approach with `<media-theme-xxx>` elements. Easy to understand what's going on and each of the `<media-theme-xxx>` themes provide a great example.

The one question I have is:

* What about in this Netflix example when the theme contains other variables like `videoTitle`? -- Would the `media-theme-netflix` specify other attributes like `<media-theme-netflix video-title="Breaking Bad" video-episode="Episode 2">`

heff:
> What about in this Netflix example when the theme contains other variables like videoTitle

Good call that this needs to be clear as part of this.

Slots are the first answer for me. Themes may have lots of use cases for embedding custom content, and using attributes might feel overloaded. Slots are a custom element's answer to content "variables". And compared to attributes, updates to the slot contents are automatically handled.

```html
<media-theme-abc>
  <video slot="media"></video>
  <span slot="media-title">My great title</span>
</media-theme-abc>
```

A second answer for things like title specifically is potentially defining media metadata elements inside a media element. I think I mentioned this in another issue somewhere.

```html
<media-theme-abc>
  <video slot="media">
    <media-metadata-title>My great title</media-metadata-title>
  </video>
</media-theme-abc>
```

With that, if you swap the video, for example as part of playlist, the metadata comes/goes with it. Doesn't even need to be a real custom element, I think. A downside is handling changes to those elements. Might require yet more mutation observers. But upside is it's compatible with the native audio/video elements, which hide that content by default but still make it available to javascript.

So it might end up being both. For example, the `<media-title>` UI element may have a `media-title="My title"` attribute that can be set by `<media-controller>` (the title being read from the `<media-metadata-title>` element). And the also you can have a slot within the theme that is child content and overrides the attribute value, `<media-title><slot name="media-title"></slot></media-title>` which overrides the attribute value. 

Does that make sense? Any thoughts or ideas on that?

dylanjha:
@heff I love using slots for this.

I'm on board with the idea that any particular "theme" like `<media-theme-netflix>` can define a few of it's own slots. That approach makes a lot of sense.

themes/media-theme-netflix.js
```diff
@@ -0,0 +1,164 @@
+import MediaTheme from './media-theme.js';
```
cjpillsbury, line 1: Fine for now, but there are several inconsistencies with the current browser-based netflix player. Should make followup issue(s) to clean up. Also may want to consider splitting out the part that sizes this to full width from the rest of the theme.

themes/media-theme-netflix.js
```diff
@@ -0,0 +1,164 @@
...
+      <svg slot="exit" viewBox="0 0 28 28"><g transform="translate(3, 6)"><polygon transform="translate(19.000000, 3.000000) scale(-1, 1) translate(-19.000000, -3.000000) " points="22 0 20 0 20 4 16 4 16 6 22 6"></polygon><polygon transform="translate(19.000000, 13.000000) scale(-1, -1) translate(-19.000000, -13.000000) " points="22 10 20 10 20 14 16 14 16 16 22 16"></polygon><polygon points="6 0 4 0 4 4 0 4 0 6 6 6"></polygon><polygon transform="translate(3.000000, 13.000000) scale(1, -1) translate(-3.000000, -13.000000) " points="6 10 4 10 4 14 0 14 0 16 6 16"></polygon></g></svg>
+    </media-fullscreen-button>
+  </media-control-bar>
+</media-controller>
+`;
+
+class MediaThemeNetflix extends MediaTheme {
+  constructor(options={}) {
+    options = Object.assign({
```
cjpillsbury, line 152: nit: no need to mutate `options` (may cause unexpected side effects for implementors). consider instead simply inline object spread, a la:
```js
super(template, { /* allow ...defaultOptions, */ ...options });
```

themes/media-theme.js
```diff
@@ -0,0 +1,31 @@
+
+class MediaTheme extends HTMLElement {
+  constructor(template=``, options={}) {
+    super();
+
+    options = Object.assign({
```
cjpillsbury, line 6: see comment ☝️

src/js/media-container.js
```diff
@@ -143,7 +143,12 @@ class MediaContainer extends MediaChromeHTMLElement {
 
   // First direct child with slot=media, or null
   get media() {
-    return this.querySelector(':scope > [slot=media]');
+    let media = this.querySelector(':scope > [slot=media]');
+
+    // Chaining media slots for media templates
+    if (media.nodeName == 'SLOT') media = media.assignedElements({flatten:true})[0];
```
cjpillsbury, line 149: Fine for now. May be some weird edge cases that are missed, but we can file followup issues if/when they show up and are a priority.

cjpillsbury (changes requested):
mostly nits.

themes/example.html
```diff
@@ -0,0 +1,18 @@
+<html>
```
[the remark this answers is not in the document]
heff: 👍

src/js/media-controller.js
```diff
@@ -86,7 +86,10 @@ class MediaController extends MediaContainer {
         super[fullscreenApi.enter]();
       },
       MEDIA_EXIT_FULLSCREEN_REQUEST: () => {
-        this.getRootNode()[fullscreenApi.exit]();
+        document[fullscreenApi.exit]();
+
+        // Shadow root throws an error for this function
+        // this.getRootNode()[fullscreenApi.exit]();
```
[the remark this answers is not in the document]
heff: It seems like it's gonna be context specific, depending on if we need the shadowRoot when that's the element's context and if it supports the needed APIs.

themes/media-theme-netflix.js
heff: How is `options` getting mutated? Either way I'm good with the spread.

heff:
Cleaned up and finally merging in.

dylanjha:
yooooo 🙌🏼 🎉

PR #43, opened 2021-08-02
Improv/dev env

cjpillsbury:
This PR moves the media-chrome dev + build environment to some standards with other open source projects for mux, including:
- use yarn consistently (instead of npm)
- update dependencies based on npm audit warnings
-  use webpack for build and dev server
- reorganize examples locations and filenames to keep them self-contained


webpack.common.js
```diff
@@ -0,0 +1,32 @@
...
+  resolve: {
+    extensions: ['.js'],
+  },
+  output: {
+    path: path.resolve(__dirname, 'dist'),
+    filename: '[name].js',
+    library: {
+      type: 'module',
+    },
+  },
```
mmcc, line 28: Does this all the work the same way it used to? Basically everything in `/src` should be transpiled and output to `dist`, with 2 additional files that include All The Things (`/index.js` and `/extras.js`).

The goal is for people to be able to just `import 'media-chrome'` and get the kitchen sink, or be able to just import the individual elements they want to use. Same for extras. If I'm recalling Webpack correctly, this is going to output 2  bundles based on the `entry` key.

mmcc (approved):
:wave: farewell, weird build.js

webpack.common.js
mmcc: err I meant to do "request changes" not "approve" because I think my output script question is still pretty important.

cjpillsbury: per chat with @heff - my concern with that goal is it's very likely that people will accidentally end up with duplicate modules (some bundled, some unbundled) which could lead to both bloat and unexpected behaviors. Thoughts?

mmcc: Well if the kitchen sink bundle is an ES Module, it's just importing everything anyway, so I don't _think_ duplication would do anything there since imports are only imported once. Right?

cjpillsbury: I'm currently creating modules that are also bundles, but I can def refactor our build to keep each module independent instead (different tradeoffs from a lib perspective).

cjpillsbury:
Closing per chat w/@mmcc. Will issue new PR for final build tool(s) to best achieve unbundled dist builds.

PR #47, opened 2021-08-09
Improv/a11y phase 1

cjpillsbury:
WIP - Marking as draft pull request.

Tasks:
- [ ] Update Media Chrome UI Elements so there are no issues reported in
  - [ ] Lighthouse (Chrome)
  - [ ] Firefox built in A11Y Issue Reporting
- [ ] Update all examples html so there are no issues reported in
  - [ ] Lighthouse (Chrome)
  - [ ] Firefox built in A11Y Issue Reporting
- [ ] Add npm script to run chrome headless + lighthouse via CLI


src/js/media-mute-button.js
```diff
@@ -45,6 +46,12 @@ slotTemplate.innerHTML = `
   <slot name="high">${highIcon}</slot>
 `;
 
+const updateAriaLabel = (el) => {
+  const muted = el.getAttribute(MediaUIAttributes.MEDIA_VOLUME_LEVEL) === 'off';
+  const label = muted ? verbs.UNMUTE() : verbs.MUTE();
+  el.nativeEl.setAttribute('aria-label', label);
+};
```
dylanjha, line 53: In general are we moving to a more functional coding-style?

Instead of an instance method on the class: `this.updateAriaLabel()`

fwiw I like the more functional style, I think it can be easier to reason about -- but just want to be explicit about the coding style we're going for in this project.

src/js/media-seek-backward-button.js
```diff
@@ -25,6 +33,8 @@ class MediaSeekBackwardButton extends MediaChromeButton {
   }
 
   connectedCallback() {
+    // NOTE: currently don't support changing the seek value, so only need to set this once on initialization.
```
dylanjha, line 36: good callout, I could see us very soon making the seek value configurable

dylanjha (approved):
this looks really good @cjpillsbury 

For posterity, as part of this PR it might be nice to include a screenshot of the rules that were failing before and a follow up screenshot now that all the rules are passing (we talked about both Lighthouse and the FF a11y panel.

We also discussed having a `yarn check-a11y` script or something like that we can run either locally or as part of CI to validate these steps and prevent regressions from being accidentally introduced down the road.

src/js/media-container.js
```diff
@@ -18,6 +19,7 @@ template.innerHTML = `
     :host {
       box-sizing: border-box;
       position: relative;
+      align-items: stretch;
```
heff, line 22: Did this fix something specific?

src/js/media-container.js
```diff
@@ -66,8 +73,10 @@ template.innerHTML = `
       width: 100%;
     }
   </style>
-  <slot name="media"></slot>
-  <slot></slot>
+  <section>
```
heff, line 76: Why add section here? (I think I just removed an internal container a few PRs ago) The downside of having an internal container is  if you want to change the flex-direction of the media-container, you can't easily when the setting is on a shadow dom element. I needed that feature in the spotify audio player demo.

src/js/labels/labels.js
```diff
@@ -0,0 +1,20 @@
...
+    ENTER_FULLSCREEN: () => 'enter fullscreen mode',
+    EXIT_FULLSCREEN: () => 'exit fullscreen mode',
+    ENTER_PIP: () => 'enter picture in picture mode',
+    EXIT_PIP: () => 'exit picture in picture mode',
+    SEEK_FORWARD_N_SECS: ({ seekOffset }) => `seek forward ${seekOffset} seconds`,
+    SEEK_BACK_N_SECS: ({ seekOffset }) => `seek back ${seekOffset} seconds`,
+};
+
+
+export const nouns = {
```
heff, line 15: Just interested, what's the benefit of differentiating nouns from verbs here?

src/js/media-chrome-button.js
```diff
@@ -78,7 +79,7 @@ template.innerHTML = `
   }
 </style>
 
-<button></button>
+<button part="inner-button"></button>
```
heff, line 82: It's worth debating/discussing whether we should use a part or vars here. If we use a part, it makes it really flexible but also a breaking change if we switch to not using a native button, whereas vars could be made to work the same way.

heff (commented):
Awesome to see progress here.

examples/standard.html
```diff
@@ -10,30 +11,33 @@
...
-      <media-playback-rate-button></media-playback-rate-button>
-      <media-pip-button></media-pip-button>
-      <media-fullscreen-button></media-fullscreen-button>
-    </media-control-bar>
-  </media-controller>
-  <div class="examples">
-    <a href="./">View more examples</a>
-  </div>
+  <main>
+    <h1>Media Chrome Standard Video Usage Example</h1>
```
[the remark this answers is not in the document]
cjpillsbury: "standard.html" was what used to be index.html in the project root. It's been around for awhile so I didn't want to assume it should be gutted, but happy to remove.

examples/themes/youtube-theme.html
```diff
@@ -133,36 +133,37 @@
...
-    <media-fullscreen-button></media-fullscreen-button>
-  </media-control-bar>
-</media-controller>
-
-<script>
-  const mediaMuteButton = document.querySelector('media-mute-button');
-
-</script>
-
+<main>
```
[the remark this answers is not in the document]
cjpillsbury: Correct, "landmarks" are recommended for a11y.

src/js/labels/labels.js
cjpillsbury: was aiming for a bigger picture semantics (since nouns make sense for certain UI/a11y metaphors and verbs make sense for others), but happy to remove since we can easily re-add it or implement it in whatever way makes sense longer term (e.g. if/when we support a11y+i18n and if we end up going with well-defined format string/labeling standards like the ICU message syntax, used by CLDR JSON data https://formatjs.io/docs/core-concepts/icu-syntax/ or similar)

src/js/media-chrome-button.js
cjpillsbury: Do you mean if we replace `<button/>` with nothing or we replace `<button/>` with e.g. a different element? I could see how the latter would lead to a breaking change, but not the former. Regardless, wouldn't these already be true, given that each element instance has a public `nativeEl` property?

src/js/media-container.js
cjpillsbury: This is also for applying descriptive landmarks for a11y (unfortunately we cannot assign the role to the custom element itself). Let me know if you'd like to touch base on this. Otherwise, I'll plan on ensuring that the spotify example works/lays out with the new DOM structure.

src/js/media-container.js
cjpillsbury: yes. It ensures that the newly-added landmark `section` sizes to the parent for both dimensions.

src/js/media-chrome-button.js
```diff
@@ -13,7 +13,7 @@ template.innerHTML = `
     background-color: var(--media-control-background, rgba(20,20,30, 0.7));
 
     /* Default width and height can be overridden externally */
-    padding: 10px;
+    padding: 0px;
```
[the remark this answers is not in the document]
cjpillsbury: Yeah. For a handful of things, it's looking like custom elements + a11y via ARIA will require compromises/decisions. My leaning is to prioritize a11y and then make our custom elements as easy to use within those constraints. (the same goes with adding a section to the `media-container`, below).

examples/standalone-controls.html
```diff
@@ -11,39 +12,38 @@
...
-  <media-seek-backward-button></media-seek-backward-button>
-  <media-seek-forward-button></media-seek-forward-button>
-  <media-volume-range></media-volume-range>
-  <media-time-range></media-time-range>
-  <media-playback-rate-button></media-playback-rate-button>
-  <media-pip-button></media-pip-button>
-  <media-fullscreen-button></media-fullscreen-button>
-</media-control-bar>
+  <h2>Inside the control-bar element</h2>
+  <p>Note: You only need to set the <code>media-controller=</code> attribute on the <code>media-control-bar</code>. The media-controller will pass the media state details to their child media-chrome elements automatically.</p>
```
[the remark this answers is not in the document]
cjpillsbury: Good catch. I don't believe it is. I'll create an issue and make this work with the new architecture.
Proposal for you: 
1. instead of assuming an `id` (which has to be globally unique on the DOM), we support any query selector. 
2. If there are concerns for backwards compatibility, we can do (1) as a separate attribute (`media-controller-selector`?)

src/js/media-mute-button.js
cjpillsbury: Short version: yeah, I'd say we should be using more functions/abstractions separable from the element classes. I would definitely prefer leaning more on using functions if/when it has a generic purpose that can be independently tested. Particularly with "pure" functions, but also with best practices on impure functions, this also helps us avoid unnecessary coupling/assumptions.

cjpillsbury:
> For posterity, as part of this PR it might be nice to include a screenshot of the rules that were failing before and a follow up screenshot now that all the rules are passing (we talked about both Lighthouse and the FF a11y panel).

I'll assume you'll want one documentation per A11Y testing per page unless I hear otherwise. I've also been using the free axe devtools Chrome extension as a 3rd audit. Let me know if you'd like me to include them as well.

examples/standard.html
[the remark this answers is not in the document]
heff: Yeah, we can remove standard. Maybe look for anything interesting in standard that is missing in basic.

dylanjha:
> I'll assume you'll want one documentation per A11Y testing per page unless I hear otherwise

@cjpillsbury If there's a lot of overlap, it doesn't need to be too exhaustive, but just a couple high level screenshot(s) of before/after to show: "these are a bunch of the issues that we had" and "now look, they're fixed". It just gives some more context around what exactly this PR accomplishes

cjpillsbury:
Resolved from other PR. Will still need a followup for baseline test scripts and snapshots.

PR #49, opened 2021-08-11
improv/a11y-plus-styling-container: directly apply aria roles and lab…

cjpillsbury:
…els to container/controller custom element to make custom styling easier.

Instead of adding a semantic html element (<section/>) as a root child of the <media-container/>/<media-controller/> custom element, we should apply a11y-related attributes to the custom element itself, primarily to make customization of styling and the like easier for implementors.

Confirmed:
- [x]  No errors reported related to this approach from axe Chrome extension audit
- [x]  No errors reported related to this approach from built-in Chrome Lighthouse dev tool audit
- [x]  Works as expected with macOS VoiceOver (announces region in addition to focusable element when tabbing into controls)
- [x]  Layout of examples/themes/spotify-theme.html still rendered as expected

NOTE: Although Firefox's built-in Accessibility dev tool reports an issue here, I believe this is actually a bug

### DIFF
--- a/src/js/labels/labels.js
+++ b/src/js/labels/labels.js
@@ -1,0 +1,4 @@
+export const nouns = {
+    AUDIO_PLAYER: () => 'audio player',
+    VIDEO_PLAYER: () => 'video player',
+};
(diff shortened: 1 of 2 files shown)


dylanjha (approved):
LGTM!

heff:
LGTM

Not directly relevant to this, but FYI, I'm still debating if the `audio` attribute is the best approach for allowing audio styling. It'd be nice if that was just automatic, but that's not the easiest to do unless we say the media-element must have `audio` in the name. Even then you may want to predefine the layout, or use a video layout for an audio media element. 🤷

PR #55, opened 2021-08-18
Add webkitEnterFullScreen for safari iphone (ios)

Nodalailama:
Hi,

Using an iphone app with a webview, I was not able to enter fullscreen. 

This pull request correct that behavior.

For testing which function name to use, "document.webkitExitFullscreen" exist but "document.webkitEnterFullScreen" do not.
Thus, I use "document.webkitExitFullscreen" to test the correct function name to enter fullscreen; "webkitEnterFullScreen" is the good one for iphone safari.

Take a look at https://developers.google.com/web/fundamentals/media/mobile-web-video-playback#toggle_fullscreen_on_button_click

And caniuse says RequestFullScreen is only available for ipad (with webkit prefix) not iphone: https://caniuse.com/?search=RequestFullScreen

Voilà.

### DIFF
--- a/src/js/utils/fullscreenApi.js
+++ b/src/js/utils/fullscreenApi.js
@@ -9,7 +9,7 @@ export const fullscreenApi = {
 };
 
 if (document.fullscreenElement === undefined) {
-  fullscreenApi.enter = "webkitRequestFullScreen";
+  fullscreenApi.enter = document.webkitExitFullscreen != null ? "webkitEnterFullScreen" : "webkitRequestFullScreen";
   fullscreenApi.exit = document.webkitExitFullscreen != null ? "webkitExitFullscreen" : "webkitCancelFullScreen";
   fullscreenApi.event = "webkitfullscreenchange";
   fullscreenApi.element = "webkitFullscreenElement";


dylanjha (approved):
This LGTM

dylanjha:
@heff I'm going to go ahead and merge this. Appears to be an oversight that this conditional was missing originally

heff:
👍 Maybe just test it on Safari desktop to make sure. The fullscreen API has had some weird transitions.

src/js/utils/fullscreenApi.js
```diff
@@ -9,7 +9,7 @@ export const fullscreenApi = {
 };
 
 if (document.fullscreenElement === undefined) {
-  fullscreenApi.enter = "webkitRequestFullScreen";
```
dylanjha, line 12: I'm running into a bug with this in Safari (Desktop) - it turns out:

* `document.webkitExitFullscreen` is defined, but `document.webkitEnterFullScreen` is not (see screenshot below):

webkit-exit-enter_2021-09-06_01-00-03

The way this logic works, in Desktop Safari under these conditions is assigning the `fullscreenApi.enter` to `webkitEnterFullScreen`, which is undefined, leading to the following error:

enter-not-a-function_2021-09-06_01-01-17



This is all in Safari `Version 14.1 (16611.1.21.161.6)`

dylanjha (changes requested):

luwes:
iPhone only supports requesting fullscreen on the video element directly as far as I know. 
this PR doesn't implement that I think.
https://github.com/muxinc/media-chrome/blob/c6e224e3f6169e554167c9f24dafad3b6050d7ee/src/js/media-controller.js#L90-L90

https://developer.mozilla.org/en-US/docs/Web/API/Element/requestFullScreen#browser_compatibility
https://developer.apple.com/documentation/webkitjs/htmlvideoelement/1630649-webkitenterfullscreen
https://github.com/sindresorhus/screenfull/issues/175

src/js/utils/fullscreenApi.js
luwes: Safari has old and new api's :(
https://github.com/sindresorhus/screenfull/blob/main/index.js#L12-L31

PR #58, opened 2021-09-08
Improv/a11y phase 1

cjpillsbury:


dylanjha:
Wait a second, I recognize this face. Is this the "Accessibility & Media Players" [guy from Demuxed](https://2021.demuxed.com/#speakers)?

cjpillsbury:
> Wait a second, I recognize this face. Is this the "Accessibility & Media Players" [guy from Demuxed](https://2021.demuxed.com/#speakers)?

🙈 🙉 🙊

dylanjha (approved):
Really good and thorough work here. 


https://user-images.githubusercontent.com/764988/132577204-38b035b4-56e9-4a07-8032-483a557052f9.mp4

src/js/media-chrome-button.js
```diff
@@ -99,6 +100,9 @@ class MediaChromeButton extends window.HTMLElement {
     this.nativeEl = buttonHTML.querySelector('div');
     
     this.setAttribute('role', "button");
+    if (isSafari) {
```
heff, line 103: What's the context around this and can we not avoid a browser UA check? Definitely not excited to get a bunch of these in the codebase.

src/js/utils/time.js
```diff
@@ -1,3 +1,55 @@
...
+
+  const timeString = timeParts
+    // Convert non-0 values to a string of the value plus its unit
+    .map((timeUnitValue, index) => timeUnitValue && toTimeUnitPhrase(timeUnitValue, index))
+    // Ignore/exclude any 0 values
+    .filter(x => x)
+    // join into a single comma-separated string phrase
+    .join(', ');
+
+  // If the time was negative, assume it represents some remaining amount of time/"count down".
```
heff, line 47: Thank you for the commenting! :)

src/js/media-chrome-button.js
cjpillsbury: short summary: browsers are inconsistent with screen reader behavior wrt aria. On macOS w/VoiceOver, Firefox/Chrome will announce changes to `aria-label`s while in focus. For Safari, it requires `aria-live`. However, if `aria-live` is set on Firefox/Safari, VoiceOver will announce changes even when not in focus. This ensure a decent user experience for those relying on screen readers across these browsers.

dylanjha: > Definitely not excited to get a bunch of these in the codebase.

I agree with this, and would like to avoid these checks whenever possible (if a check is required, much better to check for a specific feature).

IMO for a11y I think it makes sense to smooth out inconsistent browser behavior with VoiceOver. If the browser check is necessary for that, then I'm in favor of it.

It would be good to add a comment to explain the context about this exactly so that if we discover the browser inconsistency is no longer a problem in the future we can remove the check.

cjpillsbury: Given the concern of browser checks and the different edge cases, I think I'm leaning towards removing the condition altogether. Since Safari is the browser that is inconsistent with the spec in this case and since it has a "workaround" with VoiceOver (you can e.g. hold the cmd key and it should announce the label of the currently-focused element), I'm convinced that we should just remove it given the other concerns.

heff: Thanks for all the context, and I appreciate where we landed.

dylanjha: Cool, makes sense.

If the inconsistency was more critical I might push harder that we smooth over that edge, but since there's a workaround this feels like the right decison.

PR #61, opened 2021-09-11
Fix up some examples, update package.json for 0.3.0

dylanjha:
Here's the diff between 0.2.0 and main: https://github.com/muxinc/media-chrome/compare/v0.2.2...main

- [x] fix youtube theme styles
- [x] fix netflix theme styles
- [x] update version to 0.3.0 in package.json
- [x] add a link to spotify theme in the examples  
- [x] fix a bug where attributes were set in the constructor which made React DOM very not happy (465303252f285e151682e64ee085947294605778)

Sidenote: the Youtube media element example was busted b/c of a race condition where we query the media slot state before the youtube player in ready, so I have a fix for that proposed:

https://github.com/muxinc/youtube-video-element/pull/9

# Youtube theme was a little off

### Before

youtube-slightly-busted_2021-09-10_15-09-39

### After

youtube-fix_2021-09-10_15-08-04

# Netflix too

## Before

Screen Shot 2021-09-10 at 3 20 52 PM

## After

netflix-better_2021-09-10_15-40-09

# Other examples look good

basic_2021-09-10_15-42-17

standalone_2021-09-10_15-42-28

spotify_2021-09-10_15-42-00


examples/themes/netflix-theme.html
```diff
@@ -17,6 +17,8 @@
     /* Keep the buttons from overflowing */
     min-width: 420px;
     height: 720px;
+    flex-direction: column;
```
heff, line 20: If this needs to be added now it makes me worry that we lost something in media-container that did this already.

cjpillsbury: May have been lost earlier... I checked out `6b916dbd788af02836bf0dbf436f3f06c75208cc` locally (only changes were build/devops-related) and did a from scratch install + dev run and here's what I'm seeing:
Screen Shot 2021-09-13 at 1 24 02 PM

dylanjha: `git bisect` helped me figure out that this is the commit that introduced the change in behavior

https://github.com/muxinc/media-chrome/commit/cd77c4cba7fcc8c4b6bb71f55fcfe7548805b1a1

Can verify by the commit before it:

`f8b8e888f783c7e4e4f7688a744bd1051813eb3f`

f8b8-netflix_2021-09-14_00-27-20

`cd77c4cba7fcc8c4b6bb71f55fcfe7548805b1a1`

cd77_2021-09-14_00-27-35

The specific diff is actually [this tiny change](https://github.com/muxinc/media-chrome/commit/cd77c4cba7fcc8c4b6bb71f55fcfe7548805b1a1#diff-a093df03191bf45d0036001fefb7ca7217096f130642377d56e0cd90312b3e23L68-L70)

```diff
- <div id="container">
-   <slot></slot>
- </div>
+ <slot></slot>
```

The reason is because `<media-container>` is `display: flex` (or `display: inline-flex`, either one, doesn't really matter). All the child elements of media-container were previously wrapped in a `<div>`. The actual styles being applied to `#container` didn't matter, just the fact that the slot was wrapped in a div meant that there was only 1 child inside the flexbox. By removing that container then in cases where there are multiple children stacked on top caused this theme to break.

Now I'm thinking about what the default should be. We may actually want `justify-content: flex-end;` to be applied to `media-controller`  by default, that would seem more intuitive.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "media-chrome",
-  "version": "0.3.2",
+  "version": "0.3.0",
```
heff, line 3: Don't manually bump the version. I use [np](https://www.npmjs.com/package/np) which handles bumping, tagging, and github releases, so a manual bump messes that up.

heff:
@dylanjha let's catch up on this one and get it resolved.

dylanjha:
out of date, these examples have been fixed.

Issue #57, opened 2021-08-30
Preview images appear to be broken


heff:
Compare main branch youtube theme

![image](https://user-images.githubusercontent.com/166/131271416-2c9da27e-ef65-4266-ba0c-c27443c09383.png)

to https://unpkg.com/media-chrome@0.2.2/examples/themes/youtube-theme.html

![image](https://user-images.githubusercontent.com/166/131271439-87767ed3-3752-4ea1-b340-c2816e20b423.png)

cjpillsbury:
This was identified and should already be resolved in a pending branch (yet to be PR-ed based on current backlog. Thanks for filing it. I'll be sure to associate it with the branch/PR once it's ready.

cjpillsbury:
Resolved by #62

Issue #52, opened 2021-08-11
Simplify and expand ways media ui elements may associate themselves with a media-controller


cjpillsbury:
- Remove the need to add `media-chrome-attributes` attribute for custom media ui elements, so simply having `observedAttributes` with well-defined media ui attributes is sufficient for identifying a custom media ui element
  - Re-introduce `whenDefined` to (asynchronously) defer checking if a custom element is a media ui element until after it's been defined
  - Ensure custom elements containing media ui element descendants in their `shadowRoot` also work.
- Add well-defined events as an alternative mechanism to associate/unassociate media ui elements to an ancestor (via event bubbling) to handle edge cases/as an "escape hatch" solution.

NOTE: After reviewing the [Web Components Context Proposal](https://github.com/webcomponents/community-protocols/blob/main/proposals/context.md), I believe this introduce unnecessary coupling between our "simple" media ui elements/components (which will be made even simpler by resolving this issue) and the `media-controller`. That said, it _**roughly**_ parallels similar concepts in other frameworks (e.g. the React context, context Provider, and the assumed ancestor/descendant relationship), so we may want to revisit using this in some way as a potential shareable abstraction across target frameworks.

cjpillsbury:
Resolved by #62

PR #62, opened 2021-09-13
Improv/rearch simplify expand mue association v2

cjpillsbury:
* adds features articulated by #52 
  * (re) adds support for event-style association/unassociation (currently unused by any "core"/"built in" MCEs)
  * removes now-fully-redundant attribute-based identifiers of "core"/"built-in" MCEs (still supported for convenient custom html-only MCEs)
* (re)adds support for async checking if any descendants of "monitored elements" (including `media-controller` itself) are MCEs if/when they are custom elements that are not (yet) registered.
* bugfix for #57 (resolved via new expanded architecture. Bug was introduced when we removed the events option from our "built in"/"default" media chrome elements in the new architecture)


cjpillsbury:
As an FYI, I found some "pre-existing conditions" (since before the overall rearchitecture) if/when a user adds a media element asynchronously/programmatically. Nothing ends up throwing an error or anything, but some of the video element's attributes are not respected/acknowledged by media-chrome/the media-controller. I'll be sure to write up a generic issue that we can triage/break into subtasks as we see fit.

src/js/media-controller.js
```diff
@@ -457,6 +445,16 @@ const shouldGetMediaUIElementDescendants = (el) => {
...
 };
 
+const isUndefinedCustomElement = (el) => {
+  const name = el?.nodeName.toLowerCase();
+  return name.includes('-') && !window.customElements.get(name);
+};
+
+const getRegisteredCustomElementPromise = (el) => {
+  const name = el?.nodeName.toLowerCase();
+  return window.customElements.whenDefined(name).then(() => el);
```
dylanjha, line 455: TIL `window.customElements.whenDefined`

dylanjha (approved):
✅  Not having to set `MEDIA_CHROME_ATTRIBUTES` is a huge win
✅  Also a big win fixing #57

src/js/media-controller.js
```diff
@@ -468,6 +466,10 @@ const shouldGetMediaUIElementDescendants = (el) => {
  *  - Have a `media-chrome-attributes` attribute with at least one well-defined media chrome attribute
  */
 const getMediaUIElementDescendants = (rootNode) => {
+  // If the rootNode is a custom element that's not yet defined/ready, so return an array of a promise to yield the
+  // element once it's registered for us to check at that time.
+  if (isUndefinedCustomElement(rootNode)) return [getRegisteredCustomElementPromise(rootNode)];
```
heff, line 471: Returning a promise that we then need to check for later is creating some pretty complicated logic. Can we instead check the element defined-state when we need to operate on it and defer that operation then?

src/js/media-controller.js
```diff
@@ -532,16 +534,58 @@ const propagateMediaState = (els, stateName, val) => {
...
+    const el = evt?.composedPath()[0] ?? evt.target;
+    associateCallback(el);
+  };
+
+  const unassociateElementHandler = (evt) => {
+    const el = evt?.composedPath()[0] ?? evt.target;
+    unassociateCallback(el);
+  };
+
+  root.addEventListener(MediaUIEvents.MEDIA_ASSOCIATE_ELEMENT_REQUEST, associateElementHandler);
```
heff, line 572: Why does every root el need this handler?

heff (commented):
I immediately regret making my media-chrome time at the end of the day... I can't follow the element association logic anymore. I'm gonna need you walk me through it IRL.

src/js/media-controller.js
cjpillsbury: If I'm understanding your proposal correctly, we'd want to make sure any "not yet registered but eventually will be identified as media ui elements" are "initialized" correctly with any values, and we'd (probably) want to **_not_** "associate" "not yet registered but eventually but eventually will definitely not be identified as media ui elements", but we can talk through alternative ways of handling these cases when we go through the code together.

src/js/media-controller.js
cjpillsbury: This was for consistency of the "escape hatch event". If someone is relying on this for a (likely dynamic sub-)component, this ensures that the event would work whether it's a direct descendant of `<media-controller/>` or is used in a "standalone component" manner.

heff: Got it. Makes sense.

src/js/media-controller.js
```diff
@@ -447,75 +427,51 @@ const setAttr = (child, attrName, attrValue) => {
...
-const hasDescendants = (el) => {
-  return !!(el?.childNodes?.length || el?.shadowRoot?.childNodes?.length);
-};
+const isMediaSlotElementDescendant = (el) => !!el.closest?.('*[slot="media"]');
 
-const shouldGetMediaUIElementDescendants = (el) => {
-  return !isMediaSlotElement(el) && hasDescendants(el);
+const isUndefinedCustomElement = (el) => {
+  const name = el?.nodeName.toLowerCase();
+  return name.includes('-') && !window.customElements.get(name);
```
heff, line 434: We unfortunately can't rely on `customElements.get(name)` for readiness of the element. See this note in my themes PR...
https://github.com/muxinc/media-chrome/blob/5d9773ca957b65a33e5f8dbaa882332bf2895bf5/src/js/media-controller.js#L415
It feels like a major bug, and was not easy to figure out what the hell was going wrong in a previous iteration of this.

What that means is we just have to always use the `whenDefined` promise. But that's not the worst and will simplify the code a little.

src/js/media-controller.js
```diff
@@ -447,75 +427,51 @@ const setAttr = (child, attrName, attrValue) => {
...
+
+  const { children = [] } = rootNode ?? {};
+  const shadowChildren = rootNode?.shadowRoot?.children ?? [];
+  const allChildren = [...children, ...shadowChildren];
+
+  // Traverse all children (including shadowRoot children) to see if they are/have Media State Receivers
+  allChildren.forEach(child => traverseForMediaStateReceivers(child, mediaStateReceiverCallback));
 };
 
 const propagateMediaState = (els, stateName, val) => {
```
heff, line 477: Answering your comment: Checking for slot=media here I don't think is needed anymore since we're checking it before adding an element as a state receiver.

src/js/media-controller.js
```diff
@@ -532,16 +488,73 @@ const propagateMediaState = (els, stateName, val) => {
...
+    mutationsList.forEach(mutationRecord => {
+      const { addedNodes = [], removedNodes = [], type, target, attributeName } = mutationRecord;
+      if (type === 'childList') {
+        // For each added node, register any Media State Receiver descendants (including itself)
+        Array.prototype.forEach.call(addedNodes, node => traverseForMediaStateReceivers(node, registerMediaStateReceiver));
+        // For each removed node, unregister any Media State Receiver descendants (including itself)
+        Array.prototype.forEach.call(removedNodes, node => traverseForMediaStateReceivers(node, unregisterMediaStateReceiver));
+      } else if (type === 'attributes' && attributeName === MediaUIAttributes.MEDIA_CHROME_ATTRIBUTES) {
+        if (isMediaStateReceiver(target)) {
+          // Changed from a "non-Media State Receiver" to a Media State Receiver: register it.
```
heff, line 534: This isn't necessarily true if the attr list just changed, right? Though we're checking if it already is registered later so not an issue.

heff (commented):
This is feeling really good. A few comments that I hope aren't too disruptive.

src/js/media-controller.js
cjpillsbury: Yup that's exactly right. I'll update the code comment for further clarity.

src/js/media-controller.js
cjpillsbury: Great callout. Will refactor and put a very wordy code comment where we're implementing the logic.

src/js/media-controller.js
```diff
@@ -447,75 +427,51 @@ const setAttr = (child, attrName, attrValue) => {
...
-    // which will yield an array of arrays
-    // 2. flatten that into a single array (aka map+flat aka flatMap)
-    // 3. spread this as the rest of the descendant arrays' elements to return
-    ...Array.prototype.flatMap.call(allChildNodes, getMediaUIElementDescendants)
-  ];
-};
+const traverseForMediaStateReceivers = (rootNode, mediaStateReceiverCallback) => {
+  // We (currently) don't check if descendants of the `media` (e.g. <video/>) are Media State Receivers
+  // See also: `propagateMediaState`
+  if (isMediaSlotElementDescendant(rootNode)) {
```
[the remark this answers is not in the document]
cjpillsbury: Yeah this was a carry-over/migration from something in the old architecture:
https://github.com/muxinc/media-chrome/blob/f8b8e888f783c7e4e4f7688a744bd1051813eb3f/src/js/media-controller.js#L393-L395

src/js/media-controller.js
```diff
@@ -450,101 +428,132 @@ const setAttr = (child, attrName, attrValue) => {
...
+
+  // Custom Elements (and *only* Custom Elements) must have a hyphen ("-") in their name. So, if the rootNode is
+  // a custom element (aka has a hyphen in its name), wait until it's defined before attempting traversal to determine
+  // whether or not it or its descendants are Media State Receivers.
+  // IMPORTANT NOTE: We're intentionally *always* waiting for the `whenDefined()` Promise to resolve here
+  // (instead of using `window.customElements.get(name)` to check if a custom element is already defined/registered)
+  // because we encountered some reliability issues with the custom element instances not being fully "ready", even if/when
+  // they are available in the registry via `window.customElements.get(name)`.
+  const name = rootNode?.nodeName.toLowerCase();
+  if (name.includes('-') && !isMediaStateReceiver(rootNode)) {
```
heff, line 471: Took me a minute to figure out why we check for `!isMediaStateReceiver` rather than just treat every undefined custom element the same, but I get now the assumption is that if it's a state receiver then it must be defined. Which isn't technically true with the special attr that could exist before it's defined. I can't come up with why someone would ever do that so I'll leave it up to you if it's worth addressing.

PR #65, opened 2021-09-17
fix/range-background: Make sure numeric values are removed as attribu…

### LINKED ISSUE #56: No progress bar background initially
Looks like we lost a background on the main progress bar at some point. It may only happen when preload="none".

image

Compare /examples/standalone-controls.html

image

to https://unpkg.com/media-chrome@0.2.2/examples/standalone-controls.html

image

cjpillsbury:
…tes if/when NaN (instead of set).

* fixes #56

### DIFF
--- a/src/js/media-controller.js
+++ b/src/js/media-controller.js
@@ -444,6 +444,9 @@ const setAttr = (child, attrName, attrValue) => {
     if (attrValue) return child.setAttribute(attrName, '');
     return child.removeAttribute(attrName);
   }
+  if (Number.isNaN(attrValue)) {
+    return child.removeAttribute(attrName);
+  }
   return child.setAttribute(attrName, attrValue);
 };
 


src/js/media-controller.js
```diff
@@ -444,6 +444,9 @@ const setAttr = (child, attrName, attrValue) => {
     if (attrValue) return child.setAttribute(attrName, '');
     return child.removeAttribute(attrName);
   }
+  if (typeof attrValue === 'number' && Number.isNaN(attrValue)) {
+    return child.removeAttribute(attrName);
+  }
```
dylanjha, line 449: Makes sense 👍🏼

dylanjha (approved):

src/js/media-controller.js
heff: If you use `Number.isNaN` the type check is supposed to be not necessary.

heff (approved):

src/js/media-controller.js
cjpillsbury: Yup good callout. Was modeling it on "type" + specific behavior, but it's redundant in this case. Will remove before merging.

PR #74, opened 2021-10-11
Improv/styling rework

cjpillsbury:
Reworking some of the DOM and styling structures to make use cases such as responsive design easier, less code, and more predictable for those less familiar with media-chrome.

Changes include:

* media-container
  * no longer has a max-width constraint
    * Motivation: more intuitive behavior for devs (make it work like “any other DOM element”)
    * Breaking change: Anyone who was relying on automatically setting a max-width for media-chrome
      * Fix for use case: apply max-width: 100% to media-controller/media-container “from the outside” via css
    * New/Preferred Approach: devs should style media-controller/media-container’s width as they see fit, just like they would with other DOM elements.
  * no longer uses “column-reverse” layout
    * Motivation: more intuitive behavior for devs
    * Breaking change: Anyone who has already built a player chrome relying on this layout
      * Fix for use case: change DOM order so top-most component is first in the DOM, etc.
    * New/Preferred Approach: use new slots & layers architecture
  * Slots & layers: See discussion, below.
  * Added all MediaUIAttributes to media-container
    * Motivation: Allows for easier styling both internally and externally based on attribute values
  * Removed “built-in” play/pause click handler
    * Motivation: Many players do not have this behavior and there was no way to configure this.
    * Breaking change: Anyone who was relying on touch/click to play/pause the media
      * Fix for use case: Manually add a component to the gestures slot/layer that toggles play/pause when clicked/touched (e.g. a customized non-visible media-play-button)
    * New/Preferred Approach: Can use the newly-added gestures slot/layer to add any region-specific gesture behavior. Should have a discussion of whether we want anything to be added to the default or add & document an example of this use case (perhaps along with other gestures examples).
* media-control-bar
  * Layout changes (no longer wraps)
    * Motivation: layout behavior was harder to control for the most common use case of a single control bar, particularly making auto-scaling/sizing easier for mobile-first and responsive designs.
    * Breaking change: Anyone currently relying on wrap.
      * Fix for use case: Apply flex-wrap: wrap to media-control-bar “from the outside” via css
    * New/Preferred Approach: Use more than one control bar or other DOM elements ordered in whatever order you want them to appear vertically on the DOM (using the new slots & layers architecture). Make sizing of ancestor media-control-bar plus media-controller/media-container easier to automatically size/layout controls.
* media-chrome-buttons
  * Removed explicit sizing from default SVGs
    * Motivation: Allows the button to determine size of SVG based on intrinsic aspect ratios identified by the viewport, making auto-scaling/sizing easier for mobile-first and responsive designs.
    * New/Preferred Approach: An SVG has an intrinsic aspect ratio and maintains it by default. By setting only height or width (but not both) “from the outside” in the media-chrome-button, this allows auto-scaling without distortion by default. For more on this, see e.g. https://css-tricks.com/scale-svg/ 
  * Simplified and updated DOM structure and styling, including removing hard constraints on sizes and certain “baked-in” text-related size styles (e.g. line-height)
    * Motivation: This allows media-chrome-buttons to behave more like standard button DOM elements and making auto-scaling/sizing easier for mobile-first and responsive designs. It also greatly simplifies the template’s DOM and CSS.
    * Breaking change: Anyone who was using “icon buttons” in a fairly non-standard way. Anyone who was relying on not allowing icons to responsively “shrink”/“grow”.
      * Fix for use case: Apply styles to slotted text for non-standard use case. Apply styles to button “from the outside” and/or use supported CSS vars to force size constraints if the button would be “shrunk” from its parent.
    * New/Preferred Approach: This is part of making mobile-first and responsive designs easier via auto-scaling/sizing.
      * * NOTE: Some of the built in styling for text-specific buttons was removed. This is because it put hard constraints on the possible sizes of the buttons, making auto-shrinking cases not work without significant effort, which is particularly relevant for simple/easy mobile-first and responsive design use cases. See the various media element examples for how text will render by default now, and see the refactored media-playback-rate-button for the current solution for media-chrome-buttons that will frequently have text children instead of icon children.
* media-text-display
  * Simplified styling and added container wrapper for slotted text content
    * Motivation: Makes auto-sizing/scaling easier. Makes it easier to use media-text-display directly for custom use cases.
    * Breaking change: Anyone who was relying on the original default styling for more complex use cases.
      * Fix for use case: Update styles “from the outside” and potentially use —media-text-content-height variable to handle these use cases.
    * New/Preferred Approach: This is part of making mobile-first and responsive designs easier via auto-scaling/sizing, but also makes it easier to throw text and text-related DOM elements inside a media-text-display for automatic styling/sizing behavior.
      * * NOTE: See the various media element examples for how text will render by default now. See updated examples such as the updated netflix theme for custom use cases.
media-playback-rate-button
  * Update template to handle one-off text sizing/layout for text-centric media-chrome-button
    * Motivation: Makes auto-sizing/scaling easier for most (icon-based) media-chrome-buttons while still supporting text-based buttons
    * New/Preferred Approach: N/A* (wasn’t a feature prior to refactor)
      * * NOTE: This is effectively a “mash-up” of media-chrome-button and (parts of) the way the refactored media-text-display works. We may want to revisit this if/when we end up with more use cases that would likely have text-based buttons to come up with more extensible/reusable code (either through inheritance or composition)


src/js/media-container.js
```diff
@@ -19,58 +19,100 @@ template.innerHTML = `
...
       position: relative;
-
-      /* Position controls at the bottom  */
-      display: inline-flex;
-      flex-direction: column-reverse;
+      display: inline-block;
 
       /* Max out at 100% width for smaller screens (< 720px) */
-      max-width: 100%;
+      // max-width: 100%;
```
cjpillsbury, line 25: NOTE (opinionated): This should be applied externally, as it's a simple style adjustment, familiar to anyone who may want it, and would be more generically predictable styling/layout behavior for someone unfamiliar with media-chrome.

src/js/media-container.js
```diff
@@ -19,58 +19,100 @@ template.innerHTML = `
...
-  <slot></slot>
+
+  <span part="layer media-layer">
+    <slot name="media"></slot>
+  </span>
+  <span part="layer gesture-layer">
+    <slot name="gestures-overlay"></slot>
+  </span>
+  <!--
+  <span part="layer text-tracks-layer">
```
cjpillsbury, line 98: Could uncomment this if we want to anticipate our "let media-chrome optionally manage sub/cc rendering" option (though we'd still need to pin down some of the details before we could fully implement).

src/js/media-playback-rate-button.js
```diff
@@ -11,16 +11,34 @@ import { nouns } from './labels/labels.js';
 const DEFAULT_RATES = [1, 1.25, 1.5, 1.75, 2];
 const DEFAULT_RATE = 1;
 
+const slotTemplate = document.createElement('template');
```
cjpillsbury, line 14: Since we removed some of the font-centric "baked in" stuff from media-chrome-button, this is the current proposed solution for buttons that will use text. (We could also add a "text-button" superclass that media-playback-rate-button extends)

examples/basic.html
```diff
@@ -13,14 +13,16 @@ <h1>Media Chrome Basic Audio and Video Usage Example</h1>
...
     ></audio>
-    <media-control-bar>
-      <media-play-button></media-play-button>
-      <media-time-display show-duration></media-time-display>
-      <media-time-range></media-time-range>
-      <media-playback-rate-button></media-playback-rate-button>
-      <media-mute-button></media-mute-button>
-      <media-volume-range></media-volume-range>
-    </media-control-bar>
+    <span slot="custom-chrome">
```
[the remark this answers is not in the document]
heff: Yeah, I really don't like that this is in "basic.html", as in it's a required element for basic usage. Media chrome is all about "custom chrome", so needing to add another element to do "custom chrome"... wut?

Let's get specific on what we'd be fixing with this. If this is needed I'd hope it's for the 10-20% of use cases, not 80% of use cases.

dylanjha:
Trying to piece together some documentation to understand what this looks like from the DevEx perspective.

# Controls

Often times for your web player you want two sets of controls, one set for mobile on small screens that have a touch interface and another set for desktop devices that have a mouse interface.

For former (mobile) often times you want to overlay the controls on the media player. You can do something like this:

## Slots

Slots are used to tell the `media-controller` where you want your controls positioned.

### slot="top-controls"

Specify an element like `<div slot="top-controls"></div>` and this element will overlay at the top area of the media element

### slot="controls-overlay"

Specify an element like `<div slot="controls-overlay"></div>` and this element will overlay directly on top of the media element

### slot="bottom-controls"

Specify an element like `<div slot="bottom-controls"></div>` and this element will overlay at the bottom area of the media element

### <media-control-bar>

Media control bar by default will be a control bar pinned to the bottom of the media element

```html
<media-controller>
  <!-- use the slot="top-controls" to overlay controls at the top of the player -->
   <div slot="top-controls">
      <media-mute-button></media-mute-button> <!-- put the mute button in the top left -->
      <div class="spacer"></div>
      <media-pip-button> </media-pip-button> <!-- put the pip and fullscreen buttons in the top right -->
      <media-fullscreen-button></media-fullscreen-button>
   </div>
  <!-- use the slot="controls-overlay" to overlay controls in the middle of the player -->
   <div slot="controls-overlay">
      <media-seek-backward-button></media-seek-backward-button>
      <media-seek-forward-button></media-seek-forward-button>
   </div>
  <!-- use the slot="bottom-controls" to overlay controls at the bottom of the player -->
   <div slot="bottom-controls">
   </div>
  <!-- media-control-bar will positioned at the bottom of the player by default -->
   <media-control-bar>
     <media-time-range></media-time-range>
     <media-time-display></media-time-display show-duration>
   </media-control-bar>
</media-controller>
```

### Here's some examples

mc_2021-10-14_09-23-56

mc-mobile_2021-10-14_09-25-24


----

^ Is this roughly what the documentation would look like for using the new "slots"?

# Bugs, Takeaways, Possible improvements

1. Whats the difference between `<media-control-bar>` and `slot="bottom-controls"`

2. The naming feels a little inconsistent "top-controls", "controls-overlay", and "bottom-controls" -- technically the top and bottom are "overlays" too (in a sense that they are overlaid onto the media element. `top-overlay`, `center-overlay`, `bottom-overlay` -- would that make more sense?

3. When using `slot="controls-overlay"` I was expecting things to be automatically centered, but I had to add some code for that, is this expected? It felt like this should have maybe been done for me:

```
.overlay {
  flex-grow: 1;
  display: flex;
  align-items: center;
  justify-content: center;
}
```

Similarly, I think I had to do the same thing for `slot="top-controls"`

4. I had to add some code like this to get the autohide behavior to hide the top overlay and controls-overlay:

```
media-controller[user-inactive="user-inactive"]:not([media-paused]) :is(.top-controls, .overlay, media-control-bar)) {
  opacity: 0.01;
  transition: opacity 1s;
}
```

^ This feels like a lot to figure out, could we make this easier? I'm not sure why autohide wasn't automatically working, and I'm also wondering if all overlays should have autohide behavior by default when the media is playing.

src/js/media-chrome-button.js
```diff
@@ -42,44 +45,23 @@ template.innerHTML = `
...
-    vertical-align: middle;
-    border: none;
-    margin: 0;
-    padding: 0;
-    text-decoration: none;
-    background: transparent;
-    color: #ffffff;
-    font-family: sans-serif;
-    font-size: 14px;
-    line-height: 24px;
```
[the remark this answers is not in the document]
heff: Pay close attention to the layout impacts then, like at the single pixel level. At least the way I CSS, consistent pixel heights is usually dependent on the line height.

src/js/media-fullscreen-button.js
```diff
@@ -12,12 +12,12 @@ import { Window as window, Document as document } from './utils/server-safe-glob
 import { MediaUIEvents, MediaUIAttributes } from './constants.js';
 import { verbs } from './labels/labels.js';
 
-const enterFullscreenIcon = `<svg aria-hidden="true" xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24">
```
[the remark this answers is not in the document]
heff: Nothing about that sounds wrong. It'd be helpful to see an example of this working.

src/js/media-container.js
heff: I wish I had documented the issue better, but it's solving something... clearly something with smaller screens. An element limited to the width of its container doesn't feel super "unpredictable" to me. Is there more I'm missing?

cjpillsbury:
> ^ Is this roughly what the documentation would look like for using the new "slots"?
Roughly, yeah. High level: 
* The container is broken into "layer"s that cover the entire height/width of a `<media-controller/>` (/"`<media-container/>`"), at least by default.
* Most slots (e.g. the "media" slot and the "controls-overlay" slot you're using) will have a 1-1 relationship with a "layer", aka anything you slot there will be the sole descendant of the layer that covers the entire height/width.
* "x-controls" slots (as well as the default/unnamed slot) are the exception here. They are all in one of these layers (the "controls-layer") and (again, by default) use a column flex layout (+a spacer) to allow for top/bottom controls based on your specified or unspecified slot.

Here's the "core" of the `media-container`'s DOM "under the hood":

```html
  <span part="layer media-layer">
    <slot name="media"></slot>
  </span>
  <span part="layer gesture-layer">
    <slot name="gestures-overlay"></slot>
  </span>
  <!--
  <span part="layer text-tracks-layer">
    <slot name="text-tracks-renderer"></slot>
  </span>
  -->
  <span part="layer controls-overlay-layer">
    <slot name="controls-overlay"></slot>
  </span>
  <span part="layer controls-layer">
    <slot name="top-controls"></slot>
    <slot name="middle-controls"><span class="spacer"></span></slot>
    <slot></slot>
    <slot name="bottom-controls"></slot>
  </span>
```

> 1. Whats the difference between <media-control-bar> and slot="bottom-controls"
* `<media-control-bar/>` is a visual component (a "container component") that can be used on its own (see our "standalone-controls" example), can be used multiple times in a `media-controller` and its slots (see the updated "youtube-theme"/"netflix-theme"/"demuxed-2021-theme" examples in this PR). 
* Slots don't "take up space", but what is slotted in them can and often do. So e.g. you might want to have something like:

```html
<media-controller class="media-theme-youtube">
        <video
          slot="media"
          src="https://stream.mux.com/DS00Spx1CV902MCtPj5WknGlR102V5HFkDe/high.mp4"
          muted
          preload="none"
          crossorigin
        >
        </video>
        <media-control-bar slot="bottom-controls">
           <!-- controls go here -->
        </media-control-bar>
      </media-controller>
```

I *_think_* your broader point is maybe something like "why do we need a `slot="bottom-controls"` at all"? The main reason I did that is to have a guaranteed and well-defined "whatever gets slotted here is always going to be the last node in the controls-layer". That said, this was more about parity with things like the "top-controls", so I can definitely see arguments to removing it altogether unless/until a use case comes up.

src/js/media-container.js
```diff
@@ -130,17 +172,15 @@ class MediaContainer extends window.HTMLElement {
   }
 
   static get observedAttributes() {
-    return ['autohide'].concat(super.observedAttributes || []);
+    return ['autohide'].concat(MEDIA_UI_ATTRIBUTE_NAMES);
```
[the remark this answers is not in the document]
heff: Makes sense to me. 👍

src/js/media-container.js
cjpillsbury: Hm. I disagree. That's not really how any html elements behave "out of box". With pretty much any visual HTMLElement, if you don't apply any CSS to them, the child will size itself and the parent will be sized by its children (with different default sizing/overflowing/etc. behavior based on the default display/layout rules of the elements in question).

examples/basic.html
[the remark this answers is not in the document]
cjpillsbury: The audio styling/layout behavior was more about deciding whether we want the baked in styling/layout of media-container to be more presumptuous or less presumptuous, since our current architecture doesn't allow us to use the data management parts of media-chrome without also being tied to the UI.

That said, I think your point is a reasonable preference for "default" audio layout behavior. I'll rework so a "media-control-bar only" case "just works" for audio (I think it will also work reasonably well for most of the other plausible audio chromes).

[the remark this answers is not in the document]
cjpillsbury: Yeah I should be able to make this "just work" for audio (I suspect even for the spotify theme). Just need to add some additional attribute selectors + appropriate CSS under the hood.

heff:
> I like the specificity of this, but what are you defining as "controls" here? Are we assuming that all the top-chrome, centered-chrome, bottom-chrome slots all contain controls?

Sorry, by "controls" I really meant "chrome". So if you know you want to layer something over the chrome (like an endscreen, or settings menu) you can put it there. Or put something like a full-screen preview thumbnails under the chrome (like youtube scrubbing) you can use under-chrome. Does that make sense? Open to other ideas.

> but I don't understand the name "timed-text"

By "timed-text" I meant the parent concept of captions and subtitles, like the "text-track-layer" already in this PR. I also kind of assumed the only people who cared/used that slot would be building the captions display component. But that's a bad assumption and too specific a name for this one. You probably also want to put end-screen things in this slot so that they don't overlap with the top and bottom chrome.

image

But I do think we could use this slot for captions also, and would need to if we want them to move with the controls.

Is it totally confusing if we have both:
* `middle-chrome` - for captions, end-screen items, etc. - same layer as top/bottom chrome and doesn't overlap
* `centered-chrome` - for big play button, mobile overlaid controls, etc. - always vertically centered to the video, different layer from top/bottom chrome and could overlap them.

> `<media-control-bar slot="bottom-chrome">` love this, for the reasons you specified

Cool. If you like that, what do you think the default slot should be?

src/js/media-container.js
```diff
@@ -19,58 +19,111 @@ template.innerHTML = `
...
+  <span part="layer text-tracks-layer">
+    <slot name="text-tracks-renderer"></slot>
+  </span>
+  -->
+  <span part="layer controls-overlay-layer">
+    <slot name="controls-overlay"></slot>
+  </span>
+  <span part="layer controls-layer">
+    <slot name="top-controls"></slot>
+    <span class="spacer centered"><slot name="middle-controls"></slot></span>
```
heff, line 118: Why do we need a "spacer" around this slot?

src/js/media-container.js
```diff
@@ -19,58 +19,111 @@ template.innerHTML = `
...
-      flex-direction: column-reverse;
+      display: inline-block;
 
       /* Max out at 100% width for smaller screens (< 720px) */
-      max-width: 100%;
+      // max-width: 100%;
       background-color: #000;
     }
 
+    :host(:not([audio])) *[part~=layer] {
```
heff, line 29: I'm not sure if we want this flex layout to be the default for every layer we've talked about. I see different needs/expectations for the layer called "centered", the layer that has top/bottom/middle chrome in it, and then any less specific layer where you might just expect it to work like any other html element.

heff (commented):
Sounds like we should chat about default audio slots IRL, but just to call out my assumptions:

I'd expect the audio player to have no height unless you put something in it.
I'd also expect that audio themes are probably using one slot, not needing the top/middle/bottom that a video UI needs.
So then it would seem that you could still put audio controls in `bottom-chrome` (whether the default or not), and even though it's "bottom" wouldn't make a difference, because you're not using the other slots and it's still expanding the height of the controller/container as expected.

heff:
The layout of custom media elements appears to be broken. Sorry, I considered pushing this to a later issue but we should really fix this before a release.

https://media-chrome-err0093s8-mux.vercel.app/examples/media-elements/hls.html
https://media-chrome-err0093s8-mux.vercel.app/examples/media-elements/dash.html

image

heff:
Our changes have destroyed the media-controller example. 😆 😭 

image


https://media-chrome-err0093s8-mux.vercel.app/examples/control-elements/media-controller.html

heff:
How important is the clip selector for the next release? [because it's broken](https://media-chrome-err0093s8-mux.vercel.app/examples/control-elements/media-clip-selector.html).

Who's ready for some automated tests?! 😄 ✋

cjpillsbury:
> Our changes have destroyed the media-controller example. 😆 😭

Hah this was another example of some "presumptuous styling under the hood" instead of "less presumptuous styling under the hood". Before I was always applying `align-items: stretch`. I've updated it to apply "align-items: start" by default (which will be more in line with what people would generally expect) and instead added an "align-self: stretch" for slotted `<media-control-bar/>`s. Will be pushing soon.

cjpillsbury:
> The layout of custom media elements appears to be broken. Sorry, I considered pushing this to a later issue but we should really fix this before a release.

Ok I've got a "fix" that resolves the issue, but I did some digging, and this is because the implementation of `<dash-video/>` and `<hls-video/>` are written in such a way that they don't have intrinsic dimensions (unlike `<video/>` or `<mux-video/>`. I think we should expect implementors to not break the intrinsic dimensions of their media tags, but this is something we can treat as a followup and can discuss in our weekly.

heff (approved):
> are written in such a way that they don't have intrinsic dimensions

Ah got it, yeah we can move towards that with those elements. Thanks for the fix in the mean time.

Issue #60, opened 2021-09-09
Show/Hide Subtitle/CC Text Track Events


cjpillsbury:
Q: What should events to show/hide subtitle/cc text tracks look like?

Assumptions:
* Media Chrome Elements will receive attributes in the form proposed in #59 
* Elements will very likely need to parse these back into JS objects resembling a (partial) [`TextTrack`](https://developer.mozilla.org/en-US/docs/Web/API/TextTrack)
* It's common to have a single control element for showing/hiding both Subtitles & CC
* While uncommon, there are cases where users/players would want multiple Subtitles/CCs showing simultaneously.
* It's common to have a UI that "disables"/hides any and all currently showing Subtitles & CC

## Proposal
Create 2 generic events: one for showing and one for hiding 1+ subtitles/cc text tracks (or both)

### Show Text Tracks
* `type`: `"mediashowtexttracksrequest"`
* `detail` (required): `[{ kind: TEXT_TRACK_KIND, label?: OPTIONAL_TRACK_LABEL, language: TRACK_LANGUAGE }, ...]` || `{ kind: TEXT_TRACK_KIND, label?: OPTIONAL_TRACK_LABEL, language: TRACK_LANGUAGE }`

**Result**: changes any matching `TextTrack` to `mode="showing"` (NOTE: In this proposal, this would _**not**_ change the `mode` of any currently `"showing"` `TextTrack`)

**Examples**: 
```js
// Example 1
const track = { kind: "captions", language: "en_US", label: "English (US)" };
const evt1 = new CustomEvent("mediashowtexttracksrequest", { composed: true, bubbles: true, detail: track });
this.dispatchEvent(evt1);

// Example 2 (multi-select)
const track1 = { kind: "captions", language: "en_US" }; // Assumes this caption track has no `label`
const track2 = { kind: "subtitles", label: "English (with descriptions)", language: "en_US" };
const evt2 = new CustomEvent("mediashowtexttracksrequest", { composed: true, bubbles: true, detail: [track1, track2] });
this.dispatchEvent(evt2);
```

### Hide Text Tracks
* `type`: `"mediahidetexttracksrequest"`
* `detail` (optional): `[{ kind: TEXT_TRACK_KIND, label?: OPTIONAL_TRACK_LABEL, language: TRACK_LANGUAGE }, ...]` || `{ kind: TEXT_TRACK_KIND, label?: OPTIONAL_TRACK_LABEL, language: TRACK_LANGUAGE }` || `undefined`

[...]

PR #76, opened 2021-10-14
Feat/cc button simple

cjpillsbury:
* Adds initial support for closed captions via a `<media-captions-button/>` that turns on/off captions.
  * Supports subtitles fallback by default, which can be disabled via `no-subtitles-fallback` attribute
* Implements state management in `<media-controller/>` per spec discussion in #60 (only for "showing"/"disabled" modes to support use case and expected behavior of use case per conversation)
* Updates examples to demonstrate usage.


src/js/utils/captions.js
```diff
@@ -0,0 +1,89 @@
+// NOTE: This is generic for any CSS/html list representation. Consider renaming and moving to generic module.
```
cjpillsbury, line 1: Since these functions use "functional programming paradigms" (such as higher order functions and some branching polymorphism), I'll plan on adding some more code comments/descriptions here before final merge.

heff (commented):
Still working my way through it but threw out an idea worth covering sooner than later around defaulting to the fallback.

examples/control-elements/media-captions-button.html
```diff
@@ -0,0 +1,35 @@
...
+
+<h3>Default (no captions or subtitles)</h3>
+<media-captions-button></media-captions-button>
+
+<h3>Closed Captions on ("captions showing")</h3>
+<media-captions-button media-captions-showing="en:English%20Closed%20Captions" media-captions-list="en:English%20Closed%20Captions"></media-captions-button>
+<!-- 
+<h3>Closed Captions on with subtitles fallback ("subtitles showing")</h3>
+<media-captions-button 
+  use-subtitles-fallback
```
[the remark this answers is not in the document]
dylanjha: ➕ I went back and forth on this, and ultimately I agree to make `use-subtitles-fallback` the default.

Even though I like that the currently default is more technically & semantically correct, in a practical sense I think it hurts us more than it helps.

As a matter of fact, I tripped up on this with Demuxed because the Demuxed captions track on the live stream was being incorrectly marked as subtitles (despite the fact that the manifest had the attributes that should have presented them as captions ). I need to follow up on that, actually. Not sure if it was an Hls.js bug or that the manifest was incorrect.

So yes, from personal experience if `use-subtitles-fallback` was the default it would have been easier for me and I imagine other people would get tripped up by that too.

src/js/media-captions-button.js
```diff
@@ -0,0 +1,199 @@
...
+
+  handleClick(_e) {
+    const ccIsOn = isCCOn(this);
+    if (ccIsOn) {
+      const captionsShowingStr = this.getAttribute(
+        MediaUIAttributes.MEDIA_CAPTIONS_SHOWING
+      );
+      if (captionsShowingStr) {
+        const evt = new window.CustomEvent(
+          MediaUIEvents.MEDIA_DISABLE_CAPTIONS_REQUEST,
```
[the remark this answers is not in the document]
dylanjha: I was wondering this same thing 👍🏼

src/js/utils/captions.js
```diff
@@ -0,0 +1,89 @@
...
+  // 1. Filter to only include tracks to update
+  // 2. Update each of those tracks to the appropriate mode.
+  Array.from(tracks)
+    .filter(isTrackToUpdate)
+    .forEach((textTrack) => {
+      textTrack.mode = mode;
+    });
+};
+
+export const getTextTracksList = (media, filterPredOrObj = () => true) => {
```
dylanjha, line 80: Can we re-use the `getTextTracksList` util to get the "thumbnail" text tracks in `media-thumbnail-preview`?

```js
getTextTracksList(media, { kind: "thumbnails"})
```

P.S., I really like the interface for this util function^ that makes a ton of sense, but I am having a little trouble following the individual pieces of code through the multiple levels of meta-programming, `textTrackObjAsPred` and `isMatchingPropOf` are twisting my brain.

If I had to change any of this I would be scared, (some unit tests covering these utils would help a lot)

src/js/utils/captions.js
```diff
@@ -0,0 +1,89 @@
...
+
+export const formatTextTracks = (textTracks = []) => {
+  return Array.prototype.map.call(textTracks, formatTextTrackObj).join(' ');
+};
+
+// NOTE: This is a generic higher order fn. Consider and moving to generic module.
+export const isMatchingPropOf = (key, value) => (obj) => obj[key] === value;
+// NOTE: This is a generic higher order fn. Consider renaming and moving to generic module.
+export const textTrackObjAsPred = (filterObj) => {
+  const preds = Object.entries(filterObj).map(([key, value]) => {
```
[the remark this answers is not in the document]
cjpillsbury: Exactly correct, @dylanjha.

src/js/constants.js
```diff
@@ -10,6 +10,12 @@ export const MediaUIEvents = {
     MEDIA_PREVIEW_REQUEST: 'mediapreviewrequest',
     MEDIA_ENTER_PIP_REQUEST: 'mediaenterpiprequest',
     MEDIA_EXIT_PIP_REQUEST: 'mediaexitpiprequest',
+    MEDIA_SHOW_TEXT_TRACKS_REQUEST: 'mediashowtexttracksrequest',
+    MEDIA_HIDE_TEXT_TRACKS_REQUEST: 'mediahidetexttracksrequest',
+    MEDIA_SHOW_CAPTIONS_REQUEST: 'mediashowcaptionsrequest',
+    MEDIA_SHOW_SUBTITLES_REQUEST: 'mediashowsubtitlesrequest',
+    MEDIA_DISABLE_CAPTIONS_REQUEST: 'mediadisablecaptionsrequest',
+    MEDIA_DISABLE_SUBTITLE_REQUEST: 'mediadisablesubtitlesrequest',
```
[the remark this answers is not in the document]
cjpillsbury: ah yeah that's one of the fixes that made it into our demuxed branch that I missed 😬. Will re-add to this branch. Good catch.

src/js/utils/captions.js
cjpillsbury: Awesome. Code comments + unit tests sound 👌.

src/js/utils/captions.js
```diff
@@ -0,0 +1,89 @@
...
+  }
+  return [trackOrTracks];
+};
+
+export const formatTextTrackObj = ({ label, language } = {}) => {
+  if (!label) return language;
+  return `${language}:${encodeURIComponent(label)}`;
+};
+
+export const formatTextTracks = (textTracks = []) => {
```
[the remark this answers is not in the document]
heff: Cool. I only have one example here, but if the assumption is "people will just understand what this does so we don't need to add more context in the name", I'm not a fan.

src/js/media-captions-button.js
```diff
@@ -0,0 +1,207 @@
...
+          // If we have at least one subtitles track (and didn't have any captions tracks), request for the first one to be showing as a fallback for captions.
+          const evt = new window.CustomEvent(
+            MediaUIEvents.MEDIA_SHOW_SUBTITLES_REQUEST,
+            { composed: true, bubbles: true, detail: subTrackStr }
+          );
+          this.dispatchEvent(evt);
+        }
+      }
+      else {
+        // If we end up here, it means we have an enabled CC-button that a user has clicked on but there are no captions and no subtitles (or we've disabled subtitles fallback).
```
heff, line 198: I don't think we need to solve this in this PR, but I'm wondering if this needs to be elevated to the end-user's view. Otherwise they could just be clicking this button and seeing no result. Ideally the UI isn't even showing the CC button if in this case, so we should probably at minimum give an example of how to do that.

src/js/media-controller.js
```diff
@@ -124,26 +125,50 @@ class MediaController extends MediaContainer {
...
-          if (cue) {
-            const url = new URL(cue.text);
-            const previewCoordsStr = new URLSearchParams(url.hash).get('#xywh');
-            this.propagateMediaState(MediaUIAttributes.MEDIA_PREVIEW_IMAGE, url.href);
-            this.propagateMediaState(MediaUIAttributes.MEDIA_PREVIEW_COORDS, previewCoordsStr.split(',').join(' '));
-          }
-        }
-      }
+        // Since this isn't really "global state", we may want to consider moving this "down" to the component level,
+        // probably pulled out into its own module/set of functions (CJP)
```
heff, line 143: Maybe, but I can think of at least two components that would use this info.

src/js/media-controller.js
```diff
@@ -124,26 +125,50 @@ class MediaController extends MediaContainer {
       },
       MEDIA_PREVIEW_REQUEST: (e) => {
         const media = this.media;
+        // No media (yet), so bail early
+        if (!media) return;
+
+        
+        const [track] = getTextTracksList(media, { kind: TextTrackKinds.METADATA, label: 'thumbnails' });
```
heff, line 132: Nice refactor

heff (approved):
Awesome 👍

PR #78, opened 2021-10-20
Chore/testing infrastructure

cjpillsbury:
This PR adds the basic infrastructure for unit/integration testing of web components/custom elements as well as JS modules, including:
* basic dependencies and scripts for running and writing unit+integration tests
* a non-exhaustive example of writing tests for a web component/custom element (media-play-button) for a POC/reference usage
* a non-exhaustive example of testing an ESM util module (captions) for a POC/reference usage.

## NOTES:
* This PR is not meant to be exhaustive, but rather to provide the baseline infrastructure so other efforts may now include tests. This will also allow for iterations on our testing infrastructure, including baseline example(s) of integration-level tests, multi-browser testing, using with the CI, and git hooks integration.
* For the test runner, we are using Web Test Runner, the [recommended test runner for web components](https://open-wc.org/guides/developing-components/testing/) and @open-wc/testing for web component/custom element-specific testing utils (e.g. fixtures), with all other test dependencies (chai+plugins for assertions, sinon for mocks/spies/stubs) following from these.
  * Ideally, we would be able to use `jest` plus `testing-library` for better assertion & querying semantics, especially for the DOM & web components/custom elements. However, jest currently has some node/JSDOM assumptions that require some [hack solutions](https://github.com/facebook/jest/issues/11065) to work with Web Testing Library, and `testing-library` currently has [no plans for shadowDOM support](https://github.com/testing-library/dom-testing-library/issues/413). If we prefer these semantics, we can iterate to come up with a viable solution.

### DIFF
--- a/__tests__/unit/media-play-button.spec.js
+++ b/__tests__/unit/media-play-button.spec.js
@@ -1,0 +1,138 @@
+import { spy } from 'sinon';
+import { fixture, expect } from "@open-wc/testing";
+import { MediaUIAttributes, MediaUIEvents } from "../../src/js/constants";
+import { verbs } from "../../src/js/labels/labels";
+import "../../src/js/media-play-button";
+
+describe("<media-play-button/>", () => {
+  it("passes the a11y audit", async () => {
+    const el = await fixture(`<media-play-button></media-play-button>`);
+    await expect(el).shadowDom.to.be.accessible();
+  });
+
+  describe('default/playing', () => {
+
+    let el;
+
+    beforeEach(async () => {
+      el = await fixture(`<media-play-button></media-play-button>`);
+    });
+
+    afterEach(() => {
+      el = undefined;
+    });
+
+    it("should expect pause actions", async () => {
+      expect(el.getAttribute('aria-label')).equals(verbs.PAUSE());
+    });
+
+    it("should show pause", () => {
+      const icon = el.shadowRoot.querySelector('slot[name="pause"] > *')
+      expect(icon).to.be.visible;
+    });
+
+    /** @TODO Figure out if this test failure is due to assertion assumptions of visible vs. invalid component initialization condition (CJP) */
+    it.skip("should not show play", () => {
+      const icon = el.shadowRoot.querySelector('slot[name="play"] > *')
+      expect(icon).to.not.be.visible;
+    });
+
+    it("should request pause when clicked", () => {
+      const requestHandler = spy();
+      el.addEventListener(MediaUIEvents.MEDIA_PAUSE_REQUEST, requestHandler);
+      el.click();
+      expect(requestHandler.calledOnce).to.be.true;
+    });
+
+    it("should not request play when clicked", () => {
+      const requestHandler = spy();
+      el.addEventListener(MediaUIEvents.MEDIA_PLAY_REQUEST, requestHandler);
+      el.click();
+      expect(requestHandler.called).to.be.false;
+    });
+
+    it.skip("should request pause when Enter pressed", () => {
+      /** @TODO Implement me (CJP) */
+      expect(false).to.be.true;
+    });
+
+    it.skip("should request pause when Space pressed", () => {
+      /** @TODO Implement me (CJP) */
+      expect(false).to.be.true;
+    });
+
+  });
+
+  describe('paused', () => {
+
+    let el;
+
+    beforeEach(async () => {
+      el = await fixture(`<media-play-button ${MediaUIAttributes.MEDIA_PAUSED}></media-play-button>`);
+    });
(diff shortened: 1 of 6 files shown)


__tests__/unit/media-play-button.spec.js
```diff
@@ -0,0 +1,145 @@
+import { spy } from 'sinon';
```
heff, line 1: Why all the underscores in `__tests__`?

__tests__/unit/media-play-button.spec.js
```diff
@@ -0,0 +1,145 @@
...
+import { fixture, expect } from '@open-wc/testing';
+import { MediaUIAttributes, MediaUIEvents } from '../../src/js/constants';
+import { verbs } from '../../src/js/labels/labels';
+import '../../src/js/media-play-button';
+
+describe('<media-play-button/>', () => {
+  it('passes the a11y audit', async () => {
+    const el = await fixture(`<media-play-button></media-play-button>`);
+    await expect(el).shadowDom.to.be.accessible();
```
heff, line 10: wtf, you can just do `to.be.accessible()`? Wow.

__tests__/unit/media-play-button.spec.js
dylanjha: `__tests__` is a default for Jest, pretty common: https://github.com/facebook/jest/issues/637

__tests__/unit/media-play-button.spec.js
```diff
@@ -0,0 +1,145 @@
...
+    it.skip('should request pause when Enter pressed', () => {
+      /** @TODO Implement me (CJP) */
+      expect(false).to.be.true;
+    });
+
+    it.skip('should request pause when Space pressed', () => {
+      /** @TODO Implement me (CJP) */
+      expect(false).to.be.true;
+    });
+  });
```
dylanjha, line 62: Maybe just remove the ones that are skipped rn? I generally feel like having `@TODO` littered around source + test files is an anti-pattern because they're not very visible and often times never get done.

If it's *truly* a TODO -- file an issue, if not, then just remove the clutter.

That's my opinion, anyway.

__tests__/unit/utils/captions.spec.js
```diff
@@ -0,0 +1,131 @@
...
+        {
+          label: 'English (with descriptions)',
+          kind: 'captions',
+          language: 'en',
+        },
+        { label: 'Spanish', kind: 'subtitles', language: 'es' },
+        { label: 'English', kind: 'subtitles', language: 'en-US' },
+      ]);
+    });
+  });
```
dylanjha, line 97: Having these tests makes understanding the source wayyyy easier, and now I'd feel much more comfortable updating/refactoring any of that source code 👍🏼

package.json
```diff
@@ -13,7 +13,7 @@
     "build": "snowpack build --config snowpack.prod.config.js",
     "dev": "snowpack dev --config snowpack.dev.config.js",
     "start": "yarn dev",
-    "test": "echo \"Error: no test specified\" && exit 1",
+    "test": "web-test-runner --coverage",
```
dylanjha, line 16: Can we add a `.github/workflows/ci.yml` file with something like

```
name: CI

on: [push, pull_request]

jobs:
  build:
    runs-on: ubuntu-latest

    steps:
      - uses: actions/checkout@v2
      - name: Use Node.js ${{ matrix.node-version }}
        uses: actions/setup-node@v1
        with:
          node-version: ${{ matrix.node-version }}
      - run: yarn install --frozen-lockfile
      - run: yarn test
      - run: yarn build
```

Something like that -- goal would just be to make sure tests pass and the build succeeds on every push

dylanjha (changes requested):
Some small suggestions, but overall really happy with this 🙌🏼 

I generally default to Jest, and I see you tried that first, but we're hamstrung by your comment in the notes, so this all looks like the right setup.

If Jest does play nicer with web components world, then it would be nice to transition at some point down the road, only because Jest + testing-library feels like the more "modern" setup. Practically speaking though, this all feels like a great way to have unit test coverage.

cjpillsbury:
> If Jest does play nicer with web components world in the future, then it would be nice to transition at some point down the road, only because Jest + testing-library feels like the more "modern" setup. Practically speaking though, this all feels like a great way to have unit test coverage.

💯 agree.

__tests__/unit/media-play-button.spec.js
cjpillsbury: yeah. it's not perfect, but it's a good audit to have. It's built on axe core (https://www.npmjs.com/package/chai-a11y-axe)

PR #85, opened 2021-10-27
Build/react wrapper

cjpillsbury:
First pass implementation of DIY React wrapper "compiler" that creates React component wrappers for each identified custom element to reduce friction for React devs working with media-chrome.

Also updates `media-chrome` to correctly identify itself as `"type": "module"` per new node/npm ESM support.

### To Build/Use:
* run `yarn build`
* run `yarn build:react`
* confirm modules were generated in `./dist/react/*`

### To test (for the PR before the release):
* Create a react app using e.g. `create-react-app`
* Add `media-chrome` as a dependency
* Clone `media-chrome` if you haven't already, check out this branch (you will have to set up my fork as an additional remote), and `npm link` media-chrome to the newly-created react app project.
* Add the react wrapper implementation of components to your react app and confirm they work as expected. Below is an example use case.

### Some problems it solves:
* Automatically translates boolean props
* Automatically removes undefined props
* Automatically converts style objects into CSS style strings
* Automatically renames react-specific native prop names to the appropriate native attribute equivalents
* More intuitive import usage.
* component names follow standard React component naming conventions.
* resolves #24 via wrapper implementation

### Example Usage demonstrating different "react-expected" behaviors:
```jsx
import styled from "styled-components";
import {
  MediaController,
  MediaControlBar,
  MediaPlayButton,
  MediaSeekBackwardButton,
  MediaSeekForwardButton,
  MediaMuteButton,
  MediaVolumeRange,
  MediaTimeRange,
  MediaTimeDisplay,
  MediaCaptionsButton,
  MediaPlaybackRateButton,
  MediaPipButton,
  MediaFullscreenButton,
} from "media-chrome/dist/react";

const StyledMediaPlayButton = styled(MediaPlayButton)`
  --media-button-icon-height: 24px;
`;

function App() {
  return (
    <MediaController>
      <video
        slot="media"
        muted
        src="https://stream.mux.com/DS00Spx1CV902MCtPj5WknGlR102V5HFkDe/high.mp4"
        crossOrigin=""
      >
        <track
          label="thumbnails"
          default
          kind="metadata"
          crossOrigin=""
          src="https://image.mux.com/DS00Spx1CV902MCtPj5WknGlR102V5HFkDe/storyboard.vtt"
        ></track>
        <track
          label="English"
          kind="captions"
          srcLang="en"
          crossOrigin=""
          src="./vtt/en-cc.vtt"
        ></track>
      </video>
      <MediaControlBar>
        <StyledMediaPlayButton>
[...]

PR #90, opened 2021-11-08
build/esm-package: update media-chrome to use type:module. Update CJS…

cjpillsbury:
… modules to have expected cjs extension. Update build script references.

This issue is also resolved in the react wrapper PR ( #85 ) but issuing as a narrower PR to accelerate merge+release process. After the last release, we were no longer compiling to UMD modules, so we need to make these changes to support using media-chrome in node environments (e.g. next.js) . Once merged, we should cut another release.


dylanjha (approved):
LGTM -- I was running into this problem with Next.js

heff (approved):
Awesome 👍

PR #95, opened 2021-12-01
docs/per-element-docs: Add stubs for relevant elements/components. St…

cjpillsbury:
…art docs for media-captions-button for reference.

Adding as a draft PR for now to make iteration/conversation easier.


dylanjha (approved):
I like this direction 👏🏼

dylanjha (approved):
🙌🏼  this is a good improvement. I think we'll find ways to iterate on what we have here but it's a great starting point

PR #96, opened 2021-12-02
fix/css-var-syntax: fixing nested vars syntax in media-chrome-range.

cjpillsbury:
https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties for reference


dylanjha (approved):
good catch, looks like a syntactical oversight 👌🏼

PR #105, opened 2022-01-04
Add architecture diagrams

luwes:
Docs link: 
- https://github.com/luwes/media-chrome/blob/ui-controller-media-diagram/docs/architecture-diagrams.md

Figma links:
- https://www.figma.com/file/JfpS4VVSJgvywPHEYAr7Ie/Media-Chrome-Diagrams?node-id=0%3A1
- https://www.figma.com/file/JfpS4VVSJgvywPHEYAr7Ie/Media-Chrome-Diagrams?node-id=26%3A84


heff:
These are cool! Is the goal to publish these in the repo?

> In case 2. below the UI element would not be nested under the <media-controller>
What do we mean by "below the UI element" here?

image

You can't use the media-controller attr on a generic div today (unless I missed something). You would need to use the `associateElement` API to register the div with the controller. If the div is a child of the controller or an associated element, it should only need the media-chrome-attribute attr.

luwes:
Added the markdown version 
https://github.com/muxinc/media-chrome/blob/26a41ce799d1ca2ded7e00bd5d71cb9de1587cf7/docs/architecture-diagrams.md

docs/architecture-diagrams.md
```diff
@@ -0,0 +1,45 @@
...
+![media chrome diagram](./assets/media-chrome-diagram-media-controller-nesting.png)
+
+[View Figma embed](https://www.figma.com/embed?embed_host=share&url=https%3A%2F%2Fwww.figma.com%2Ffile%2FJfpS4VVSJgvywPHEYAr7Ie%2FMedia-Chrome-Diagrams%3Fnode-id%3D0%253A1)
+  
+<br>
+<br>
+
+### 2\. by `<media-controller>` ID
+
+![media chrome diagram](./assets/media-chrome-diagram-media-controller-id.png)
```
cjpillsbury, line 43: Fine for now/forever (unless @heff has concerns), but the "Send" arrow from the elements is slightly lying, since it's bubbling to the `media-control-bar` ("associated element") and that's what is getting picked up by the event handlers in the `MediaController` instance. Since you discuss this in the description ☝️, I'm ok with this visual diagram.

docs/architecture-diagrams.md
```diff
@@ -0,0 +1,45 @@
+Media Chrome - Architecture Diagrams
+====================================
+
+Any HTML element can send user input to the `MediaController` and receive media state from the `MediaController`.
+
+### Sending
+
+The `MediaController` receives user input via `MediaUIEvents` like `MediaUIEvents.MEDIA_PLAY_REQUEST` or `MediaUIEvents.MEDIA_SEEK_REQUEST` for example.
```
cjpillsbury, line 8: consider adding 
> The `MediaController` may receive these events in one of two ways:
(leave it up to you for final discretion to add)

docs/architecture-diagrams.md
```diff
@@ -0,0 +1,45 @@
...
+
+For this case an **associated element** is created by targeting the `<media-controller>` via a `[media-controller="my-media-controller"]` attribute. The `<media-controller>` element should have a corresponding id attribute, `[id="my-media-controller"]` in this instance.
+
+All Media Chrome elements support the `[media-controller]` attribute and can be made an **associated element**. Simple HTML elements can be made associated elements but require some JavaScript to get this to work.
+
+### Receiving
+
+The `MediaController` propagates media state by setting `MediaUIAttributes` on observing UI elements.
+
+Any **associated element** or any of its descendants can receive media state from the `MediaController` but they need to be observing one or more `MediaUIAttributes`. The native Media Chrome web components will have this built in and they do this by having the `MediaUIAttributes` listed in the web component `observedAttributes` array.
```
cjpillsbury, line 22: I think we should introduce the term "media state receiver" here. Something like:
> Any **associated element** or any of its descendants can receive media state from the `MediaController`, as long as the elements are identifiable as something that _*should*_ receive media state (aka identifiable as a **media state receiver**). Elements are identified as media state receivers in one of two ways:

Then you can discuss the `observedAttributes` and `[media-chrome-attributes]`

docs/architecture-diagrams.html
```diff
@@ -0,0 +1,70 @@
+<!DOCTYPE html>
```
cjpillsbury, line 1: Fine to keep for now to make iteration easier, but let's plan on deleting this before merging to `#main`

docs/architecture-diagrams.md
```diff
@@ -0,0 +1,45 @@
+Media Chrome - Architecture Diagrams
```
cjpillsbury, line 1: Let's add a link to this doc somewhere in the main `README.md`

cjpillsbury (changes requested):
Some small feedback.

luwes:
new link: https://github.com/muxinc/media-chrome/blob/2a0ef1f99a0cc7ac184f94b13d6ed7f7b845fbb2/docs/architecture-diagrams.md

cjpillsbury (approved):

docs/architecture-diagrams.md
heff: Yeah, it's probably fine for this high level understanding of how the system works. Otherwise it feels like we'd have to break out of the simple send/receive concepts and get into the specifics of emitting events, listeners, calling functions, etc.

heff (approved):
These are great. 👍  Thanks @luwes.

PR #109, opened 2022-01-12
Add youtube-video-element as dependency

luwes:
Not sure if we want to add this but it makes it a lot easier to debug mux-chrome with a NPM linked youtube-video-element.


luwes:
Good call yes, would make more sense in the examples folder. Just keeping this branch alive for quick testing.

PR #113, opened 2022-01-14
Fix type error in unset callback

luwes:
Ran into a type error today that happens when media is removed. Is this old code maybe?


cjpillsbury (approved):
I have a strong suspicion you're correct (antiquated code)

PR #114, opened 2022-01-14
fix/update-youtube-media-element-version: Update our youtube example …

cjpillsbury:
…to use the latest version of youtube-video-element.


luwes (approved):
👍

PR #120, opened 2022-01-23
Fix ios fullscreen

### LINKED ISSUE #110: Make fullscreen work on iOS
There is a potential PR for this fix but it's missing some code I believe.
https://github.com/muxinc/media-chrome/pull/55

For a standard video element media-chrome has to call `.webkitEnterFullScreen()` on the video element itself.
https://developer.apple.com/documentation/webkitjs/htmlvideoelement/1630649-webkitenterfullscreen

heff:
Fixes #110 Closes #55

If element fullscreen APIs don't exist, the media controller will now try APIs on the media element itself.

### DIFF
--- a/src/js/media-controller.js
+++ b/src/js/media-controller.js
@@ -77,17 +77,30 @@ class MediaController extends MediaContainer {
 
       // This current assumes that the media controller is the fullscreen element
       // which may be true in most cases but not all.
-      // The prior version of media-chrome support alt fullscreen elements
+      // The prior version of media-chrome supported alt fullscreen elements
       // and that's something we can work towards here
       MEDIA_ENTER_FULLSCREEN_REQUEST: () => {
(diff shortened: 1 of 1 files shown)


luwes:
Awesome! should there be a `webkitExitFullscreen` too?
https://developer.apple.com/documentation/webkitjs/htmlvideoelement/1629468-webkitexitfullscreen

heff:
I might be missing an edge case, but I think we don't need to change anything there because:
1. Phone iOS fullscreen kills all custom controls, so the iOS native controls will handle it
2. I believe you can call webkitExitFullscreen on the document, and we're [already doing that](https://github.com/muxinc/media-chrome/blob/main/src/js/utils/fullscreenApi.js#L13).

luwes (approved):
LGTM

PR #127, opened 2022-01-24
feat/loading-indicator

### LINKED ISSUE #80: Create loading indicator component
We had this for Demuxed, and want to make sure it gets added back:

```
<media-loading-indicator>
```

For demuxed, we were controlling `is-loading` attribute from the outside, but media-controller should be able to automatically handle the loading state based on `media.readyState`

luwes:
Fixes #80 

A rejig of @cjpillsbury 's `<media-loading-indicator>`.

- Fixed a small bug with the `loading-delay` always being `0s`
- Added a snippet to propagate the `media-loading` state from controller to component.

I have a slight concern that depending on `waiting` and `stalling` is not robust enough.
Are these events always fired? Many iframe players don't expose these events.

Something that could also work is deriving this loading state from the playing state and whether the playhead is moving or not. Any thoughts on this?

### DIFF
--- a/examples/advanced.html
+++ b/examples/advanced.html
@@ -40,19 +40,6 @@
       </media-control-bar>
     </media-controller>
     <div class="examples">
-      <button id="loadingBtn">Toggle Loading</button>
-      <script>
-        const loadingBtn = document.querySelector('#loadingBtn');
-        const loadingEl = document.querySelector('media-loading-indicator');
-        loadingBtn.addEventListener('click', () => {
-          const nextState = !loadingEl.hasAttribute('media-loading');
-          if (nextState) {
-            loadingEl.setAttribute('media-loading', '');
-          } else {
-            loadingEl.removeAttribute('media-loading');
-          }
-        })
-      </script>
       <a href="./">View more examples</a>
     </div>
   </main>
--- a/examples/advanced.html
+++ b/examples/advanced.html
@@ -33,12 +33,26 @@
         <media-time-range></media-time-range>
         <media-time-display show-duration remaining></media-time-display>
         <media-captions-button></media-captions-button>
+        <media-loading-indicator media-loading></media-loading-indicator>
         <media-playback-rate-button></media-playback-rate-button>
         <media-pip-button></media-pip-button>
         <media-fullscreen-button></media-fullscreen-button>
       </media-control-bar>
     </media-controller>
     <div class="examples">
+      <button id="loadingBtn">Toggle Loading</button>
+      <script>
+        const loadingBtn = document.querySelector('#loadingBtn');
+        const loadingEl = document.querySelector('media-loading-indicator');
+        loadingBtn.addEventListener('click', () => {
+          const nextState = !loadingEl.hasAttribute('media-loading');
+          if (nextState) {
+            loadingEl.setAttribute('media-loading', '');
+          } else {
+            loadingEl.removeAttribute('media-loading');
+          }
+        })
+      </script>
       <a href="./">View more examples</a>
     </div>
   </main>
--- a/examples/advanced.html
+++ b/examples/advanced.html
@@ -8,6 +8,10 @@
     .examples {
       margin-top: 20px;
     }
+
+    media-loading-indicator {
+      --media-loading-icon-width: 100px;
+    }
   </style>
 </head>
 <body>
@@ -25,7 +29,7 @@
         <track label="English" kind="captions" srclang="en" src="./vtt/en-cc.vtt"></track>
       </video>
       <div slot="centered-chrome" no-auto-hide>
-        <media-loading-indicator media-loading style="--media-loading-icon-width: 100px;"></media-loading-indicator>
+        <media-loading-indicator media-loading></media-loading-indicator>
       </div>
       <media-control-bar>
         <media-play-button></media-play-button>
--- a/examples/advanced.html
+++ b/examples/advanced.html
@@ -24,6 +24,9 @@
         <track label="thumbnails" default kind="metadata" src="https://image.mux.com/DS00Spx1CV902MCtPj5WknGlR102V5HFkDe/storyboard.vtt"></track>
         <track label="English" kind="captions" srclang="en" src="./vtt/en-cc.vtt"></track>
       </video>
+      <div slot="centered-chrome" no-auto-hide>
+        <media-loading-indicator media-loading style="--media-loading-icon-width: 100px;"></media-loading-indicator>
+      </div>
       <media-control-bar>
         <media-play-button></media-play-button>
         <media-seek-backward-button></media-seek-backward-button>
@@ -33,7 +36,6 @@
         <media-time-range></media-time-range>
         <media-time-display show-duration remaining></media-time-display>
         <media-captions-button></media-captions-button>
-        <media-loading-indicator></media-loading-indicator>
         <media-playback-rate-button></media-playback-rate-button>
         <media-pip-button></media-pip-button>
         <media-fullscreen-button></media-fullscreen-button>
--- a/examples/advanced.html
+++ b/examples/advanced.html
@@ -33,7 +33,7 @@
         <media-time-range></media-time-range>
         <media-time-display show-duration remaining></media-time-display>
         <media-captions-button></media-captions-button>
-        <media-loading-indicator media-loading></media-loading-indicator>
+        <media-loading-indicator></media-loading-indicator>
         <media-playback-rate-button></media-playback-rate-button>
         <media-pip-button></media-pip-button>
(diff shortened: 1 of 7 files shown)


src/js/media-controller.js
```diff
@@ -238,6 +238,10 @@ class MediaController extends MediaContainer {
       },
       'ratechange': () => {
         this.propagateMediaState(MediaUIAttributes.MEDIA_PLAYBACK_RATE, this.media.playbackRate);
+      },
+      'waiting,stalled,playing': () => {
+        const isLoading = !this.media?.paused && this.media?.readyState < 3;
+        this.propagateMediaState(MediaUIAttributes.MEDIA_LOADING, isLoading);
       }
```
luwes, line 245: small snippet for checking loading state is here

src/js/media-controller.js
```diff
@@ -238,6 +238,10 @@ class MediaController extends MediaContainer {
       },
       'ratechange': () => {
         this.propagateMediaState(MediaUIAttributes.MEDIA_PLAYBACK_RATE, this.media.playbackRate);
+      },
+      'waiting,stalled,playing': () => {
+        const isLoading = !this.media?.paused && this.media?.readyState < 3;
```
cjpillsbury, line 243: proposal: Let's not conflate `loading` state with `paused` state. Instead, let's have `<media-loading-indicator/>` accept both `MediaUIAttributes .MEDIA_PAUSED` and `MediaUIAttributes.MEDIA_LOADING` (which it already does) and by default style it to show only when `loading` and not `paused`.

cjpillsbury:
> I have a slight concern that depending on `waiting` and `stalling` is not robust enough. Are these events always fired? Many iframe players don't expose these events.
> 
> Something that could also work is deriving this loading state from the playing state and whether the playhead is moving or not. Any thoughts on this?

I can for sure see potential edge cases. I think it's a much bigger concern to end up with "false positives" (putting the `<media-controller/>` in a loading state/showing the `<media-loading-indicator/>` when we're not actually loading) compared to "false negatives" (failing to put the `<media-controller/>` in a loading state/failing to show the `<media-loading-indicator/>` when we are actually loading), since those edge cases are far more interruptive to usage. I'd recommend adding with the narrow case for now, as it is safer, unless other folks are confident about additional events & props to monitor.

examples/advanced.html
```diff
@@ -33,12 +33,26 @@ <h1>Media Chrome Advanced Video Usage Example</h1>
...
         <media-time-display show-duration remaining></media-time-display>
         <media-captions-button></media-captions-button>
+        <media-loading-indicator></media-loading-indicator>
         <media-playback-rate-button></media-playback-rate-button>
         <media-pip-button></media-pip-button>
         <media-fullscreen-button></media-fullscreen-button>
       </media-control-bar>
     </media-controller>
     <div class="examples">
+      <button id="loadingBtn">Toggle Loading</button>
```
cjpillsbury, line 43: we should update this example to not rely on the javascript snippet and instead only rely on the built-in behavior + CSS. (this was added for testing/iteration as I was implementing, but wasn't expected to be the reference example).

cjpillsbury (changes requested):
A couple of requests, but otherwise good catches/improvements on the initial impl.

examples/advanced.html
```diff
@@ -24,6 +24,9 @@ <h1>Media Chrome Advanced Video Usage Example</h1>
         <track label="thumbnails" default kind="metadata" src="https://image.mux.com/DS00Spx1CV902MCtPj5WknGlR102V5HFkDe/storyboard.vtt"></track>
         <track label="English" kind="captions" srclang="en" src="./vtt/en-cc.vtt"></track>
       </video>
+      <div slot="centered-chrome" no-auto-hide>
+        <media-loading-indicator media-loading style="--media-loading-icon-width: 100px;"></media-loading-indicator>
```
cjpillsbury, line 28: nit: let's move the `style` to the `<style>` tag:

```html
  <style>
    .examples {
      margin-top: 20px;
    }

    media-loading-indicator {
      --media-loading-icon-width: 100px
    }
  </style>
```

src/js/media-controller.js
```diff
@@ -238,6 +238,10 @@ class MediaController extends MediaContainer {
       },
       'ratechange': () => {
         this.propagateMediaState(MediaUIAttributes.MEDIA_PLAYBACK_RATE, this.media.playbackRate);
+      },
+      'waiting,stalled,playing': () => {
```
cjpillsbury, line 242: sorry I think this is inconsistent with a previous round of feedback, but I just realized we have a few more events where we're checking the `readyState` in use in our player efforts. I think we should minimally add these as well:
```waiting,stalled,playing,canplay,loadedmetadata'

**Review by cjpillsbury — CHANGES_REQUESTED** [review]
A couple of small asks, otherwise looks 😍

File: src/js/media-controller.js (hunk 5 above)
Comment (luwes) [reply src/js/media-controller.js:242]: I kinda went trough them and some make less sense to me.

- `playing` makes more sense than `timeupdate` because `timeupdate` will fire many times while the playhead is moving while only one event is needed to hide the spinner when playback is resumed.
- `canplay` fires in a dubious state,  the video might not yet be playing so it would be strange to hide the loading indicator here.
- `loadedmetadata` if this comes way before waiting or stalled and the player is playing yes makes sense then

--- 

related to `timeupdate` and efficiency should media-chrome add a gate to setAttribute if the attribute did not change?
https://github.com/muxinc/media-chrome/blob/f476f82d55dfbd22bc2ee0fcb0cba56107ffca56/src/js/media-controller.js#L508-L520

Comment (cjpillsbury) [reply src/js/media-controller.js:242]: > related to timeupdate and efficiency should media-chrome add a gate to setAttribute if the attribute did not change?

I **_thought_** `attributeChangedCallback` only got invoked if, in fact, that attribute **_changed_**. If that's not the case, I think a gate would be 👌 (less worried about extra `removeAttribute` method calls, though maybe I shouldn't be?)

Comment (cjpillsbury) [reply src/js/media-controller.js:242]: > * `canplay` fires in a dubious state,  the video might not yet be playing so it would be strange to hide the loading indicator here.
> * `loadedmetadata` if this comes way before waiting or stalled and the player is playing yes makes sense then

I specifically suggested adding these because they are directly related to changes in `media.readyState`.

That said, we can always issue a followup PR if you have sufficient concern with adding them. I'll pre-approve and defer to your discretion on whether or not you add any/all of them.

**Review by cjpillsbury — APPROVED** [review]

File: src/js/media-controller.js (hunk 5 above)
Comment (luwes) [reply src/js/media-controller.js:242]: > 

thanks, true let's see if we run into issues where the state is not updating correctly.

double checked to be sure, `attributeChangedCallback` always fires if an attribute is set
https://codesandbox.io/s/strange-golick-c10fz?file=/src/index.js

Comment (heff) [reply src/js/media-controller.js:242]: `stalled` actually shouldn't be here. I know it sounds like it should but it's related to networkState, not readyState. The _network_ can be stalling when the video is still playing. Unfortunate naming choices.

**Review by heff — APPROVED** [review]
This is great. 👍 

A few additional **_no change needed_** comments.
* I can see potential future issues in the the overlap of the term "loading" with network state. I don't think we need to change anything now, but we need to be clear with ourselves that we're talking about "user perceived" loading (starting up and stalling) not network loading.
* It'll be interesting to see how we expose customization of this down the road
* To that point, this is the first time I'm seeing rotation built into the spinner SVG itself. Pretty cool and flexibly.

File: examples/advanced.html [hunk 6]
```diff
@@ -8,6 +8,10 @@
     .examples {
       margin-top: 20px;
     }
+
+    media-loading-indicator {
+      --media-loading-icon-width: 100px;
```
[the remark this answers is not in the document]
luwes: the default icon width of 44px came from having the media-loading-indicator in the control bar on first pass
see here: https://github.com/muxinc/media-chrome/pull/127/commits/cefd0337e6278549087668316022fe36dc03e962#diff-1e408dcff60df12596ea8ba2b3c22d7c01da0eb8f5bd6b9436d7804369fd5ec9R36

I especially reduced that size so it wouldn't expand the height of the control bar. 
https://github.com/muxinc/media-chrome/pull/127/commits/5ff7e82714b1528b530fbc10bd912219d92353ee

but yes in hindsight adding the loading indicator in the controlbar is probably less used.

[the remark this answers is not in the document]
heff: Ah, interesting. I don't hate having the loading spinner icon be default the same size as every other icon. But what we can do is add this CSS to the media-container within the centered-chrome scope, so that in the context of centered-chrome, the loading indicator has a default width of 100px. It wouldn't be the first place we've intentionally styled a specific element in a specific context to create a better default.

PR #129, opened 2022-01-25
fix(playback-rate-button): update rates attr parsing

gkatsev:
Previously, rates was parsing as `1, 2, 3` but it was supposed to be
parsing as an HTML-style list like `1 2 3`. This updates the regex for
the split from `,\s?` to `\s+|,\s*` to support what is documented but
also keep around the legacy behavior.

In addition, the map was changed to use `Number(str)` for converting the
strings into numbers and the sorting was made to be numerical.

### DIFF
--- a/src/js/media-playback-rate-button.js
+++ b/src/js/media-playback-rate-button.js
@@ -45,10 +45,12 @@ class MediaPlaybackRateButton extends MediaChromeButton {
     if (attrName === 'rates') {
       // This will:
       // 1. parse the space-separated attribute string (standard for representing lists as HTML/CSS values) into an array (of strings)
+      //   The current regex has includes `|,\s*` to preserve legacy behavior for lists with commas.
       // 2. convert that list into numbers (including potentially NaN)
       // 3. filter out all NaNs for invalid values
       // 4. sort the array of numbers to ensure the expected toggle-through order for playback rate.
-      const newRates = (newValue ?? '').split(/,\s?/).map(str => str ? +str : Number.NaN).filter(num => !Number.isNaN(num)).sort();
+      const newRates = (newValue ?? '').trim().split(/\s+|,\s*/)
+        .map(str => Number(str)).filter(num => !Number.isNaN(num)).sort((a, b) => a - b);
       this._rates = newRates.length ? newRates : DEFAULT_RATES;
       return;
     }


src/js/media-playback-rate-button.js
```diff
@@ -45,10 +45,12 @@ class MediaPlaybackRateButton extends MediaChromeButton {
     if (attrName === 'rates') {
       // This will:
       // 1. parse the space-separated attribute string (standard for representing lists as HTML/CSS values) into an array (of strings)
+      //   The current regex has includes `|,\s*` to preserve legacy behavior for lists with commas.
       // 2. convert that list into numbers (including potentially NaN)
       // 3. filter out all NaNs for invalid values
       // 4. sort the array of numbers to ensure the expected toggle-through order for playback rate.
-      const newRates = (newValue ?? '').split(/,\s?/).map(str => str ? +str : Number.NaN).filter(num => !Number.isNaN(num)).sort();
+      const newRates = (newValue ?? '').trim().split(/\s+|,\s*/)
+        .map(str => Number(str)).filter(num => !Number.isNaN(num)).sort((a, b) => a - b);
```
cjpillsbury, line 53: nitpick (non-blocker): may want to reintroduce `.map(str => str ? Number(str) : Number.NaN)` to be "more forgiving" and avoid unintentional 0s.
Example:
```js
newValue = "1 , 2"
(newValue ?? '').trim().split(/\s+|,\s*/).map(str => Number(str)).filter(num => !Number.isNaN(num)).sort((a, b) => a - b)
// result: [0, 1, 2]
(newValue ?? '').trim().split(/\s+|,\s*/).map(str => str ? Number(str) : Number.NaN).filter(num => !Number.isNaN(num)).sort((a, b) => a - b) 
// result: [1, 2]

cjpillsbury (approved):
One minor non-blocking request, otherwise 👌

src/js/media-playback-rate-button.js
gkatsev: oops, missed this case. I'll take a look.

gkatsev: I updated the regex to `/\s*,?\s+/`, which should handle the case you mentioned as well. However, if you think it's still better to bring back the old map, I'm happy to do that if you think that would still be prudent.

cjpillsbury: nope works for me.

cjpillsbury (approved):

PR #130, opened 2022-01-25
 docs(playback-rate-button): add example page for playback rate button 

gkatsev:
Depends on #129, but I can rebase this when that one is merged. 0550168cc5b0f3f0d300e1ed0e6ce4894f3fc31e is the main change here.

### DIFF
--- a/examples/control-elements/media-playback-rate-button.html
+++ b/examples/control-elements/media-playback-rate-button.html
@@ -8,13 +8,13 @@
 
 <h1>&lt;media-playback-rate-button&gt;</h1>
 
-<h3>Default</h3>
+<h3>Default (no rates attribute)</h3>
 <media-playback-rate-button></media-playback-rate-button>
 
-<h3>Setting rates</h3>
+<h3>Setting rates (rates="1 2 3")</h3>
 <media-playback-rate-button rates="1 2 3"></media-playback-rate-button>
 
-<h3>Legacy rates (deprecated)</h3>
+<h3>Legacy rates (deprecated) (rates="1, 2, 3")</h3>
 <media-playback-rate-button rates="1, 2, 3"></media-playback-rate-button>
 </body>
 </html>
--- a/examples/control-elements/media-playback-rate-button.html
+++ b/examples/control-elements/media-playback-rate-button.html
@@ -8,13 +8,16 @@
 
 <h1>&lt;media-playback-rate-button&gt;</h1>
 
-<h3>Default (no rates attribute)</h3>
+<h3>Default</h3>
+<p>&lt;media-playback-rate-button&gt;</p>
 <media-playback-rate-button></media-playback-rate-button>
 
-<h3>Setting rates (rates="1 2 3")</h3>
+<h3>Setting rates</h3>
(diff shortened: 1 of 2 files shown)


examples/control-elements/media-playback-rate-button.html
```diff
@@ -0,0 +1,20 @@
...
+</head>
+<body>
+
+<h1>&lt;media-playback-rate-button&gt;</h1>
+
+<h3>Default</h3>
+<media-playback-rate-button></media-playback-rate-button>
+
+<h3>Setting rates</h3>
+<media-playback-rate-button rates="1 2 3"></media-playback-rate-button>
```
cjpillsbury, line 15: I think these examples may be confusing without some changes, either adding more context/info to the `<h3>` description (e.g. the values set) and/or demonstrating each example with no value set vs. one of the values set, roughly analogous to our `<media-play-button/>` examples:

```html
<h3>Alternate content</h3>
<media-play-button media-paused>
  <span slot="play">Play</span>
  <span slot="pause">Pause</span>
</media-play-button>
<media-play-button>
  <span slot="play">Play</span>
  <span slot="pause">Pause</span>
</media-play-button>
```

cjpillsbury (changes requested):
Small ask. Otherwise, 👌

examples/control-elements/media-playback-rate-button.html
gkatsev: That makes sense. Would it make sense to hook up a media-controller in here so that it's interactable?

gkatsev: So, I was trying to do an example where the rates set are `2 3 4`, but that uncovered another question. Should there be a `rate` attribute too? Since right now, the default rate is 1 and it will always go to one at first.
Though, maybe that will also be settled if it gets a more complete example with a `<video>` and a `<media-controller>`

cjpillsbury: There is a rate attribute! It's what's automagically set by `media-controller` on rate changes on the slotted media element:

`MediaUIAttributes.MEDIA_PLAYBACK_RATE` / `media-playback-rate`. Example:
```html
<media-playback-rate-button media-playback-rate="0.5"></media-playback-rate-button>
```

Note that this would be "overwritten" by whatever the playbackRate is on the media if the component were "wired up" (just like `media-paused` would be in the above example).

heff: I second the "hold off on media-controller" comment. We first want these buttons to stand on their own, and have docs/demos that make it clear how you would use the button if no media-controller existed. This is an interesting case. I could see two main examples, one without a `media-playback-rate` set and one with (both with `rates` set).

If you want to do a simple interactive demo, you could demo a small (and easy to read) click listener that just rotates through the rates values. Or not. Either way, let's get this merged in. 👍

gkatsev: I think that's fine, and we can hold off until there's a more complete set of examples.
I guess the question now is whether what's currently here good enough, or is there still something that needs to change or is missing? (there is a conflict, but likely easy to resolve)

cjpillsbury (approved):

PR #131, opened 2022-01-26
Add prettier pkg and yarn script

luwes:
I'm not sure this is wanted but added it just in case. 

Right now there is Prettier config but it depends on the Prettier pkg to be installed globally on the dev's computer,
and it seems Prettier wasn't used much so far 😅 

I'll leave it to someone else to run the format script and push the changes if we want to do that.
Run locally this is the git diff.

SCR-20220125-pwi

Prettier also found some HTML errors 🤠 

SCR-20220125-q2o


mmcc:
image

heff:
lol, ouch. Ok so just to confirm, someone just needs to merge this, run prettier, and push the (many) automatically added changes. Yeah? Looks like we should also build it into the dev and build scripts.

mmcc:
My vote would be to approve this, then run it and commit the gigantic change before merging this PR.

heff:
Ok so, approved. 👍  Now who takes it? I think if you want to run with the next step and add it to this PR, go for it @luwes. Get those commit points.

README.md
```diff
@@ -145,41 +161,43 @@ Use HTML to add or remove any of the controls. Then you can use CSS to style the
...
-| `<media-playback-rate-button>` | Change the speed of playback                                                                       |
-| `<media-seek-backward-button>` | Jump back 30 seconds in the media                                                                      |
-| `<media-seek-forward-button>` | Jump ahead 30 seconds in the media                                                                  |
-| `<media-captions-button>` | Show/disable captions (if no captions are available, will fallback to subtitles by default unless `no-subtitles-fallback` attribute is set)                                                                      |
-| `<media-airplay-button>`    | Bring up the AirPlay menu to select/deselect AirPlay playback (Safari only).                          |
-| More to come                 | Requests and contributions welcome                                                                   |
+| Element                        | Description                                                                                                                                                                                                 |
```
heff, line 164: Well this is a nice feature!

heff (approved):
Looks good. There's just no way I can review every line, or feel the need to. YOLO

PR #133, opened 2022-01-26
Feat/media airplay button

cjpillsbury:
### What this PR includes
* implementation of a `<media-airplay-button/>` and corresponding `<media-controller/>` request event management
* examples and docs

### What this PR does not include
* support or availability state management and default styling (core implementation should be straightforward but decisions need to be made on implementation details/architecture)

NOTE:
I've included the `<media-airplay-button/>` in our advanced example for testing purposes (as a separate commit). We can remove it before merging to main if we want to hold off on integrating it into full examples until after we add support/availability state management.


docs/media-airplay-button.md
```diff
@@ -0,0 +1,28 @@
+# `<media-seek-backward-button/>`
```
heff, line 1: Name update needed here

docs/media-airplay-button.md
```diff
@@ -0,0 +1,28 @@
...
+
+# Attributes
+
+_None_
+
+# Slots
+
+| Name      | Default Type | Description                                              |
+| --------- | ------------ | -------------------------------------------------------- |
+| `airplay` | `svg`        | The element shown for the AirPlay button's display |
```
heff, line 16: No change needed: Seeing the slot name in this context makes me thinks these button slots could use more explicit names. But that might be overthinking it.

src/js/media-airplay-button.js
```diff
@@ -0,0 +1,49 @@
...
+  constructor(options = {}) {
+    super({ slotTemplate, ...options });
+  }
+
+  connectedCallback() {
+    this.setAttribute('aria-label', verbs.SELECT_AIRPLAY());
+    super.connectedCallback();
+  }
+
+  attributeChangedCallback(attrName, oldValue, newValue) {
```
heff, line 34: This isn't needed if just passing to super, right?

src/js/labels/labels.js
```diff
@@ -16,6 +16,7 @@ export const verbs = {
   PAUSE: () => 'pause',
   MUTE: () => 'mute',
   UNMUTE: () => 'unmute',
+  SELECT_AIRPLAY: () => 'select air play',
```
heff, line 19: In the text itself, is it more accurate to say "select airplay device"?

src/js/constants.js
```diff
@@ -5,6 +5,7 @@ export const MediaUIEvents = {
     MEDIA_UNMUTE_REQUEST: 'mediaunmuterequest',
     MEDIA_VOLUME_REQUEST: 'mediavolumerequest',
     MEDIA_SEEK_REQUEST: 'mediaseekrequest',
+    MEDIA_SELECT_AIRPLAY_REQUEST: 'mediaselectairplayrequest',
```
heff, line 8: Would it be wrong to just say `MEDIA_AIRPLAY_REQUEST`?

heff (changes requested):
This is great. 👍  One needed doc change.

The way they're handling the menu popup in Safari desktop feels obvious and simple now but I wasn't expecting it. i.e. just placing it wherever the mouse is, rather than trying to align with the button that was clicked. That could be something we consider for media chrome's popup menus when needed (alongside the TV style fullscreen popup).

src/js/labels/labels.js
cjpillsbury: Not sure tbh. One of the options is to "reselect" playback in the browser. If you prefer "select air play device" though, I think it still communicates effectively enough. Just lmk.

src/js/media-airplay-button.js
cjpillsbury: will be needed if/when we add availability/support detection, but I can remove it and re-add it if you prefer.

docs/media-airplay-button.md
cjpillsbury: 👌. Def a copypasta error.

cjpillsbury:
@heff also lmk if you have preferences on including vs. removing the `<media-airplay-button/>` from our advanced example for the time being. I don't feel strongly either way.

src/js/constants.js
heff: Ok cool, makes sense. My S1 is `MEDIA_AIRPLAY_REQUEST`, and `AIRPLAY`, from the stance of the user is attempting to airplay, and doesn't necessarily know there's a select menu coming. And "airplay" itself is a verb if you squint at it. So unless you see something like "there's gonna be an `unselect-airplay`" user action that would make this weird, let's simplify it.

heff:
Yeah, we should (unfortunately) remove it from the example until we solve feature detection. Which we should kick off a discussion around and get to a design at least soon.

docs/media-airplay-button.md
```diff
@@ -0,0 +1,28 @@
+# `<media-airplay-button/>`
+
+Button to jump back 30 seconds in the media.
```
heff, line 3: Description needs updating (I should have  pointed to it specifically before, sorry)

src/js/labels/labels.js
heff: Based on the other comment, let's move this to just `AIRPLAY`.

docs/media-airplay-button.md
cjpillsbury: booooo move fast break docs?

cjpillsbury:
> Yeah, we should (unfortunately) remove it from the example until we solve feature detection. Which we should kick off a discussion around and get to a design at least soon.

I believe these should be covered by #134 and #135 but let me know if you think there are additional considerations that aren't within the scope of those two.

src/js/labels/labels.js
cjpillsbury: cool. I'm going to make it "air play" as it's consumed by screen readers and I don't want the possibility of the robots getting confused about pronunciation. 😄

cjpillsbury: will also rename the function `AIRPLAY` (realized that's what you meant).

heff: 👍

heff:
Awesome, thanks. 👍  After the name changes and doc update this is good to merge.

heff (approved):
Cool. 👍

PR #136, opened 2022-01-26
Add loading indicator example and docs

luwes:
This change adds an example and the docs for the `<media-loading-indicator>` element.
The code is already merged to main:
https://github.com/muxinc/media-chrome/commit/abc07f6a198c1fc7d630caf8ed0c63dcc7872c3b

### DIFF
--- a/examples/control-elements/media-loading-indicator.html
+++ b/examples/control-elements/media-loading-indicator.html
@@ -13,11 +13,11 @@
 
 <h1>&lt;media-loading-indicator&gt;</h1>
 
-<h3>Default (not loading)</h3>
+<h3>Default (hidden by default, media is not buffering)</h3>
 
 <media-loading-indicator></media-loading-indicator>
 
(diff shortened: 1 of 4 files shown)


examples/control-elements/media-loading-indicator.html
```diff
@@ -0,0 +1,30 @@
...
+    media-loading-indicator {
+      --media-icon-color: #f0f;
+    }
+  </style>
+</head>
+<body>
+
+<h1>&lt;media-loading-indicator&gt;</h1>
+
+<h3>Default (not loading)</h3>
```
cjpillsbury, line 16: nit (non-blocking): since this is the only case where our component ends up being auto-hidden based on attributes, maybe add mention of it in the `<h3>` description

cjpillsbury (approved):
One minor ask (non-blocking). Otherwise, 👌 & pre-approved.

PR #146, opened 2022-01-31
improv/icons-cleanup: first pass of icons replacemeent and examples c…

cjpillsbury:
…leanup. Changes to button styles under the hood. ad hoc example for spotify theme use case.


tomByrer:
I noticed the icon heights are different; is this intentional?
image

You can see the Play button has a modest black space on the bottom, the 30sec forward/backward buttons have almost no black underneath, & mute button has quite a bit of black space above & below.

There are different philosophies how sizes should harmonize; literally & visually (by 'weight', so the thin edges of a triangle would extend further than a square).

PR #150, opened 2022-02-03
Feat/configurable seek

### LINKED ISSUE #124: Feature Request: made the seek backward/forward time configurable
Please consider making the seek backward/forward time configurable, presumably through an attribute on the element such as in the following:

```html
<media-seek-backward-button amount="10"></media-seek-backward-button>
```

The amount, 30 seconds, is currently hardcoded as a constant used in the line: https://github.com/muxinc/media-chrome/blob/6c6d8576be6c80cf7f0acebf951763fa42dd21db/src/js/media-seek-backward-button.js#L44

The small visible number in the control bar curved arrow would also need to update to reflect the new seek back amount.

cjpillsbury:
Resolves #124 
Resolves #115

### DIFF
--- a/docs/media-seek-backward-button.md
+++ b/docs/media-seek-backward-button.md
@@ -7,7 +7,11 @@ Button to jump back 30 seconds in the media.
 
 # Attributes
 
-_None_
+# Attributes
+
+| Name          | Type     | Default Value | Description                                                         |
+| ------------- | -------- | ------------- | ------------------------------------------------------------------- |
+| `seek-offset` | `number` | `30`          | Adjusts how much time (in seconds) the playhead should seek forward |
 
 # Slots
 
--- a/src/js/media-seek-forward-button.js
+++ b/src/js/media-seek-forward-button.js
@@ -10,7 +10,8 @@ import { getSlotted, updateIconText } from './utils/element-utils.js';
 
 const DEFAULT_SEEK_OFFSET = 30;
 
-const forwardIcon = `<svg xmlns="http://www.w3.org/2000/svg" aria-hidden="true" viewBox="0 0 15 18"><defs><style>.text{font-size:8px;font-family:Arial-BoldMT, Arial;font-weight:700;}</style></defs><text id="value" class="text" transform="translate(5.9 17.6)">${DEFAULT_SEEK_OFFSET}</text><path d="M7,3V0l5.61,4L7,7.94V5A5.54,5.54,0,0,0,5.1,15.48V17.6A7.5,7.5,0,0,1,7,3Z"/></svg>`;
+// const forwardIcon = `<svg xmlns="http://www.w3.org/2000/svg" aria-hidden="true" viewBox="0 0 15 18"><defs><style>.text{font-size:8px;font-family:Arial-BoldMT, Arial;font-weight:700;}</style></defs><text class="text value" transform="translate(5.9 17.6)">${DEFAULT_SEEK_OFFSET}</text><path d="M7,3V0l5.61,4L7,7.94V5A5.54,5.54,0,0,0,5.1,15.48V17.6A7.5,7.5,0,0,1,7,3Z"/></svg>`;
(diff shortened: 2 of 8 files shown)


cjpillsbury:
This is still using the updated icons, so sizing will be off, but wanted to make this available for review earlier rather than later since updates should (mostly) be simply swapping out the svgs.

src/js/utils/element-utils.js
```diff
@@ -0,0 +1,12 @@
+export const updateIconText = (svg, value, selector = '#value') => {
```
luwes, line 1: this id `#value` should be more unique in the case we provide custom icons in the light DOM then there are 2 elements with the same id in the document. would using a class selector or even a tag name (`text`) selector be enough here?

src/js/utils/element-utils.js
```diff
@@ -0,0 +1,12 @@
...
+  svg.querySelector(selector).textContent = value;
```
luwes, line 2:
```suggestion
  const textNode = svg.querySelector(selector);
  if (textNode) {
    textNode.textContent = value;
  }
```

src/js/utils/element-utils.js
cjpillsbury: Good callout. I _*thought*_ internal SVG nodes' ids only had to be unique internal to the SVG, but I'm not 100% confident on that, so might as well change to a class selector just in case (intentionally wrote it as a selector to allow for the flexibility.

docs/media-seek-backward-button.md
```diff
@@ -7,7 +7,11 @@ Button to jump back 30 seconds in the media.
 
 # Attributes
 
-_None_
+# Attributes
+
+| Name          | Type     | Default Value | Description                                                         |
+| ------------- | -------- | ------------- | ------------------------------------------------------------------- |
+| `seek-offset` | `number` | `30`          | Adjusts how much time (in seconds) the playhead should seek forward |
```
heff, line 14: If this is always seconds, should we just call it `seek-seconds`?

I started looking into a frame advancing button and considered building it on seek forward/backward, which would affect this, but I ultimately felt like that deserved its own button, at least because the default icon would be different.

luwes (approved):

src/js/media-seek-forward-button.js
```diff
@@ -6,28 +6,40 @@ import {
...
+const DEFAULT_SEEK_OFFSET = 30;
+
+// const forwardIcon = `<svg xmlns="http://www.w3.org/2000/svg" aria-hidden="true" viewBox="0 0 15 18"><defs><style>.text{font-size:8px;font-family:Arial-BoldMT, Arial;font-weight:700;}</style></defs><text class="text value" transform="translate(5.9 17.6)">${DEFAULT_SEEK_OFFSET}</text><path d="M7,3V0l5.61,4L7,7.94V5A5.54,5.54,0,0,0,5.1,15.48V17.6A7.5,7.5,0,0,1,7,3Z"/></svg>`;
+const forwardIcon = `<svg xmlns="http://www.w3.org/2000/svg" aria-hidden="true" viewBox="-3 -3 24 24"><defs><style>.text{font-size:8px;font-family:Arial-BoldMT, Arial;font-weight:700;}</style></defs><text class="text value" transform="translate(7.4 17.6)">${DEFAULT_SEEK_OFFSET}</text><path d="M8.52,3V0l5.61,4-5.61,4V5A5.54,5.54,0,0,0,6.6,15.48V17.6A7.5,7.5,0,0,1,8.52,3Z"/></svg>`;
```
heff, line 14: There could be an approach here using a svg [foreignObject](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/foreignObject) and auto setting the `content` using the attr() value. Avoiding javascript. I don't know for sure that'd work or add any value, but it sounds cool. :)

heff (approved):
Cool. 👍 No critical comments.

src/js/media-seek-forward-button.js
cjpillsbury: I think I'm leaning toward keeping it `<text>` to make it more sympatico with svg editors like illustrator.

docs/media-seek-backward-button.md
cjpillsbury: I'm gonna keep it as is. Most of the properties/attrs don't bake in units in their names in DOM-land, so this feels more consistent with those norms/conventions.

PR #151, opened 2022-02-03
Use css prop background instead background-color

### LINKED ISSUE #141: CSS background vars
Not a biggie but in media-chrome I noticed a CSS var `--media-control-background` for the buttons and it’s set to the `background-color` property. 

In a bunch of places in the themes etc this var is set to `none` but strictly that’s not valid CSS. 
Should be `transparent`.
https://stackoverflow.com/a/8739704/268820

Maybe look into setting this CSS prop to the CSS shorthand `background` for more customization options.

luwes:
Fixes #141


cjpillsbury (approved):
Assuming there aren't any additional updates required in our examples, LGTM! 🚀

PR #154, opened 2022-02-04
Add fixes for PiP and Fullscreen

luwes:
This change adds some fixes for PiP and fullscreen when media-chrome itself is part of a shadow DOM.

`this.getRootNode();` would then point to the shadow root instead of the document which caused some issues.
Some properties and methods of both PiP and fullscreen always have to be accessed on the `document` according to the docs:

https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API
https://developer.mozilla.org/en-US/docs/Web/API/Picture-in-Picture_API

- [x] Test on mobile when preview is ready
- [x] Test no issues come up with youtube-video-element. This does not support PiP.

### DIFF
--- a/src/js/media-controller.js
+++ b/src/js/media-controller.js
@@ -94,18 +94,18 @@ class MediaController extends MediaContainer {
       //
       // Generally the fullscreen and PiP API's have the exit methods and enabled
       // flags on the `document`. The active element is accessed on the document
-      // or shadow root. Entering fullscreen or PiP is called on the element. i.e.
+      // or shadow root but Safari doesn't support this.
+      // Entering fullscreen or PiP is called on the element. i.e.
       //
       //   - Document.exitFullscreen()
       //   - Document.fullscreenEnabled
-      //   - Document.fullscreenElement / ShadowRoot.fullscreenElement
+      //   - Document.fullscreenElement / (ShadowRoot.fullscreenElement)
       //   - Element.requestFullscreen()
       //
       MEDIA_ENTER_FULLSCREEN_REQUEST: () => {
-        const docOrRoot = this.getRootNode();
         const media = this.media;
 
-        if (docOrRoot.pictureInPictureElement) {
+        if (document.pictureInPictureElement) {
           // Should be async
           document.exitPictureInPicture();
         }
@@ -128,22 +128,19 @@ class MediaController extends MediaContainer {
         document[fullscreenApi.exit]();
       },
       MEDIA_ENTER_PIP_REQUEST: () => {
-        const docOrRoot = this.getRootNode();
         const media = this.media;
 
         if (!document.pictureInPictureEnabled) return;
 
         // Exit fullscreen if needed
-        if (docOrRoot[fullscreenApi.element]) {
+        if (document[fullscreenApi.element]) {
           document[fullscreenApi.exit]();
         }
 
         media.requestPictureInPicture();
       },
       MEDIA_EXIT_PIP_REQUEST: () => {
-        const docOrRoot = this.getRootNode();
-
-        if (docOrRoot.pictureInPictureElement) {
+        if (document.pictureInPictureElement) {
           // Should be async
           document.exitPictureInPicture();
         }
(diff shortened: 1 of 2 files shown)


src/js/media-controller.js
```diff
@@ -90,14 +90,24 @@ class MediaController extends MediaContainer {
...
+      //   - Document.exitFullscreen()
+      //   - Document.fullscreenEnabled
+      //   - Document.fullscreenElement / ShadowRoot.fullscreenElement
+      //   - Element.requestFullscreen()
+      //
       MEDIA_ENTER_FULLSCREEN_REQUEST: () => {
         const docOrRoot = this.getRootNode();
         const media = this.media;
 
         if (docOrRoot.pictureInPictureElement) {
```
heff, line 108: Ok so understanding which el is pip needs the shadow root when in the shadow dom but exiting always requires the document. Confusing.

src/js/media-controller.js
```diff
@@ -90,14 +90,24 @@ class MediaController extends MediaContainer {
...
           // Should be async
-          docOrRoot.exitPictureInPicture();
+          document.exitPictureInPicture();
         }
 
         if (super[fullscreenApi.enter]) {
```
heff, line 113: We can delete the "Shadow root throws an error" comment now that we understand the landscape.

heff (commented):
In general it looks good to me. The testing tasks are a good idea. (I'm not in a place to test it myself right at the moment)

luwes:
exiting fullscreen in Safari doesn't work with this build in the mux-player 😕 
but for standalone media-chrome it does work.

heff:
iOS safari specifically, right? Seems to be working on Safari desktop for me. I wouldn't be surprised if media-is-fullscreen isn't set when the video element itself goes fullscreen on iOS.

Is that a blocker for merging this?

src/js/media-controller.js
```diff
@@ -659,7 +667,7 @@ class MediaController extends MediaContainer {
 
   get isPip() {
     return (
-      this.media && this.media == this.getRootNode().pictureInPictureElement
+      this.media && this.media == document.pictureInPictureElement
```
heff, line 670: Just to be sure, did you test this and the fullscreen version in browsers other than Safari, with MC in the shadow dom? I have a suspicion this doesn't or shouldn't work. It's a bit of a leaky shadow dom if elements in the shadow dom can be exposed on the higher level doc like this.

luwes: what do you think of just checking `!!document.pictureInPictureElement` if PiP is enabled?

as far as I know there's no browsers that allow more than 1 at a time.
https://www.w3.org/TR/picture-in-picture/#one-pip-window

heff: Oh interesting, if contains() can pierce the shadow dom, awesome! I guess if it's `open` then it should be able to.

src/js/media-controller.js
```diff
@@ -666,9 +667,10 @@ class MediaController extends MediaContainer {
   }
 
   get isPip() {
-    return (
-      this.media && this.media == document.pictureInPictureElement
-    );
+    const pipElement =
+      this.getRootNode().pictureInPictureElement ??
```
heff, line 671: Double check that it doesn't [throw an error](https://github.com/muxinc/media-chrome/pull/154/files#diff-f093c2ba0a65c5a566a4ad86659ffc5bee29bfbbc241f863bed8f634b7a0f9f7L120) to check this on the shadow root in a browser somewhere. The test suite for this change is big... Every browser, with MC both in and out of the shadow dom, both Fullscreen and PiP, entering and exiting. Any of those combinations could need special handling.

src/js/utils/element-utils.js
```diff
@@ -14,3 +14,9 @@ export const getAllSlotted = (el, name) => {
 };
 
 export const getSlotted = (el, name) => getAllSlotted(el, name)[0];
+
+export const containsWithShadow = (refNode, otherNode) => {
+  if (!refNode || !otherNode) return false;
+  if (refNode.contains(otherNode)) return true;
+  return containsWithShadow(refNode.shadowRoot, otherNode);
```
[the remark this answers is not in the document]
gkatsev: Seems like the reason why it doesn't is that the shadow root is a Document Fragment, so, technically, the shadow DOM isn't in the main document DOM.
The other approach would've been to iterate through `parentNode` (and not `parentElement` because the DocumentFragment is a node), but this way is better and more targeted to our needs.

src/js/media-controller.js
luwes: I checked Chrome and Safari desktop, and iOS. Didn't see an error being thrown.

heff (approved):
Someday we'll have a test suite we can pull up in different browsers and test all of this quickly, thank for doing the manual work in the mean time. Looks good.

PR #156, opened 2022-02-07
feat/availability-states: Add support for availability detection and state management. Update advanced and mobile examples for demo purposes.

cjpillsbury:
This PR adds basic support for availability/support detection, as discussed in #134 
Currently implemented generic volume and airplay detection, with pending attributes for closed captions. Note there are still some decisions "around the edges" that have yet to be resolved.

Issuing this now to facilitate the conversation, start preliminary review, and potentially merge with the possibility of followup PRs to pin down any outstanding discussion points.

### DIFF
--- a/examples/advanced.html
+++ b/examples/advanced.html
@@ -12,6 +12,15 @@
       media-loading-indicator {
         --media-loading-icon-width: 100px;
       }
+
+      media-airplay-button[media-airplay-unavailable] {
+        display: none;
+      }
+
+      media-volume-range[media-volume-unavailable] {
+        display: none;
+      }
+
     </style>
   </head>
   <body>
@@ -53,6 +62,7 @@
           <media-playback-rate-button></media-playback-rate-button>
           <media-pip-button></media-pip-button>
           <media-fullscreen-button></media-fullscreen-button>
+          <media-airplay-button></media-airplay-button>
         </media-control-bar>
       </media-controller>
       <div class="examples">
--- a/examples/advanced.html
+++ b/examples/advanced.html
@@ -21,6 +21,10 @@
         display: none;
       }
 
+      media-pip-button[media-pip-unavailable] {
+        display: none;
+      }
+
     </style>
   </head>
   <body>
--- a/src/js/constants.js
+++ b/src/js/constants.js
@@ -23,13 +23,18 @@ export const MediaUIEvents = {
 };
 
 export const MediaUIAttributes = {
+  MEDIA_AIRPLAY_UNAVAILABLE: 'media-airplay-unavailable',
   MEDIA_PAUSED: 'media-paused',
   MEDIA_MUTED: 'media-muted',
   MEDIA_VOLUME_LEVEL: 'media-volume-level',
   MEDIA_VOLUME: 'media-volume',
+  MEDIA_VOLUME_UNAVAILABLE: 'media-volume-unavailable',
   MEDIA_IS_PIP: 'media-is-pip',
   MEDIA_CAPTIONS_LIST: 'media-captions-list',
   MEDIA_SUBTITLES_LIST: 'media-subtitles-list',
+  // Include this for styling convenience or exclude since it
+  // can be derived from MEDIA_CAPTIONS_LIST && MEDIA_SUBTITLES_LIST? (CJP)
+  MEDIA_CAPTIONS_UNAVAILABLE: 'media-captions-unavailable',
   MEDIA_CAPTIONS_SHOWING: 'media-captions-showing',
   MEDIA_SUBTITLES_SHOWING: 'media-subtitles-showing',
   MEDIA_IS_FULLSCREEN: 'media-is-fullscreen',
@@ -63,4 +68,9 @@ export const ReadyStates = {
   HAVE_CURRENT_DATA: 2,
   HAVE_FUTURE_DATA: 3,
   HAVE_ENOUGH_DATA: 4,
+};
+
+export const AvailabilityStates = {
+  UNAVAILABLE: 'unavailable',
+  UNSUPPORTED: 'unsupported',
 };
--- a/src/js/constants.js
+++ b/src/js/constants.js
@@ -24,6 +24,7 @@ export const MediaUIEvents = {
 
 export const MediaUIAttributes = {
   MEDIA_AIRPLAY_UNAVAILABLE: 'media-airplay-unavailable',
+  MEDIA_PIP_UNAVAILABLE: 'media-pip-unavailable',
   MEDIA_PAUSED: 'media-paused',
   MEDIA_MUTED: 'media-muted',
   MEDIA_VOLUME_LEVEL: 'media-volume-level',
--- a/src/js/constants.js
+++ b/src/js/constants.js
@@ -33,9 +33,6 @@ export const MediaUIAttributes = {
   MEDIA_IS_PIP: 'media-is-pip',
   MEDIA_CAPTIONS_LIST: 'media-captions-list',
   MEDIA_SUBTITLES_LIST: 'media-subtitles-list',
-  // Include this for styling convenience or exclude since it
-  // can be derived from MEDIA_CAPTIONS_LIST && MEDIA_SUBTITLES_LIST? (CJP)
-  MEDIA_CAPTIONS_UNAVAILABLE: 'media-captions-unavailable',
   MEDIA_CAPTIONS_SHOWING: 'media-captions-showing',
   MEDIA_SUBTITLES_SHOWING: 'media-subtitles-showing',
   MEDIA_IS_FULLSCREEN: 'media-is-fullscreen',
--- a/src/js/media-controller.js
+++ b/src/js/media-controller.js
@@ -21,6 +21,7 @@ import {
   MediaUIAttributes,
   TextTrackKinds,
   TextTrackModes,
+  AvailabilityStates,
 } from './constants.js';
 import {
   stringifyTextTrackList,
@@ -45,6 +46,23 @@ const {
 class MediaController extends MediaContainer {
   constructor() {
     super();
+
+    this._volumeUnavailable;
+    hasVolumeSupportAsync().then(supported => {
+      if (!supported) {
+        this._volumeUnavailable = AvailabilityStates.UNSUPPORTED;
+      }
+      this.propagateMediaState(MediaUIAttributes.MEDIA_VOLUME_UNAVAILABLE, this.volumeUnavailable);
+    });
+
+    this._airplayUnavailable;
+    if (!window.WebKitPlaybackTargetAvailabilityEvent) {
+      this._airplayUnavailable = AvailabilityStates.UNSUPPORTED;
+    }
+
+    // Include this for styling convenience or exclude since it
+    // can be derived from MEDIA_CAPTIONS_LIST && MEDIA_SUBTITLES_LIST? (CJP)
+    // this._captionsUnavailable;
 
     // Track externally associated control elements
     this.mediaStateReceivers = [];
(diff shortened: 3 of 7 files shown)


src/js/media-controller.js
```diff
@@ -880,6 +936,28 @@ const monitorForMediaStateReceivers = (
   return unsubscribe;
 };
 
+let testMediaEl;
+export const getTestMediaEl = (nodeName = "video") => {
+  if (testMediaEl) return testMediaEl;
+  if (typeof window !== "undefined") {
```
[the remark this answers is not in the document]
cjpillsbury: yeah... we need to figure out how we want to handle "SSR meets our polyfills" cases. From an unpolyfilled use case for envs like JSDOM, checking for `window` is common, but since we polyfill, may need to tweak this. to handle our very specific env.

src/js/media-controller.js
```diff
@@ -46,6 +47,23 @@ class MediaController extends MediaContainer {
   constructor() {
     super();
 
+    this._volumeUnavailable;
```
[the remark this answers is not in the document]
cjpillsbury: do you mean the async bug/behavior in mobile Safari? If so, yes. Also, invoking in the constructor to make sure we confirm only where the web component will be instantiated (aka in a browser). We could consider having `_volumeUnavailable` be a module-level variable, since it should be consistent per runtime and across media content, even if we only check it client-side.

luwes:
This all looks good to me. 
It makes sense to wait with the "unknown" state, it might not be needed and it easier to add than remove later.

PiP support for Firefox might also be a good candidate

src/js/media-controller.js
```diff
@@ -880,6 +936,28 @@ const monitorForMediaStateReceivers = (
...
+let testMediaEl;
+export const getTestMediaEl = (nodeName = "video") => {
+  if (testMediaEl) return testMediaEl;
+  if (typeof window !== "undefined") {
+    testMediaEl = document.createElement(nodeName);
+  }
+  return testMediaEl;
+};
+
+export const hasVolumeSupportAsync = async (
```
heff, line 948: We should execute this test immediately and store the results instead of waiting for a media-controller init. There's essentially two stages to the volume test:
1. Does the underlying platform support it (i.e. even `<youtube-video>` will be limited by the results)
2. Does the current media support volume. Even if the platform does, the specific media may not for some reason. We could choose to kick this one down the road, but just calling it out as a possibility.

cjpillsbury: I'll have to test that in an SSR environment, but I get why you're calling that out. Also, all of the async media elements will add complexities here, so I'm glad you're calling that out.

src/js/media-controller.js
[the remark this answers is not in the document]
heff: I think this is related to the comment I just added.

> Also, invoking in the constructor to make sure we confirm only where the web component will be instantiated (aka in a browser)

I'm not totally following here. Is this just related to SSR?

src/js/media-controller.js
cjpillsbury: >  Promise.then()
I think I did in a previous incarnation? But def worth trying again, since I can't recall the details (also I prefer promise.resolve or requestAnimationFrame over setTimeout 0)

src/js/media-controller.js
```diff
@@ -937,27 +943,33 @@ const monitorForMediaStateReceivers = (
...
+export const hasVolumeSupportAsync = async (mediaEl = getTestMediaEl()) => {
   if (!mediaEl) return false;
   const prevVolume = mediaEl.volume;
   mediaEl.volume = prevVolume / 2 + 0.1;
-  return new Promise((resolve, _reject) => {
-    setTimeout(() => {
-      resolve(mediaEl.volume !== prevVolume);
-    }, 0);
+  return Promise.resolve().then(() => {
+    mediaEl.volume !== prevVolume;
```
luwes, line 957: nice, it still works with a Promise

examples/advanced.html
```diff
@@ -12,6 +12,19 @@
       media-loading-indicator {
         --media-loading-icon-width: 100px;
       }
+
+      media-airplay-button[media-airplay-unavailable] {
```
heff, line 16: I'm wondering if we should just go ahead and build this into the control bar. But that may open up other issues so it can wait for a later PR either way.

src/js/constants.js
```diff
@@ -23,13 +23,19 @@ export const MediaUIEvents = {
 };
 
 export const MediaUIAttributes = {
+  MEDIA_AIRPLAY_UNAVAILABLE: 'media-airplay-unavailable',
+  MEDIA_PIP_UNAVAILABLE: 'media-pip-unavailable',
   MEDIA_PAUSED: 'media-paused',
   MEDIA_MUTED: 'media-muted',
   MEDIA_VOLUME_LEVEL: 'media-volume-level',
   MEDIA_VOLUME: 'media-volume',
+  MEDIA_VOLUME_UNAVAILABLE: 'media-volume-unavailable',
```
heff, line 32: I'll say I'm torn between `media-volume-unavailable` and `media-volume-control-unavailable`. The first is nice and simple but not totally accurate, while the second is accurate but more verbose. Interested in other thoughts on that. That said, I lean towards keeping it as the simple version, because as far as the UI is concerned, we can't rely on any of the volume info, even to display the current volume.

src/js/constants.js
```diff
@@ -23,13 +23,19 @@ export const MediaUIEvents = {
...
   MEDIA_IS_PIP: 'media-is-pip',
   MEDIA_CAPTIONS_LIST: 'media-captions-list',
   MEDIA_SUBTITLES_LIST: 'media-subtitles-list',
+  // Include this for styling convenience or exclude since it
+  // can be derived from MEDIA_CAPTIONS_LIST && MEDIA_SUBTITLES_LIST? (CJP)
+  MEDIA_CAPTIONS_UNAVAILABLE: 'media-captions-unavailable',
```
heff, line 38: How would we determine the difference between "unsupported" vs. "unavailable" in this case? Since an empty captions list isn't the same as "unsupported"? Or do we not differentiate?

src/js/media-controller.js
```diff
@@ -301,6 +324,28 @@ class MediaController extends MediaContainer {
...
+        } else if (event?.availability === 'not-available') {
+          this._airplayUnavailable = AvailabilityStates.UNAVAILABLE;
+        }
+        this.propagateMediaState(
+          MediaUIAttributes.MEDIA_AIRPLAY_UNAVAILABLE,
+          this.airplayUnavailable
+        );
+      };
+      // NOTE: only adding this if airplay is supported, in part to avoid unnecessary battery consumption per
+      // Apple docs recommendations (See: https://developer.apple.com/documentation/webkitjs/adding_an_airplay_button_to_your_safari_media_controls)
```
heff, line 342: Do we think there's actually something under the hood that makes this necessary or is the doc writer being overly helpful in how to write javascript?

src/js/media-controller.js
```diff
@@ -937,27 +943,33 @@ const monitorForMediaStateReceivers = (
...
-    setTimeout(() => {
-      resolve(mediaEl.volume !== prevVolume);
-    }, 0);
+  return Promise.resolve().then(() => {
+    mediaEl.volume !== prevVolume;
   });
 };
 
+export const hasPipSupport = (mediaEl = getTestMediaEl()) =>
+  typeof mediaEl?.requestPictureInPicture === 'function';
```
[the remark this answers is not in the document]
heff: We should probably use this inside of the pipUnavailable getter, so that a media-controller with no media can have a smart default, but then if a media that doesn't have the pip API exposed is added it can update correctly.

src/js/media-controller.js
```diff
@@ -880,6 +942,34 @@ const monitorForMediaStateReceivers = (
...
+
+const pipSupported = hasPipSupport();
+
+let volumeSupported;
+const volumeSupportPromise = hasVolumeSupportAsync().then((supported) => {
+  volumeSupported = supported;
+});
+
+const airplaySupported = !!window.WebKitPlaybackTargetAvailabilityEvent;
```
heff, line 971: This also depends on the specific media element having the appropriate API, right?

src/js/constants.js
```diff
@@ -64,3 +70,8 @@ export const ReadyStates = {
   HAVE_FUTURE_DATA: 3,
   HAVE_ENOUGH_DATA: 4,
 };
+
+export const AvailabilityStates = {
```
heff, line 74: Don't hate me... I think I'm convinced we should just separate these into two attributes, -unsupported and -unavailable. Seeing the varying use cases on all sides of the equation has helped.

src/js/constants.js
cjpillsbury: No, we would never be in an "unsupported" case (unless we come up with a scenario that entails "captions can never be used on this platform/playback env").

src/js/media-controller.js
cjpillsbury: No idea... I can speculate that having registered event listeners could cause some kind of polling of the network "under the hood", which could be a larger battery drain than a "normal" event listener. Doesn't seem too wild.

src/js/constants.js
cjpillsbury: All good. Let's discuss tradeoffs in our 1:1.

src/js/constants.js
luwes: I favor `media-volume-unavailable` as well, nice and simple

src/js/constants.js
heff: After a chat we decided to stick with a single attr for now, for the general simplicity of handling both situations at once.

src/js/media-controller.js
```diff
@@ -880,6 +942,34 @@ const monitorForMediaStateReceivers = (
   return unsubscribe;
 };
 
+let testMediaEl;
+export const getTestMediaEl = (nodeName = 'video') => {
```
heff, line 946: The interaction between nodeName and testMediaEl could have unexpected results if you actually need an audio or video el to be returned and one was already created. I would just remove the nodeName option until we know we need it.

gkatsev: Alternatively, make `testMediaEl` a Map keyed off of nodeName (but  likely yagni)

src/js/media-controller.js
heff: In the case of `<youtube-video>` (and similar iframe medias) the platform will support airplay, the media may/will not expose the required `media.webkitShowPlaybackTargetPicker` API. So as implemented, the airplay button will still show itself, but when you click it, you'll [get a console warning today](https://github.com/muxinc/media-chrome/blob/05eee08657a0716c70093be5b3d6a20eb7c8f9f7/src/js/media-controller.js#L224).

We don't need to address that in this PR, but want to make sure we're on the same page as the issue. For a youtube-video + media-chrome use case, I don't think the console warning or just permanently disabling the button is the right answer. However we get there, the button should ultimately be hidden in the same way that it is on platforms that don't support airplay.

heff (approved):
A few comments but generally approved.

luwes:
this should be fixed before merge.
https://github.com/muxinc/media-chrome/pull/156#discussion_r809512582

src/js/media-controller.js
```diff
@@ -880,6 +942,34 @@ const monitorForMediaStateReceivers = (
...
+  testMediaEl = document?.createElement?.(nodeName);
+  return testMediaEl;
+};
+
+export const hasVolumeSupportAsync = async (mediaEl = getTestMediaEl()) => {
+  if (!mediaEl) return false;
+  const prevVolume = mediaEl.volume;
+  mediaEl.volume = prevVolume / 2 + 0.1;
+  return Promise.resolve().then(() => {
+    mediaEl.volume !== prevVolume;
```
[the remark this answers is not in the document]
cjpillsbury: Good catch. Thanks, @luwes! Also adding a resolve value to the other promise in case it ends up being useful.

src/js/media-controller.js
cjpillsbury: Cool good call. Will remove for now.

PR #159, opened 2022-02-08
Add more selective css to prevent overwrites

luwes:
This change only adds the opacity and transition CSS if the element does not have `no-auto-hide`.

This fixes an issue where these non auto hiding elements want to set opacity or transition themselves and it would get overwritten by the CSS in media-container.

Also removed `visibility: visible` in this change as it's unneeded I believe. it's the default.

### DIFF
--- a/src/js/media-container.js
+++ b/src/js/media-container.js
@@ -82,21 +82,24 @@ template.innerHTML = `
       height: 100% !important;
     }
 
-    /* Hide controls when inactive and not paused and not audio */
     ::slotted(:not([slot=media])) {
+      pointer-events: auto;
+    }
+
+    /* Only add these if auto hide is not disabled */
+    ::slotted(:not([slot=media]):not([no-auto-hide])) {
       opacity: 1;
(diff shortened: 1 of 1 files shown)


src/js/media-container.js
```diff
@@ -82,22 +82,25 @@ template.innerHTML = `
       height: 100% !important;
     }
 
-    /* Hide controls when inactive and not paused and not audio */
     ::slotted(:not([slot=media])) {
-      opacity: 1;
-      transition: opacity 0.25s;
-      visibility: visible;
```
luwes, line 89: this line was removed

gkatsev (approved):

luwes:
I ran into the same issue with `pointer-events` now 🤔 , it's also overridden in `<media-dialog>`
SCR-20220208-gby

heff (approved):
Cool. 👍 The pointer events probably need to be handled separately, or at least I can't immediately speak to the impact of removing that and the gesture layer.

PR #160, opened 2022-02-13
Add media-has-played attribute

mmcc:
When working on a custom poster experience I really found myself missing the `videojs-has-started` class. In the Media Chrome world that felt like it could be a `media-has-played` attribute that's `true` forever after the `playing` event fires for the first time. This is proposing a new attribute, so normally I'd go through a discussion first, but since this was so tiny I figured it was ok to put out there for feedback. 

How should I be thinking about testing this?

### DIFF
--- a/src/js/media-controller.js
+++ b/src/js/media-controller.js
@@ -257,6 +257,9 @@ class MediaController extends MediaContainer {
       'play,pause': () => {
         this.propagateMediaState(MediaUIAttributes.MEDIA_PAUSED, this.paused);
       },
+      playing: () => {
+        this.propagateMediaState(MediaUIAttributes.MEDIA_HAS_PLAYED, !this.media?.paused);
+      }, 
       volumechange: () => {
         this.propagateMediaState(MediaUIAttributes.MEDIA_MUTED, this.muted);
         this.propagateMediaState(MediaUIAttributes.MEDIA_VOLUME, this.volume);
--- a/src/js/constants.js
+++ b/src/js/constants.js
@@ -24,6 +24,7 @@ export const MediaUIEvents = {
 
 export const MediaUIAttributes = {
   MEDIA_PAUSED: 'media-paused',
+  MEDIA_HAS_PLAYED: 'media-has-played',
   MEDIA_MUTED: 'media-muted',
   MEDIA_VOLUME_LEVEL: 'media-volume-level',
   MEDIA_VOLUME: 'media-volume',


gkatsev:
I think autoplay scenarios, particularly rejected autoplay, are likely the main scenarios where it could trip it up or have unexpected results.

luwes:
LGTM! I'm curious about this custom poster experience 😃, I've been thinking about this one too.

src/js/media-controller.js
```diff
@@ -257,6 +257,9 @@ class MediaController extends MediaContainer {
       'play,pause': () => {
         this.propagateMediaState(MediaUIAttributes.MEDIA_PAUSED, this.paused);
       },
+      playing: () => {
+        this.propagateMediaState(MediaUIAttributes.MEDIA_HAS_PLAYED, !this.media?.paused);
```
heff, line 261: Any reason we couldn't just set this to `true`?

heff (approved):
Looks good to me.

Either in this in a follow on, we should consider what we want to happen when a new media is loaded, e.g. [resetting](https://github.com/muxinc/media-chrome/blob/main/src/js/media-controller.js#L416). Is this "player as played" or "media has played"? And what are the implications for a playlist?

src/js/media-controller.js
gkatsev: This likely also helps it from being set for failed autoplay, so, 👍🏻

heff: > If you just set it to true it gets incorrectly set during initial setup when everything fires

I'm not sure the logic here is right. If there's no `media`, has-played will be true. That may not be the specific problem.

mmcc: The logic _should_ make sense in that scenario, since this should be functionally the same as `this.media && !this.media.paused`. Unless we want `has-played` to be true in that case?

The "initial setup" thing I'm talking about is that all of these get called for every event during initialization. If the thought is just doing this (adding a `console.log` for the screenshot), this is the behavior I'm seeing when I load the basic example.

```javascript
playing: () => {
  console.log('oh hai, playing event');
  this.propagateMediaState(MediaUIAttributes.MEDIA_HAS_PLAYED, true);
}, 
```

CleanShot 2022-02-14 at 15 11 37

What am I missing?

heff: > The logic should make sense in that scenario, since this should be functionally the same as this.media && !this.media.paused

Console run this: `let foo = {}; !foo.media?.paused`
Equals true, not false, thanks the to the exclamation before everything compared to `this.media && !...`

> every event during initialization

Yeah, that's a good point. Since we're using the event handlers for setting init state, just setting `true` wouldn't work. But we need the right logic there to determine has-played. It should only ever be set to true when there is a media element, that media element has a working paused API, and that paused API is false. I would try using `=== false`.

mmcc: Ahh yeah you're totally right, good catch, probably clearer just to avoid optional chaining altogether there.

mmcc: Addressed in the poster image proposal, but if the element is rejected I'll submit that as a standalone PR. https://github.com/muxinc/media-chrome/pull/161

PR #161, opened 2022-02-15
feat/media-poster-image: Poster image element with an optional placeholder src.

mmcc:
Adds a poster image element. Cool bit is that this also allows for a `placeholder-src` attribute, which could be something like a base64 encoded, inline image. Use-case would be to allow folks to show a low-res, or [blurhash](https://blurha.sh)-type image to be able to immediately render something.


cjpillsbury:
High level: 

- This is 😎 
-  This feels like it should go in one of our well-defined slots or have its own slot, and not the "bottom-chrome" slot.
  - This would change some of the assumptions in the component's styles.
- Any reason we're using background-image instead of an ``? (This is mostly just curiosity about the choice.)
- Any reason we wouldn't want the image/blur-hash elements to be a default slotted element, instead of just "baked in" parts (note they could be both)? (This is a low priority, but could with little effort make the component more versatile)

src/js/media-poster-image.js
```diff
@@ -0,0 +1,70 @@
...
+import { MediaUIAttributes } from './constants.js';
+
+const template = document.createElement('template');
+
+template.innerHTML = `
+  <style>
+    :host {
+      height: 100%;
+      width: 100%;
+      position: relative;
```
luwes, line 15: might need `pointer-events: none;` so the gesture layer can still receive mouse events
https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events

src/js/media-poster-image.js
```diff
@@ -0,0 +1,70 @@
...
+      background-repeat: no-repeat;
+    }
+
+    :host([${MediaUIAttributes.MEDIA_HAS_PLAYED}]) div {
+      display: none;
+    }
+  </style>
+
+  <div part="placeholder"></div>
+  <div part="image"></div>
```
heff, line 35: If we're gonna use parts here, just understand they become part of the API, with all the downsides harder changes in the future.

src/js/media-poster-image.js
```diff
@@ -0,0 +1,70 @@
...
+`;
+
+const setBackgroundImage = (el, image) => {
+  el.style['background-image'] = `url('${image}')`;
+}
+
+class MediaPosterImage extends window.HTMLElement {
+  static get observedAttributes() {
+    return [MediaUIAttributes.MEDIA_HAS_PLAYED, 'placeholder-src', 'src'];
```
heff, line 44: `<image>` has a `lowsrc` attribute already. Any reason we wouldn't just reuse that terminology?

heff (commented):
A few comments but it's looking good.

src/js/media-poster-image.js
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    :host([${MediaUIAttributes.MEDIA_HAS_PLAYED}]) div {
+      display: none;
+    }
+  </style>
+
+  <div part="placeholder"></div>
+  <div part="image"></div>
+`;
```
dylanjha, line 36: @cjpillsbury 

> This feels like it should go in one of our well-defined slots or have its own slot, and not the "bottom-chrome" slot.

Would something like `slot="centered-chrome"` be specified in the template (we wouldn't want to require the user to specify the slot)

Is that what you had in mind?

dylanjha: Also, if we do that, could the user still specify a second `slot="centered-chrome"`, for example:

User wants to use a placeholder and have a play button in the center:

```
<media-controller>
  <media-poster-image></media-poster-image>
  <div slot="centered-chrome">
  </div>
</media-controller>
```

In either case, I want to make sure that use case works, which I imagine will be pretty common

heff: The poster element feels like it belongs in the 'basic' example, and I'd love to avoid making people learn the layout slots in basic if we can avoid it. If we want to get more specific with the poster placement in 'advanced', no problem.

dylanjha: Correct, I agree with that. The user definitely shouldn't have to specify any slot.

I'm just referring to the template for `media-poster-image` using the `"centered-chrome"` slot under the hood.

src/js/media-poster-image.js
mmcc: I'm totally fine with swapping these out for IDs. My reasoning was that it could be nice to make it easier to target these for things like specifying a different `background-size` or adding a blur to the placeholder. I'm totally fine with keeping this as simple as possible and just pointing people towards using a div if they want to customize things. Thoughts?

src/js/media-poster-image.js
mmcc: Hmmm I like that philosophically, but do we care that it's deprecated?

mmcc:
> High level:
> 
> * This is 😎

▨-▨¬ლ(•_•) (▨_▨¬): (⌐▨_▨)

> * This feels like it should go in one of our well-defined slots or have its own slot, and not the "bottom-chrome" slot.

There's some other discussion on the PR, but I just worry that this feels like a pretty basic element to not be able to "just use" without understanding the different slots. If we can do the slot stuff _for_ the end-user I'm totally cool with that, but it'd feel bad if using the poster element was too complicated for the "basic" example.

> * This would change some of the assumptions in the component's styles.
> * Any reason we're using background-image instead of an ``? (This is mostly just curiosity about the choice.)

There's a lot of habit there. I've always used background styles for this kind of application because of the benefits of `background-size` and such (particularly in situations like this). If there's a reason to go with `img` tags instead I'm totally open to it.

> * Any reason we wouldn't want the image/blur-hash elements to be a default slotted element, instead of just "baked in" parts (note they could be both)? (This is a low priority, but could with little effort make the component more versatile)

Just to clarify, by this do you mean having a slot for both the placeholder and the image? I actually started to go down that path, but then pulled back because it seemed like it would just be easier to point people to using their own `div` once they want to start customizing too much. Thoughts?

luwes:
agreeing with @cjpillsbury, `` might be better here, the CSS prop `object-fit` should be able to cover it.
also if the user wants to optimize for SEO, can always add an `alt` attribute.
another potential here is taking in account responsiveness, and alternative image formats => picture element / source

heff:
I'm not clear if this is where the conversation is headed, but I'll say I'm cautious of an element that supports both a `src` attr and an ``, with two ways to do the same thing and a conflict if both are set. I think if we want simplicity, we're using attrs on an element that feels like ``, if we want to unlock more flexibility we could consider a `<media-poster-container>` element where the clear intention is to drop images inside of it, for more advanced use cases. I think we're in a pretty small percentile of use cases then that to Matt's point, wouldn't be that hard to build manually.

Re: slotting - My S1 is I don't feel great about magically slotting an element and none of the existing slots feel intuitive for the poster. I do see a slight conflict in using the poster in the default slot, where unless we special case media-poster-image and auto-absolute position it, it would butt up against top-chrome when used.
```
  <span part="layer vertical-layer">
    <slot name="top-chrome"></slot>
    <span class="spacer"><slot name="middle-chrome"></slot></span>
    <!-- default, effectively "bottom-chrome" -->
    <slot></slot>
  </span>
```
Special casing may be fine for this case, but it's worth understanding it would be needed.

I'm thinking a good answer could be both a `media-poster-image` component and a `slot="poster"`.
* The basic demo would use `<media-poster-image slot="poster">`
* Anything using `slot="poster"` gets appropriately hidden, so you can use your own image(s) if you want.
* `<media-poster-image>` would still hide itself, for cases where you're using it outside of a media-controller.

Re: using `` internally, sounds ok/good to me. I don't have a strong opinion, but the choice tells me we shouldn't expose the internal `parts` initially.

Minus the `parts`, I feel pretty good about the API and current functionality of this component, and feel like some of these conversations could be later enhancements. Who knew the poster image could get so complicated. :)

src/js/media-poster-image.js
heff: Maybe, if there's a negative understanding of it. I don't really understand why they removed it when there's not a clear alternative. Apparently they suggest making a progressive jpeg... I don't see a way `<picture>` solves the same issue.

I think I still like `lowsrc` but I'm good either way. Up to you.

gkatsev: Is `lowsrc` this https://html.com/attributes/img-lowsrc/? If so, it seems to be one of the features that  `<picture>` targets: a way to provide image alternatives depending on the page size or device pixel density.

Neither provide the ability to display one of the sources until the "main" source has finished loading. I think that `placeholder` is more direct and clear as to what it's use is.

dylanjha:
>  I do see a slight conflict in using the poster in the default slot

That was my feeling too

> none of the existing slots feel intuitive for the poster

I thought that "centered-chrome" felt intuitive enough, so that was my suggestion.

> I'm thinking a good answer could be both a media-poster-image component and a slot="poster".
> The basic demo would use <media-poster-image slot="poster">

A new slot for `"poster"` makes sense IMO, but I don't think the user should have to specify it when they use `<media-poster-image>`, it feels unnecessarily verbose for a common use case.

----

The only 2 things I feel strongly about i:

1. The user should have to specify a `"slot"` manually (if we magically slot it in the internals to either an existing or new slot, I think that's fine)
2. There should be some simple way to use `<media-poster-image>` + a top chrome slot to satisfy the use case of showing a centered play button on top of the poster image (which is a common thing people want to do)

As long as those are satisfied, I don't see anything in this discussion that I would have a problem with. The current code satisfies number (1) for me, @mmcc is number (2) possible with the current code?

dylanjha (approved):
💖

dylanjha:
QA'd this on staging basic example. I know @cjpillsbury and @heff synced on the details of this and I think they landed on the right solution for: ids instead of parts and a designated poster slot

PR #165, opened 2022-02-19
Replace snowpack w/ esbuild

luwes:
For a cleaner root dir and way less dependencies.
It's seems a better fit for a bundling a library than Snowpack.

~One thing that is removed is the local redirect when `http://127.0.0.1:8000/` is opened locally.~
~Maybe this could be added in the NPM script command or print out the `/examples` path in the terminal?~


dylanjha:
That's awesome 🙌🏼 

> One thing that is removed is the local redirect when http://127.0.0.1:8000/ is opened locally.

That is a bit of a bummer. Before we had that redirect, a number of folks were confused when opening localhost:8000 and seeing a dead page. It led confusion and people thinking they didn't have their dev env set up correctly. 

I'd love to find a way to preserve that or find some other way to launch the dev setup in a way that doesn't cause any confusion. I think that the log is helpful, but probably not enough to avoid confusion. Speaking for myself, I don't know if I would catch that detail in the log, I would probably just see the port number and open up `localhost:8000`.

heff (approved):
I have [no real opinion](https://www.youtube.com/watch?v=ussCHoQttyQ) on this one. Looks good. 👍  Thanks for moving our build system forward.

README.md
```diff
@@ -252,7 +252,7 @@ Web components. @heff spoke about [the potential of web components for video](ht
 
 Listen for the `update` event on the element to get the selected start and end timestamps. Full example in examples/clip-selector.html.
 
-![media clip selector example](./screenshots/media-clip-selector.gif?raw=true)
+![media clip selector example](./docs/assets/media-clip-selector.gif?raw=true)
```
cjpillsbury, line 255: 👌

cjpillsbury (approved):
looking 😎

PR #166, opened 2022-02-19
Remove TODO.md

luwes:
Seems not needed anymore, better to move to a discussion or a pinned issue.

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -1,9 +1,0 @@
-## To-dos for the project
-
-- [] Build multiple example skins, video and audio
-- [] Build multiple player integration examples
-  - [] `<audio>`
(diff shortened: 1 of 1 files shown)


cjpillsbury (changes requested):
As a prerequisite, let's make sure we add appropriate discussions and/or issues before removing this.

luwes:
These should cover it I think:
- https://github.com/muxinc/media-chrome/discussions/174
- https://github.com/muxinc/media-chrome/discussions/173
- https://github.com/muxinc/media-chrome/discussions/158

cjpillsbury (approved):
🏄‍♂️

PR #167, opened 2022-02-19
Move test related in one test folder

luwes:
This change moves all test related files in the `/test` folder.
`__tests__` is pattern from Jest I think and it looks a bit ugly. Might be my OCD kicking in 😅


dylanjha (commented):
Fine with me. I agree that I think the only place I've seen `__tests__` before is Jest.

I don't feel strongly at all one way or the other.

cjpillsbury:
I also don't feel strongly at all one way or another.

cjpillsbury (approved):

PR #169, opened 2022-02-24
Create Sandbox on Postinstall

cjpillsbury:
This will make sure sandbox exists on a fresh/updated install of media-chrome for devs, based on assumptions in the current build configs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,8 @@
     "dev": "snowpack dev --config snowpack.dev.config.cjs",
     "start": "yarn dev",
     "test": "web-test-runner --coverage --config test/web-test-runner.config.js",
-    "publish-patch": "yarn run build && np patch --no-tests --any-branch"
+    "publish-patch": "yarn run build && np patch --no-tests --any-branch",
+    "postinstall": "mkdirp sandbox"
   },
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -16,7 +16,8 @@
     "dev": "snowpack dev --config snowpack.dev.config.cjs",
     "start": "yarn dev",
     "test": "web-test-runner --coverage --config test/web-test-runner.config.js",
-    "publish-patch": "yarn run build && np patch --no-tests --any-branch"
+    "publish-patch": "yarn run build && np patch --no-tests --any-branch",
+    "postinstall": "mkdirp sandbox"
```
heff, line 20: I'm a little unclear why this is necessary. If all we need to do is add an unignored file like in video.js, we should just do that. Or maybe it's pushing the sanbox dir to the repo without it being ignored first.

cjpillsbury: not at all necessary if we do what you proposed. I just already added this. Happy to back out and do what you proposed.

dylanjha:
do we still need/want this?

luwes:
not needed anymore without Snowpack

PR #171, opened 2022-02-28
improv/thumbnails-different-dims: Update thumbnail and time range to …

cjpillsbury:
…handle different dims for thumbnails instead of hard-coded assumptions. Update mux-video example to demo and cleanup attrs.

Playback ID for portrait aspect-ratio content/thumbnails: dMTG3n6VfNyhRPM02mXY5AhPSk9E34bFqyqVm02gAAHD00

Note that this currently relies on different styling to handle relative positioning within the thumbnail image and resizing the thumbnail to a reasonable size for most content (using scale). While this is an improvement on the current implementation, which assumed more specific dimensions and didn't actually reliably fit the thumbnail content, there are cases where extremely large or non-standard proportions will take up a notably larger amount of space. We may want to do a followup to support configurable (re)scaling via e.g. css variable(s), but I think this could be treated as a followup PR to handle the more complex position + scale mathematical relationships.

### DIFF
--- a/examples/media-elements/mux-video.html
+++ b/examples/media-elements/mux-video.html
@@ -19,10 +19,15 @@
           metadata-video-id="video-id-12345"
           metadata-video-title="Mad Max: Fury Road Trailer"
           metadata-viewer-user-id="user-id-6789"
-          stream-type="vod"
-          autoplay
+          stream-type="on-demand"
           muted
         >
+          <track
+            label="thumbnails"
+            default
+            kind="metadata"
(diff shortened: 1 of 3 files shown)


heff:
> extremely large or non-standard proportions

Or smaller, making 50% too small. Would a `max/min-width/height` somewhere be a quick fix to any of that?

cjpillsbury:
> > extremely large or non-standard proportions
> 
> Or smaller, making 50% too small. Would a `max/min-width/height` somewhere be a quick fix to any of that?

Maybe...? I couldn't get it working by using min/max dims at least... 😬

heff (approved):

PR #172, opened 2022-03-02
fix/handle-src-state-changes: trigger relevant state updates when src…

cjpillsbury:
… is reset using 'emptied'.


heff (approved):
This was a smaller update than I expected, even after we talked about it. :)

PR #175, opened 2022-03-03
Move prettier to package.json

luwes:
For even less files / noise in the root folder


dylanjha:
Is this our preferred config location? Fine by me, personally.

luwes:
closing because the `pretty-quick` package doesn't seem to support this.
my preference is to have not too many files in the root directory, can become so noisy with all those config files

PR #177, opened 2022-03-04
bugfix: fix hovering over an active controller with no poster

mmcc:
Noticed this while I was testing something unrelated, but currently hovering over a playing stream that _doesn't_ have a poster image is broken.


cjpillsbury (approved):
🚢 🇮🇹

PR #179, opened 2022-03-07
Add max/min css vars for the thumb previews

luwes:
Related to https://github.com/muxinc/media-chrome/pull/171

Adds a configurable max and min width for the thumb preview.
This uses a CSS transform: scale() to up and down scale to the max and min width via JS if needed. A small down side is that the border also scales with that, this is solved by setting each CSS property again with the scale factored in with JS.

Also changed the positioning via `style.left` to `style.transform = translate` because this should perform better.
https://www.paulirish.com/2012/why-moving-elements-with-translate-is-better-than-posabs-topleft/

### Fixes 2 small issues

- bottom black border in thumb preview
- fading out animation did not work because of the CSS display prop

### TODO

 - [ ] An improvement for future iterations is that the thumbnail preview should not go out of the player's bounds which it does currently. Not specific to this PR, this behavior is the same on main.

### Motivation

SCR-20220306-pg7

### DIFF
--- a/examples/vertical.html
+++ b/examples/vertical.html
@@ -5,6 +5,17 @@
     <title>Media Chrome Vertical Video Usage Example</title>
     <script type="module" src="../dist/index.js"></script>
     <style>
+      /** add styles to prevent CLS (Cumulative Layout Shift) */
+      media-controller:not([audio]) {
+        display: block;         /* expands the container if preload=none */
+        max-width: 360px;       /* allows the container to shrink if small */
+        aspect-ratio: 9 / 16;   /* set container aspect ratio if preload=none */
+      }
+
+      video {
+        width: 100%;      /* prevents video to expand beyond its container */
+      }
+
       .examples {
         margin-top: 20px;
       }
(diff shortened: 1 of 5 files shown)


dylanjha:
I found that the thumbnails were a bit small on stream.new, especially noticeable in a large viewport. Having some solution for that makes a lot of sense.

As far as the actual approach, `--media-thumbnail-preview-min-width` and `--media-thumbnail-preview-max-width` make sense 👍🏼 

Are there any gotchas or caveats when using these variables with vertically orientated videos?

luwes:
@dylanjha good call, we might have to find a solution for vertical videos. Add min / max height vars too?

examples/vertical.html
```diff
@@ -0,0 +1,75 @@
+<!DOCTYPE html>
```
cjpillsbury, line 1: good example add, even without the thumbnails 👌

src/js/media-time-range.js
```diff
@@ -25,47 +25,49 @@ const template = document.createElement('template');
...
-        margin-top: 10px;
-        opacity: 0;
-      }
-      100% {
-        /* transform-origin: bottom center; */
-        /* transform: scale(1); */
-        margin-top: 0;
-        opacity: 1;
-      }
+      transition: visibility .5s, opacity .5s;
```
cjpillsbury, line 68: 👌

src/js/media-time-range.js
```diff
@@ -25,47 +25,49 @@ const template = document.createElement('template');
...
-      margin: 100px;
-      background-color: #ff0;
+      width: 0;
+      height: 0;
+      position: absolute;
+      left: 50%;
+      transform: translateX(-50%);
+      bottom: -10px;
+      border-left: 10px solid transparent;
+      border-right: 10px solid transparent;
```
gkatsev, line 63: non-blocking: you can also use [`border-inline`](https://developer.mozilla.org/en-US/docs/Web/CSS/border-inline) to have this as one item.

cjpillsbury (changes requested):
blocker: Make sure you update the relevant docs.
at least before smoke testing, code LGTM

cjpillsbury (approved):
👌

PR #182, opened 2022-03-09
Feat/state change events

cjpillsbury:
This is a light weight base implementation of state change events. I figured before making any additional decisions/going any further, we should decide if we want to do any additional translations or handling on initialization. I made this example page so you can easily see the state change behavior and values under different conditions:

https://media-chrome-git-fork-cjpillsbury-feat-state-change-events-mux.vercel.app/examples/state-change-events-demo.html

### DIFF
--- a/src/js/constants.js
+++ b/src/js/constants.js
@@ -22,6 +22,32 @@ export const MediaUIEvents = {
   UNREGISTER_MEDIA_STATE_RECEIVER: 'unregistermediastatereceiver',
 };
 
+export const MediaStateChangeEvents = {
+  MEDIA_AIRPLAY_UNAVAILABLE: 'mediaairplayunavailablechange',
+  MEDIA_PIP_UNAVAILABLE: 'mediapipunavailablechange',
+  MEDIA_PAUSED: 'mediapausedchange',
+  MEDIA_HAS_PLAYED: 'mediahasplayedchange',
+  MEDIA_MUTED: 'mediamutedchange',
+  MEDIA_VOLUME_LEVEL: 'mediavolumelevelchange',
+  MEDIA_VOLUME: 'mediavolumechange',
+  MEDIA_VOLUME_UNAVAILABLE: 'mediavolumeunavailablechange',
+  MEDIA_IS_PIP: 'mediaispipchange',
+  MEDIA_CAPTIONS_LIST: 'mediacaptionslistchange',
+  MEDIA_SUBTITLES_LIST: 'mediasubtitleslistchange',
+  MEDIA_CAPTIONS_SHOWING: 'mediacaptionsshowingchange',
+  MEDIA_SUBTITLES_SHOWING: 'mediasubtitlesshowingchange',
+  MEDIA_IS_FULLSCREEN: 'mediaisfullscreenchange',
+  MEDIA_PLAYBACK_RATE: 'mediaplaybackratechange',
+  MEDIA_CURRENT_TIME: 'mediacurrenttimechange',
+  MEDIA_DURATION: 'mediadurationchange',
+  MEDIA_PREVIEW_IMAGE: 'mediapreviewimagechange',
+  MEDIA_PREVIEW_COORDS: 'mediapreviewcoordschange',
+  // MEDIA_CHROME_ATTRIBUTES: 'media-chrome-attributes',
+  // MEDIA_CONTROLLER: 'media-controller',
+  MEDIA_LOADING: 'medialoadingchange',
+  USER_INACTIVE: 'userinactivechange',
+};
+
 export const MediaUIAttributes = {
   MEDIA_AIRPLAY_UNAVAILABLE: 'media-airplay-unavailable',
   MEDIA_PIP_UNAVAILABLE: 'media-pip-unavailable',
@@ -46,6 +72,34 @@ export const MediaUIAttributes = {
   MEDIA_CONTROLLER: 'media-controller',
   MEDIA_LOADING: 'media-loading',
 };
+
+// Maps from state change event type -> attribute name
+export const StateChangeEventToAttributeMap = Object.entries(
+  MediaStateChangeEvents
+).reduce(
+  (mapObj, [key, eventType]) => {
+    const attrName = MediaUIAttributes[key];
+    if (attrName) {
+      mapObj[eventType] = attrName;
+    }
+    return mapObj;
+  },
+  { userinactivechange: 'user-inactive' }
+);
+
+// Maps from attribute name -> state change event type
+export const AttributeToStateChangeEventMap = Object.entries(
+  MediaUIAttributes
+).reduce(
+  (mapObj, [key, attrName]) => {
+    const evtType = MediaStateChangeEvents[key];
+    if (evtType) {
+      mapObj[attrName] = evtType;
+    }
+    return mapObj;
+  },
+  { 'user-inactive': 'userinactivechange' }
+);
 
 export const TextTrackKinds = {
   SUBTITLES: 'subtitles',
--- a/src/js/media-container.js
+++ b/src/js/media-container.js
@@ -12,7 +12,7 @@ import {
   Window as window,
   Document as document,
 } from './utils/server-safe-globals.js';
-import { MediaUIEvents, MediaUIAttributes } from './constants.js';
+import { MediaUIEvents, MediaUIAttributes, MediaStateChangeEvents } from './constants.js';
 import { nouns } from './labels/labels.js';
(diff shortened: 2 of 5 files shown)


src/js/constants.js
```diff
@@ -22,6 +22,32 @@ export const MediaUIEvents = {
   UNREGISTER_MEDIA_STATE_RECEIVER: 'unregistermediastatereceiver',
 };
 
+export const MediaStateChangeEvents = {
+  MEDIA_AIRPLAY_UNAVAILABLE: 'mediaairplayunavailablechange',
+  MEDIA_PIP_UNAVAILABLE: 'mediapipunavailablechange',
+  MEDIA_PAUSED: 'mediapausedchange',
```
heff, line 28: I like the consistency here, and also it's the crux of the decision. It would be odd to have both this and `play`/`pause`/`playing` events, to mimic the video element API. This is making me consider not mimicking the video element. That requires a little more thought and discussion.

heff:
Thanks for putting this together, @cjpillsbury. Very helpful to see. 

There's a few key decisions/patterns we'd be committing to with this change, so I want to make them clear to prompt feedback.

### `[ATTR]change` event names instead of intentionally named events based on action
For example, `mediapausedchange` not `mediapause` and `mediaplay`.
In general, we have a design principle of making elements [feel like native HTML](https://github.com/muxinc/media-chrome/blob/main/ARCHITECTURE.md#components-should-feel-like-native-html). Both of these patterns can be found in HTML. The pause/play naming approach could be argued is _more_ like HTML by pointing at the video element API, but I do like the predictability of consistent `change` naming, and with the addition of using `event.detail` to pass the current state there's little benefit I see to using unique action-based event names.

### Different API than `<video>`
If we emit `mediapausedchange` instead of `play` and `pause`, we now have a different API than `<video>`. (I'm not excited about a world where we do both at the same time.) I just gave a talk at Demuxed scolding player devs for designing different APIs than `<video>` and reducing ecosystem compatibility, so I'm sensitive to this one.

Today, we already have the beginnings of mimicking the `<video>` API [in media-controller](https://github.com/muxinc/media-chrome/blob/dd27a2a6620c804f51a6d496156805043edb6a44/src/js/media-controller.js#L616). It's internally used but undocumented, so we can change it if needed.

My thinking on why we _should_ have a different API than `<video>` are:
* Media Chrome is not a "player" or media element. It's a media UI/controller. The purpose of media-chrome is to handle user intent and reflect the state that's relevant to the user/UI. That overlaps with, but is different than the `<video>` API, where details like readyState/networkState are not directly relevant to users.
* A way to tell the difference between user actions and API/internally-driven events is missing with `<video>`. "Was `play` triggered by a user or something else?"
* A different API allows us to abstract complexity for UI use cases without breaking the `<video>` API contract/expectations. For example, when _scrubbing_, we may want the `media-paused` state to stay `false` even though the underlying media element needs to be paused while scrubbing.
* Media-chrome doesn't hide the media element, so if a `<video>` API is needed for some reason, it's still clearly available.

The further implication of this could mean renaming `mediaController.play()` to `mediaController.playRequest()` to match the event name and reinforce that it is a different API for a different purpose.

Please share any feedback on any of the above.

src/js/media-container.js
```diff
@@ -306,6 +316,11 @@ class MediaContainer extends window.HTMLElement {
 
       // Stay visible if hovered over control bar
       this.removeAttribute('user-inactive');
```
gkatsev, line 318: Very much unrelated, but noticed it when looking here. I'm not sure that it matters much in the real world, but thought I'd mention it here anyway.
We're adding and removing the `user-inactive` attribute unconditionally, particularly on mousemove. This causes devtools to think that the element has updated a lot. Particularly, as you move the mouse around on the various components of the control bar you can see devtools highlight `media-controller` showing that it's been updated, even if the update is removing a non-existent attribute.

gkatsev:
I think the question is: **should media-chrome be treated as a player or a controls/UI framework?**

If it should be a player, it *should* try and mimic the `<video>` element as much as possible, even down to events and that media-chrome users, once they slot a media element, don't need to reference the media element itself anymore as they should be getting everything via media-chrome.
If it's a controls/UI framework, then I think the expectations would be that for regular media element lifecycle events, you'd listen on the media-slotted element. The media-chrome is free to basically do what it wants since the events will be purely ui/media-chrome related. In this case, we'd still want to propagate events from the video element because a lot of the UI decisions are based around that.

luwes:
I like this change, it can be useful I think but I have my doubts we should promote its use too much.

I'm definitely in the camp of the media controller not extending the video element and mimicking its API.

If Media Chrome is a UI framework I'm also not convinced that the controller should be a source of truth for video state outside Media Chrome. The only reason the controller is there is to act like a middle man between MC controls and media.

What I really liked so far working with Media Chrome is not having to touch a special JS API to have a great player UI. 
It's HTML and CSS. The attributes flowing through the DOM tree sharing state to the views.

The controller is among other things an event bus and I kinda wish it would not expose much more to the outside world.
It would discourage devs to use the JS API of the media controller.

I would try to remove as much as possible of these public methods to be honest.
https://github.com/muxinc/media-chrome/blob/3aad2040963c734826a3d3fe70c5c55af58ac432/src/js/media-controller.js#L622-L766

Some are indeed very useful like entering fullscreen, PiP (has a lot of logic to make it uniform across browsers) but they can all be triggered via an event if needed. Also if you're using the media chrome controls, there shouldn't be a need right?

Other getters like paused, volume, duration are also accessible via the media controller attribute. To me it seems less confusing they would only be accessible that way. Even Shaka or hls.js doesn't expose those video property getters on their API. In this case you should use the source of truth here which is the media element.

src/js/media-container.js
dylanjha: Good find

heff: Yeah, not ideal, but also should be prioritized appropriately, i.e. if it requires a big re-architecture we should consider if the issue is worth it right now. After the call this morning it sounds like @cjpillsbury thinks it may not be a big effort, or we could fix it per-attribute as needed.

gkatsev: yeah, I don't think it's a *big* issue, but it's worth addressing. Theoretically, it's as easy as checking if the prop/attr has changed or not or only removing/add the attribute if the aren't added/removed as appropriate.

heff:
Summarizing some conclusions based on conversation:

* Media Chrome will *not* mimic the `<video>` API - it is not a "player", and will not provide a similar set of API functions
* We will use `[attr name]change` event naming pattern
* (future PR) Remove the public API methods for the basic functions (play, pause) - we will instead re-enforce that events are the primary way to interact with media-controller, and otherwise changing the media-controller attributes is another option.
  * Some functions we may keep because of the cross-browser complexity, like fullscreen and PiP
  * We may still want internal versions of some of the functions to share code within media controller, but can rename
* (future PR(s)) Address frequent duplicate events/attr updates

> The only reason the controller is there is to act like a middle man between MC controls and media.

That's a decent framing.

> don't need to reference the media element itself anymore as they should be getting everything via media-chrome

While media-chrome isn't a "player" and we won't mimic the video API, it does feel justifiable or at least very convenient to provide consistent access to the underlying media element's API. For example if there is a playlist swapping the media element, allowing an external application to know when the media element has changed. We have [a method for accessing the media](https://github.com/muxinc/media-chrome/blob/main/src/js/media-container.js#L209), we would just need a way for knowing when the media has changed, like mediaSet/mediaUnset. We should address that later though.

examples/state-change-events-demo.js
```diff
@@ -0,0 +1,36 @@
+import { constants } from '../dist/index.js';
```
heff, line 1: This example is helpful. 👍 Nit: Can we either add this JS to the HTML or make a directory for these two files to live in?

src/js/constants.js
```diff
@@ -47,6 +73,34 @@ export const MediaUIAttributes = {
...
+    const attrName = MediaUIAttributes[key];
+    if (attrName) {
+      mapObj[eventType] = attrName;
+    }
+    return mapObj;
+  },
+  { userinactivechange: 'user-inactive' }
+);
+
+// Maps from attribute name -> state change event type
```
heff, line 90: Do we need this if we already have it in the const above?

heff (approved):
A few nits, but approved. 👍

src/js/constants.js
cjpillsbury: Yeah, `'user-inactive'` is the "odd man out" because it isn't one of the `MediaUIAttributes`.

PR #183, opened 2022-03-13
Add custom elements manifest analyzer + plugin

luwes:
This change is the start for adding a custom elements manifest.
Related to https://github.com/muxinc/media-chrome/discussions/158

A little downside of Media Chrome using constant variables is that the analyzer can't pick the attributes and events up automatically. An alternative is to use special JSDoc syntax to get these picked up.
https://custom-elements-manifest.open-wc.org/analyzer/getting-started/#supported-jsdoc

I added one example for the Airplay button.

The custom function `defineCustomElement()` in Media Chrome was causing issues too, luckily this could be automated with a plugin. The analyzer expects `customElements.define('my-el')`.

### DIFF
--- a/custom-elements-manifest.config.js
+++ b/custom-elements-manifest.config.js
@@ -1,0 +1,131 @@
+import fs from 'fs';
+import {
+  getDeclarationInFile,
+  hasIgnoreJSDoc,
+  isCustomElementsDefineCall,
+} from '@custom-elements-manifest/analyzer/src/utils/ast-helpers.js';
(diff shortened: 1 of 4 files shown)


cjpillsbury:
Since this is all built on tsc plugins using its AST model, we _*should*_ be able to build something that works with our constants. Do you think this would be a value add at this stage without also updating the jsdoc comments?

luwes:
@cjpillsbury yes, I think it was possible with Babel, https://github.com/wooorm/babel-plugin-inline-constants. 
maybe TS too, might be worth trying out

luwes:
closing until we get to this

PR #185, opened 2022-03-13
Hook up media-buffered attribute to progress background

luwes:
SCR-20220313-k2p

SCR-20220313-k3v

### DIFF
--- a/src/js/media-controller.js
+++ b/src/js/media-controller.js
@@ -1004,7 +1004,7 @@ function serializeTimeRanges(timeRanges = []) {
     .map((_, i) => [
       Number(timeRanges.start(i).toFixed(2)),
       Number(timeRanges.end(i).toFixed(2)),
-    ])
+    ].join(':'))
     .join(' ');
 }
 
--- a/src/js/media-controller.js
+++ b/src/js/media-controller.js
@@ -331,6 +331,12 @@ class MediaController extends MediaContainer {
           this.duration
         );
       },
+      'progress,emptied': () => {
+        this.propagateMediaState(
+          MediaUIAttributes.MEDIA_BUFFERED,
+          serializeTimeRanges(this.media?.buffered)
+        );
+      },
       ratechange: () => {
         this.propagateMediaState(
           MediaUIAttributes.MEDIA_PLAYBACK_RATE,
@@ -992,6 +998,15 @@ const volumeSupportPromise = hasVolumeSupportAsync().then((supported) => {
 });
 
 const airplaySupported = !!window.WebKitPlaybackTargetAvailabilityEvent;
+
+function serializeTimeRanges(timeRanges = []) {
+  return Array.from(timeRanges)
+    .map((_, i) => [
+      Number(timeRanges.start(i).toFixed(2)),
+      Number(timeRanges.end(i).toFixed(2)),
+    ])
+    .join(' ');
+}
 
 defineCustomElement('media-controller', MediaController);
 
--- a/src/js/media-time-range.js
+++ b/src/js/media-time-range.js
@@ -159,7 +159,7 @@ class MediaTimeRange extends MediaChromeRange {
   get mediaBuffered() {
     const buffered = this.getAttribute(MediaUIAttributes.MEDIA_BUFFERED);
     if (buffered) {
-      return buffered.split(' ').map((timePair) => timePair.split(','));
+      return buffered.split(' ').map((timePair) => timePair.split(':'));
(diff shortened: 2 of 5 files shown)


src/js/media-controller.js
```diff
@@ -993,6 +999,15 @@ const volumeSupportPromise = hasVolumeSupportAsync().then((supported) => {
 
 const airplaySupported = !!window.WebKitPlaybackTargetAvailabilityEvent;
 
+function serializeTimeRanges(timeRanges = []) {
+  return Array.from(timeRanges)
+    .map((_, i) => [
+      Number(timeRanges.start(i).toFixed(2)),
+      Number(timeRanges.end(i).toFixed(2)),
+    ])
+    .join(' ');
```
cjpillsbury, line 1008: non-blocking: I'm cool with using `,` to separate the tuple values, but for other complex object's we've used `:`. Since this is only relevant for "advanced usage, I think it's fine if we want to start here.

src/js/media-time-range.js
```diff
@@ -148,6 +152,17 @@ class MediaTimeRange extends MediaChromeRange {
...
   // }
 
+  get mediaDuration() {
+    return +this.getAttribute(MediaUIAttributes.MEDIA_DURATION);
+  }
+
+  get mediaBuffered() {
+    const buffered = this.getAttribute(MediaUIAttributes.MEDIA_BUFFERED);
+    if (buffered) {
+      return buffered.split(' ').map((timePair) => timePair.split(','));
```
cjpillsbury, line 162: non-blocking: should this be moved to a function alongside `serializeTimeRanges` in some utils file? `parseTimeRangeString`? `deserializeTimeRanges`?

src/js/media-time-range.js
```diff
@@ -148,6 +152,17 @@ class MediaTimeRange extends MediaChromeRange {
...
+    }
+  }
+
```
cjpillsbury, line 165: blocking: since this will suddenly add buffered indication to the time bar, I'm a bit concerned about rolling it out with the current color logic (found just below):

```js
const buffPercent =
      (buffered[buffered.length - 1][1] / this.mediaDuration) * 100;
    colorsArray.splice(1, 0, [
      'var(--media-time-buffered-color, #777)',
      buffPercent,
    ]);
```

This uses the very last `end` value and will render as if everything before it is buffered. Do we want that to suddenly be the default visualization for everyone?

cjpillsbury:
blocking: Make sure you update the docs (top level README, `media-time-range` doc) and the `media-time-range` example.

cjpillsbury (changes requested):
one callout that I think should be discussed before merging plus the docs etc. ask, otherwise 👌

src/js/media-controller.js
luwes: will update

src/js/media-time-range.js
luwes: until it's needed a 2nd time maybe, with this pure vanilla ESM approach I'd prefer also not to make more file / dependency requests than necessary.

src/js/media-time-range.js
luwes: I didn't touch this logic but yes might need some changes.

I would suggest make it the end of the buffer range where the playhead currently is in, should be less confusing if they would hit a buffer stall...

src/js/media-time-range.js
```diff
@@ -161,8 +180,14 @@ class MediaTimeRange extends MediaChromeRange {
     }
 
     const buffered = this.mediaBuffered;
-    const buffPercent =
-      (buffered[buffered.length - 1][1] / this.mediaDuration) * 100;
+    let currentBufferedEnd;
+    for (const [start, end] of buffered) {
+      if (this.mediaCurrentTime >= start && this.mediaCurrentTime <= end) {
```
cjpillsbury, line 185: non-blocking: may want to store this value to not recompute unnecessarily.

src/js/media-time-range.js
```diff
@@ -161,8 +180,14 @@ class MediaTimeRange extends MediaChromeRange {
...
+        currentBufferedEnd = end;
```
cjpillsbury, line 186: 👍

cjpillsbury (approved):
Pre-approving but make sure you've tested with:
* live content (just to make sure there aren't RTEs or similar. Less worried about a "reasonable" rendering)
* no src set
* nothing buffered (should be similar to no src set)
* switching media and/or src

PR #186, opened 2022-03-14
Add a experimental expandable control-strip

luwes:
Related to https://github.com/muxinc/media-chrome/discussions/123

This adds a new component to make it super easy to add an expandable part in the control bar like
https://en.wikipedia.org/wiki/Control_Strip

For the volume range but can also work like in the Apple Magic touch bar where part of the controls are hidden and there is a special expand button.

Still experimental and needs a few more attributes to make it work like the magic touch bar.
Also thoughts on accessibility are welcome.

https://media-chrome-git-fork-luwes-control-strip-mux.vercel.app/examples/advanced.html

### DIFF
--- a/examples/advanced.html
+++ b/examples/advanced.html
@@ -63,8 +63,10 @@
           <media-play-button></media-play-button>
           <media-seek-backward-button seek-offset="15"></media-seek-backward-button>
           <media-seek-forward-button seek-offset="15"></media-seek-forward-button>
-          <media-mute-button></media-mute-button>
-          <media-volume-range></media-volume-range>
+          <media-control-strip>
+            <media-mute-button slot="open-button"></media-mute-button>
+            <media-volume-range></media-volume-range>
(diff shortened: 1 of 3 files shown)


gkatsev:
It'll be good to try and do all the opening stuff via css-only rather than events in JS. Makes it easier to make sure that things work properly for both mouse and keyboard and screen readers.

luwes:
I've been wanting to build this for a player for a while now but never got to it.
to me this seems a nicer solution than having the "gear icon" with a panel and sub menu panels like the YT player.

it's similar to Wistia in mobile view but they choose to fully hide all right side controls. Best to keep as many as possible still visible with the available space. Could even be smart about which ones to be visible, like the ones which are interacted with the most.

I'm also not sure if it needs to be part of Media Chrome, it's possible to make it even more generic since it's just a UI component.

luwes:
closing for now, needs more polishing

PR #190, opened 2022-03-16
fix: use mouseleave instead of mouseout

gkatsev:
With mouseout, we would get an event each time we cross an element
boundary, meaning we could end up marking the player as inactive as a
user is mousing around in the player. Instead, we should use mouseleave
that would only trigger when we leave the entire element including its
children.

### DIFF
--- a/src/js/media-container.js
+++ b/src/js/media-container.js
@@ -330,7 +330,7 @@ class MediaContainer extends window.HTMLElement {
     });
 
     // Immediately hide if mouse leaves the container
-    this.addEventListener('mouseout', (e) => {
+    this.addEventListener('mouseleave', (e) => {
       if (this.autohide < 0) return;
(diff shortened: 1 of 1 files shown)


src/js/media-container.js
```diff
@@ -330,7 +330,7 @@ class MediaContainer extends window.HTMLElement {
     });
 
     // Immediately hide if mouse leaves the container
-    this.addEventListener('mouseout', (e) => {
```
gkatsev, line 333: now we're even in-line with the comment 😛

luwes (approved):
stumbled on the same thing recently,
https://github.com/muxinc/media-chrome/pull/186/files#diff-e8ad02c499c9ca2df8d46313c2c32db57d591b7a5f7797aed9144ad1232f68f6R74

LGTM 🚢

PR #191, opened 2022-03-17
fix(user-inactive-poster): treat slotted poster like slotted media fo…

### LINKED ISSUE #155: Shouldn't media-controller auto-hide when a video is defined with the "autoplay" attribute?
Hi, I am currently testing the library in a WebView on iOS, using the following code:
```
<media-controller autohide="2">
    <video slot="media" autoplay>
        <source src="${ url }" type="${ mimeType }">
    </video>
    <media-control-bar>
        <media-mute-button></media-mute-button>
        <media-time-range></media-time-range>
        <media-duration-display></media-duration-display>
    </media-control-bar>
</media-controller>
```
The problem is that users have to realize, they must touch the video after it starts autoplaying for the autohide attribute to kick in. Wouldn't it be more logical for autohide to be respected without requiring user interaction in this case? In other words: The video start playing automatically, then the controls auto-hide after 2 seconds (autohide="2").

### LINKED ISSUE #188: user inactivity tracking doesn't account for mouse clicks
In media-container, we set user-inactive after a timer or when we mouse out of the video player and remove it on mouse movement or keyup. For completeness, we should also consider adding `mouseup` or `click` events. As it's possible that the user clicked on the player without moving the mouse since the player was last put into user-inactive mode.

cjpillsbury:
…r pointer-events for user-inactive behavior.

Fixes #188 
Fixes #155

### DIFF
--- a/src/js/media-container.js
+++ b/src/js/media-container.js
@@ -88,7 +88,8 @@ template.innerHTML = `
       height: 100% !important;
     }
 
-    ::slotted(:not([slot=media])) {
+    /* Need to revisit this. May be too presumptuous for user-inactive behavior */
+    ::slotted(:not([slot=media]):not([slot=poster])) {
       pointer-events: auto;
     }
 
--- a/src/js/media-container.js
+++ b/src/js/media-container.js
@@ -277,6 +277,10 @@ class MediaContainer extends window.HTMLElement {
       );
     }
 
+    // Assume user is inactive until they're not (aka user-inactive by default is true)
+    // This allows things like autoplay and programmatic playing to also initiate hiding controls (CJP)
+    this.setAttribute('user-inactive', 'user-inactive');
+
     const scheduleInactive = () => {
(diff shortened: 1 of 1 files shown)


src/js/media-container.js
```diff
@@ -276,19 +277,23 @@ class MediaContainer extends window.HTMLElement {
...
+    this.setAttribute('user-inactive', 'user-inactive');
+
     const scheduleInactive = () => {
       this.removeAttribute('user-inactive');
       const evt = new window.CustomEvent(
         MediaStateChangeEvents.USER_INACTIVE, 
         { composed: true, bubbles: true, detail: false }
       );
       this.dispatchEvent(evt);
-      window.clearTimeout(this.inactiveTimeout);
```
cjpillsbury, line 286: renamed this per Media Chrome naming conventions to indicate this should be treated as a private variable.

gkatsev (approved):
Confirmed that behavior is as expected now.

PR #195, opened 2022-03-24
Add IIFE and CJS bundle types to dist folder

luwes:
This adds the bundle types IIFE and CJS so they can be accessed easily from a CDN like unpkg in the dist folder.

Many users still want to request one single bundle request with the non-module like <script> tag.
It's also good to test because custom elements behave differently when loaded before or after the custom element HTML is parsed. 

script[type=module] are deferred by default so they will execute only when all HTML is parsed.
See [gist.github.com/jakub-g/385ee6b41085303a53ad92c7c8afd7a6](https://gist.github.com/jakub-g/385ee6b41085303a53ad92c7c8afd7a6)

Also for the React version but not sure it's as necessary, users would more likely be using a bundler with React I imagine.
https://github.com/muxinc/media-chrome/compare/main...luwes:bundle-types-react


dylanjha (approved):
IMO there's no harm in adding this, because:

1. It adds new outputs in dist/ without impacting the default ESM build (no chance in breaking anything existing)
2. Wes pointed out some clear use cases where this dists come in handy

@heff are you opposed to such a thing?

heff:
I'm not against it for use cases that really require it, but also I think we should be generally opinionated on the _right_ way if you have options, and keep the main "Using/installing" example simple. i.e. in the Readme we could have a link to another doc that says something like "Need IIFE or CJS?", rather than having a ton of different install options right off the bat. Make sense?

heff (approved):

gkatsev:
I made a tentative PR against this branch here https://github.com/luwes/media-chrome/pull/1

It expands on this change with some of the things from https://github.com/muxinc/elements/pull/163. It defaults `main` to be a cjs build, with `module` being esm, and `browser` being the iife build. It also uses the new `exports` field which is the current/forward looking field for specifying esm/cjs requires/imports.
In addition, it currently has an `unpkg` field that points at the iife, but that can be removed or made to point at the esm files depending on whether we want `<script type="module">` or `<script>` being the default.

luwes:
@gkatsev cool, that looks good. I'll merge this first

PR #199, opened 2022-04-09
Add support for media swapping in chained slots

luwes:
This change adds support for swapping out slotted media slots aka `<slot name="media" slot="media"></slot>` which might come in useful when you'd like to switch a Media Theme but keep the same video element.

See the bug without this change here (click the Switch Theme button):
https://elements-demo-vanilla-git-fork-luwes-media-theme-mux-mux.vercel.app/mux-player.html

### DIFF
--- a/src/js/media-container.js
+++ b/src/js/media-container.js
@@ -193,6 +193,26 @@ class MediaContainer extends window.HTMLElement {
 
     const observer = new MutationObserver(mutationCallback);
     observer.observe(this, { childList: true, subtree: true });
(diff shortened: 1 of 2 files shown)


heff (approved):
Is it accurate to say that this isn't just about switching themes, it also fixes any time a media element is added to a theme after initial load?

luwes:
thanks for the review! 
yup if it's added via a chained slot it is, otherwise the mutation observer above this code would catch it.

PR #200, opened 2022-04-09
Remove unneeded media like API

luwes:
Like we discussed in https://github.com/muxinc/media-chrome/pull/182#issuecomment-1067370339 the media controller shouldn't try to mimic the video API.

- This change removes all the getters that could be replaced with a getAttribute or hasAttribute
- Moves the getters that were used internally as helpers to private functions
- Removes all setters that can be replaced with an event dispatch on the controller
e.g. `controller.dispatchEvent(new Event(MEDIA_ENTER_FULLSCREEN_REQUEST));`

The goal is to remove duplication and confusion on which API should be used (media API vs controller API) and add this constraint so media controller doesn't try to do too much, become a player basically.

### DIFF
--- a/src/js/media-controller.js
+++ b/src/js/media-controller.js
@@ -222,7 +222,7 @@ class MediaController extends MediaContainer {
         );
       },
       MEDIA_SHOW_CAPTIONS_REQUEST: (e) => {
-        const tracks = this.captionTracks;
+        const tracks = getCaptionTracks(this);
         const { detail: tracksToUpdate = [] } = e;
         updateTracksModeTo(TextTrackModes.SHOWING, tracks, tracksToUpdate);
       },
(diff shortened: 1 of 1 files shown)


dylanjha (changes requested):
@luwes how can we approach this without taking on the risk of breaking existing implementations?

Is calling `play`, `showingCaptionTracks`, etc. on media-controller something that people might already be doing? Or do we document that anywhere?

src/js/media-controller.js
```diff
@@ -622,155 +625,85 @@ class MediaController extends MediaContainer {
...
-
-    if (!this.media) return level;
+  /**
+   * Media Controller should not mimic the HTMLMediaElement API.
+   * @see https://github.com/muxinc/media-chrome/pull/182#issuecomment-1067370339
+   */
+}
 
-    const { muted, volume } = this.media;
+const getPaused = (el) => {
```
heff, line 634: Nit: We might use `controller` for these args instead of `el`, to be clear what the expected API is. They kind of look like generic element functions at first.

heff (approved):
Looks good!

PR #201, opened 2022-04-09
Change default loading indicator width 100px

luwes:
Removes the need for styling it specifically most of the times.
In the control-bar change it to 44px default.


heff (approved):
Makes sense 👍

PR #205, opened 2022-04-14
Make MC sizing based on video element size

### LINKED ISSUE #103: Add dimensions (and aspect-ratio maybe) style to examples
The examples show serious layout shift on page load at the moment.
https://media-chrome.mux.dev/examples/basic.html

luwes:
Fixes #103

This change makes Media Chrome sizing based on the video element and behave the same.

It's a good thing I believe that the user is encouraged to set dimensions on Media Chrome because it will prevent cumulative layout shift. It's already too late for us to set default dimensions on Media Chrome internally because of the deferred nature of `<script type=module>`. A good example is to refresh the webpage a couple of times and see the difference between:

- https://media-chrome.mux.dev/examples/advanced.html
- https://media-chrome-git-fork-luwes-video-like-sizing-mux.vercel.app/examples/advanced.html

Look at the [View more examples](https://media-chrome.mux.dev/examples/) link or turn off javascript and compare.

Note that the default (with no video content) minimum video dimensions are 300x150 ~so some of the examples might look strange at the moment. These can fixed up by adding some simple CSS styles but I'll wait until we know for sure this is the way we like the sizing to behave.~ I added some CSS for these.

The `line-height: 0;` is there because it's an inline-block element, otherwise it would have a small bottom padding / gap.

### DIFF
--- a/examples/themes/demuxed-2021-theme.html
+++ b/examples/themes/demuxed-2021-theme.html
@@ -7,6 +7,20 @@
   </head>
   <body>
     <style>
+      .examples {
+        margin-top: 20px;
+      }
+
+      /* Hide custom elements that are not defined yet */
+      media-control-bar :not(:defined) {
+        display: none;
+      }
+
+      video {
+        width: 720px;
+        aspect-ratio: 16 / 9;
+      }
+
       media-control-bar > * {
         height: 42px;
         --media-control-background: #000;
@@ -620,6 +634,9 @@
           </media-fullscreen-button>
         </media-control-bar>
       </media-controller>
+      <div class="examples">
+        <a href="../">View more examples</a>
+      </div>
     </main>
   </body>
 </html>
--- a/src/js/media-container.js
+++ b/src/js/media-container.js
@@ -23,10 +23,11 @@ template.innerHTML = `
       box-sizing: border-box;
       position: relative;
       display: inline-block;
+      line-height: 0;
       background-color: #000;
     }
 
-    :host(:not([audio])) *[part~=layer] {
+    :host(:not([audio])) *[part~=layer]:not([part~=media-layer]) {
       position: absolute;
       top: 0;
       left: 0;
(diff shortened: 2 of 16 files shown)


gkatsev (approved):

cjpillsbury (approved):
We may want to approach the examples a bit differently, but I defer to @heff on this. For the core code, this is 👌 🔥 .

luwes:
> We may want to approach the examples a bit differently, but I defer to @heff on this. For the core code, this is 👌 🔥 .

Solution to the answer https://github.com/muxinc/media-chrome/issues/103#issuecomment-1006169200

examples/themes/demuxed-2021-theme.html
```diff
@@ -7,6 +7,20 @@
   </head>
   <body>
     <style>
+      .examples {
+        margin-top: 20px;
+      }
+
+      /* Hide custom elements that are not defined yet */
+      media-control-bar :not(:defined) {
```
heff, line 15: I did not realize you could do this. Cool.

src/js/media-container.js
```diff
@@ -43,11 +44,6 @@ template.innerHTML = `
       pointer-events: auto;
     }
     
-    ::slotted([slot=poster]) {
```
heff, line 46: Does this not break the poster sizing to the full container?

src/js/media-container.js
```diff
@@ -66,21 +62,10 @@ template.innerHTML = `
     }
 
     /* Video specific styles */
-    :host(:not([audio])) {
-      aspect-ratio: var(--media-aspect-ratio, auto 3 / 2);
```
heff, line 70: So by removing this, we're saying "just use the `aspect-ratio` CSS, don't bother with a css var", right?

heff (approved):
I called out a few things, but if those don't raise any flags then this looks good.

src/js/media-container.js
cjpillsbury: nah, it was repetitively redundant

src/js/media-container.js
cjpillsbury: yup

PR #207, opened 2022-04-18
fix: make media-time-display not wrap when small

gkatsev:
media-time-display shows both the current or remaining time and the
duration separated by a slash. Around the slash, there are spaces, so,
when the control bar is small, this causes the element to wrap onto
multiple lines. This can cause unexpected styling issues with the
control bar.

The expectation of this element is to stay on one-line like the other
time displays.

While we could add this into media-text-display, it's already used for
non time displays in the examples, though, internally in media-chrome,
media-text-display is only used in the time displays.

Before:
localhost_8000_examples_advanced html

After
localhost_8000_examples_advanced html (1)
:

### DIFF
--- a/src/js/media-time-display.js
+++ b/src/js/media-time-display.js
@@ -5,7 +5,7 @@ import { MediaUIAttributes } from './constants.js';
 import { nouns } from './labels/labels.js';
 // Todo: Use data locals: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleTimeString
 
-const DEFAULT_TIMES_SEP = ' / ';
+const DEFAULT_TIMES_SEP = '&nbsp;/&nbsp;';
 
 const formatTimesLabel = (el, { timesSep = DEFAULT_TIMES_SEP } = {}) => {
   const showRemaining = el.getAttribute('remaining') != null;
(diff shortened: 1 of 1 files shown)


src/js/media-time-display.js
```diff
@@ -52,6 +52,12 @@ const updateAriaValueText = (el) => {
 };
 
 class MediaTimeDisplay extends MediaTextDisplay {
+  constructor() {
+    super();
+
+    this.shadowRoot.styleSheets[0].addRule(':host', 'white-space: nowrap');
```
gkatsev, line 58: the way that the elements are currently set up in media-chrome, there isn't a way to add extra css to be rendered into the template in the constructor.

This adds an extra css rule after media-text-display has rendered itself.

gkatsev: While this works, &nbsp; is simpler still.

src/js/media-time-display.js
```diff
@@ -5,7 +5,7 @@ import { MediaUIAttributes } from './constants.js';
 import { nouns } from './labels/labels.js';
 // Todo: Use data locals: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleTimeString
 
-const DEFAULT_TIMES_SEP = ' / ';
+const DEFAULT_TIMES_SEP = '&nbsp;/&nbsp;';
```
cjpillsbury, line 8: nice! let's do it.

cjpillsbury (approved):

PR #209, opened 2022-04-22
feat: export MediaTheme w/ overrideable render

luwes:
This change exports the base MediaTheme class and adds a render method that can be overridden.

In the simplest case theme devs just provide a string template to their class that extends MediaTheme.

This change also fixes up the Netflix media theme and uses that in the examples. Good for testing.

```js
class MediaTheme extends HTMLElement {
  static template = '';

  constructor() {
    super();
    this.attachShadow({ mode: 'open' });
  }

  connectedCallback() {
    this.render();
  }

  render() {
    this.shadowRoot.textContent = '';

    template.innerHTML = this.constructor.template;

    // Clone the template in the shadow dom
    this.shadowRoot.append(template.content.cloneNode(true));

    // Expose the media controller if API access is needed
    this.mediaController = this.shadowRoot.querySelector('media-controller');
  }
}

if (!customElements.get('media-theme')) {
  customElements.define('media-theme', MediaTheme);
}

export default MediaTheme;
```

https://media-chrome-git-fork-luwes-media-theme-base-mux.vercel.app/examples/themes/netflix-theme.html

```js
const template = `
<style>...</style>
<media-controller>
  <slot name="media" slot="media"></slot>
  ...
</media-controller>
`;


class MediaThemeNetflix extends MediaTheme {
  static template = template;
}

if (!customElements.get('media-theme-netflix')) {
  customElements.define('media-theme-netflix', MediaThemeNetflix);
}

export default MediaThemeNetflix;
```

### DIFF
--- a/src/js/themes/media-theme.js
+++ b/src/js/themes/media-theme.js
@@ -1,28 +1,26 @@
 class MediaTheme extends HTMLElement {
-  constructor(template = ``, options = {}) {
+  static template = '';
+
+  constructor() {
     super();
+    this.attachShadow({ mode: 'open' });
+  }
 
-    options = Object.assign(
-      {
-        // Default options
-      },
-      options
-    );
+  connectedCallback() {
+    this.render();
+  }
 
-    // Expose the template publicaly for other uses
-    this.template = template;
+  render() {
+    this.shadowRoot.textContent = '';
 
-    // Not sure if this is best practice or if we should just
-    // innerHTML the template string in the shadow dom
-    const templateEl = document.createElement('template');
-    templateEl.innerHTML = template;
+    const template = document.createElement('template');
+    template.innerHTML = this.constructor.template;
 
     // Clone the template in the shadow dom
-    const shadow = this.attachShadow({ mode: 'open' });
-    shadow.appendChild(templateEl.content.cloneNode(true));
+    this.shadowRoot.append(template.content.cloneNode(true));
 
     // Expose the media controller if API access is needed
-    this.mediaController = shadow.querySelector('media-controller');
+    this.mediaController = this.shadowRoot.querySelector('media-controller');
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/js/themes/media-theme.js
```diff
@@ -1,28 +1,26 @@
...
-      options
-    );
+  connectedCallback() {
+    this.render();
+  }
 
-    // Expose the template publicaly for other uses
-    this.template = template;
+  render() {
+    this.shadowRoot.textContent = '';
```
cjpillsbury, line 14: any reason this is necessary?

src/js/themes/media-theme.js
```diff
@@ -1,28 +1,26 @@
...
+    template.innerHTML = this.constructor.template;
 
     // Clone the template in the shadow dom
-    const shadow = this.attachShadow({ mode: 'open' });
-    shadow.appendChild(templateEl.content.cloneNode(true));
+    this.shadowRoot.append(template.content.cloneNode(true));
 
     // Expose the media controller if API access is needed
-    this.mediaController = shadow.querySelector('media-controller');
+    this.mediaController = this.shadowRoot.querySelector('media-controller');
```
cjpillsbury, line 23: blocking: let's move this to its own prototype method or getter.

cjpillsbury (changes requested):
One small ask and one non-blocking ask, otherwise 😍

src/js/themes/media-theme.js
luwes: yes, to clear the existing DOM nodes if this.render() would be called more than once.

src/js/themes/media-theme.js
luwes: 👍

src/js/themes/media-theme.js
cjpillsbury: ok cool that makes perfect sense and is a hard problem longer term. still non-blocking but we'll probably need to work this out if each re-render will completely blow away the dom and re-write it from scratch. May need separate methods here for "initializing" the shadowRoot vs. "re-rendering" (may not, just thinking about an architecture for "updates over time" esp. in relation to DOM element instances and e.g. our template-parts architecture for player.

cjpillsbury (approved):
🎉

heff (approved):

Issue #210, opened 2022-04-22
HTMLElement is not defined in Next.js 12 app


daletom:
Tried adding Media Chrome to a next.js 12 app. Getting a HTMLElement is not defined error.  I copied this from the next demo for Media Chrome.  Is this something I am doing wrong? 
 I Used the following code: 

```
import MediaController from 'media-chrome/dist/react';

export default function Home() {
  return (
    <div>
      <MediaController>
      <video
            slot="media"
            src="https://stream.mux.com/DS00Spx1CV902MCtPj5WknGlR102V5HFkDe/high.mp4"
            preload="auto"
            muted
            crossOrigin=""
          >
          </video>  
      </MediaController>
    </div>
  )
}
```

Getting this error:

Server Error
ReferenceError: HTMLElement is not defined

This error happened while generating the page. Any console logs will be displayed in the terminal window.
Call Stack
eval
node_modules/media-chrome/dist/themes/media-theme.js (1:168)
Object../node_modules/media-chrome/dist/themes/media-theme.js
file:///Users/tom/Documents/GitHub/training/.next/server/pages/index.js (422:1)
__webpack_require__
file:///Users/tom/Documents/GitHub/training/.next/server/webpack-runtime.js (33:42)
eval
webpack-internal:///./node_modules/media-chrome/dist/index.js (59:81)
Object../node_modules/media-chrome/dist/index.js
file:///Users/tom/Documents/GitHub/training/.next/server/pages/index.js (132:1)
__webpack_require__
file:///Users/tom/Documents/GitHub/training/.next/server/webpack-runtime.js (33:42)
eval
webpack-internal:///./node_modules/media-chrome/dist/react/index.js (32:67)
Object../node_modules/media-chrome/dist/react/index.js
file:///Users/tom/Documents/GitHub/training/.next/server/pages/index.js (412:1)
__webpack_require__
file:///Users/tom/Documents/GitHub/training/.next/server/webpack-runtime.js (33:42)
eval
webpack-internal:///./pages/index.js (11:81)

cjpillsbury:
Thanks @daletom looks like a recent feature addition broke our SSR. Should have a fix asap!

PR #211, opened 2022-04-22
fix(mux-themes-use-polyfills): Make sure we use polyfills for themes.

cjpillsbury:


gkatsev (approved):

dylanjha (approved):
good catch!

dylanjha:
closes (#210 )

PR #212, opened 2022-04-26
Feat/media play gesture

cjpillsbury:
Basic implementation of `<media-chrome-gesture/>` and `<media-play-gesture/>`.

Supports specifying the pointer-types you want to include (default is all)

Things to discuss (either as blockers for merge or followups):
- do we want to implement any particular sizing behavior, either when slotted in `gestures-chrome` or "baked in" to the component?
- should we try to "pre-slot" this in `MediaContainer`?
- do we want to include any additional gestures now-ish?

Blockers before final merge:
- docs updates
- more examples updates


src/js/media-chrome-gesture.js
```diff
@@ -0,0 +1,124 @@
...
+    this._pointerType = undefined;
+    const pointerDownHandler = (e) => {
+      // Since not all browsers have updated to be spec compliant, where 'click' events should be PointerEvents,
+      // we can use use 'pointerdown' to reliably determine the pointer type. (CJP).
+      this._pointerType = e.pointerType;
+    };
+
+    this.addEventListener('pointerdown', pointerDownHandler);
+
+    this.addEventListener('click', (event) => {
```
cjpillsbury, line 59: we could do a "fancier"/more complex version of this that uses the first `"pointerdown"`+`"click"` to determine if `"pointerdown"` is necessary, but I figured I wouldn't over complicate this as a first pass.

heff:
Cool, thanks for kicking this off with an example. I'm trying to wrap my head around what different people want out of this feature, so any more detail on the assumptions here would be helpful.

My first assumption is 80+% of MC users just want us to solve this for them by default, as a feature of media-container, based on common viewer expectations, with an escape hatch if they have different opinions. User activity showing controls feels like an existing example of this. I think the default then looks like no change to the basic controls example and you get basic things like click-video-to-play on web and touch-video-to-toggle-controls on mobile. So clearly solving that as part of this feels important.

So, yes to "should we try to "pre-slot" this in MediaContainer?", if that's the best way to provide the default. I think it's intuitive that you could override the default slotted element(s) for advanced uses, like how we do with button content today.

From there I can see a few differently levels of abstraction we could come at this with, so interested to hear opinions on why a specific level is best. I think the `media-chrome-gesture` concept makes sense as an analog to `media-chrome-button`.

I see where you're going with `media-play-gesture` as away to build in higher level gesture components, and I like it, but also what is a "play gesture"? That doesn't read as clearly as "media play button". I could see an alternative build being like:
```html
<media-chrome-gesture type="swipe-left" request="seek" detail="30">
<media-swipe-left-gesture request="seek" detail="30">
<media-seek-swipe seconds="30">
```

> do we want to implement any particular sizing behavior, either when slotted in gestures-chrome or "baked in" to the component?

Is "gesture" defined as "a gesture made on the video content itself (other than on controls)", or do gestures have optional sizing? i.e. Can you have a left side gesture and a right side gesture? I'm assuming the former and that feels like a clearer expectation. But I can see a click on the left seeking back and click on the right seeking forward. That feels like just two transparent buttons though rather than a gesture. What's the different between a button click and gesture click?

luwes:
> no change to the basic controls example and you get basic things like click-video-to-play on web and touch-video-to-toggle-controls on mobile

def agree with this, should be enabled with no changes to the user markup. 

> What's the different between a button click and gesture click?

this makes me think the `<x-gesture>` naming is not really right. I don't think I've seen any html element with a... crap `<blink>` haha 

this new element seems like it actually could / should be a part of the basic makeup of a chrome button / play button.
I think configurable sizing of this element makes sense. naming wise "target" or "area" seems more appropriate

cjpillsbury:
I'm down with extending & potentially updating `media-chrome-button` and/or `media-play-button` (per out of band convo, this is what I started with anyway, so def makes sense to me). I do think whatever route we go this should be:
- easy/intuitive to use for this kind of use case (a large transparent click region for a first pass)
- easy/intuitive to configure based on the "pointerType" (I'll assume my approach for now, but am down for other approaches)
- easy/intuitive to disable the "default slot" usage in `<media-controller/>` (This should be achievable through an attribute or CSS var)

I can start noodling and get a new example/POC for discussion up. I strongly suspect the primary clunkiness is going to be divergent styling/UIs as a result of the inheritance model. I'll assume an analogous strategy to media-container video vs. audio for now.

wrt naming, I'll noodle with things, but @heff's question and @luwes's followup to
> What's the different between a button click and gesture click?
plus the refactor makes me think we should use an attribute instead of a whole new component. Something like:

```html
<media-play-button as-region></media-play-button>
```

Thoughts?

luwes:
indeed it might be the easiest to get this working with not too much changes, or it could just be added as default logic to media-container. initially I think there's nothing wrong with that.

`pointer-types`, `as-region` feel a bit strange to me. what would as-region do?
pointer-types="mouse" but mobile devices with touch still listen to mouse events, that's a bit confusing to devs I think.

this issue made think about good old [`hitArea`](https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/display/Sprite.html#hitArea) 😄 where is the time.

also a bit related hotspots for interactive video like https://www.wirewax.com/
but that would require coordinates, shapes, hide/show for certain times etc. probably a different beast.

gkatsev:
For the super simple base case of play toggle on click, I made https://github.com/muxinc/media-chrome/pull/213. We already had code that was almost there for toggling playback, just needed to be fixed up a bit.

src/js/media-display-gesture-receiver.js
```diff
@@ -0,0 +1,142 @@
...
+      // NOTE: Longer term, we'll likely want to delay this to support double click/double tap (CJP)
+      if (pointerType === PointerTypes.TOUCH) {
+        this.handleTap(event);
+        return;
+      } else if (pointerType === PointerTypes.MOUSE) {
+        this.handleMouseClick(event);
+        return;
+      } else if (pointerType === PointerTypes.PEN) {
+        this.handlePenTap(event);
+        return;
```
luwes, line 79: might be a little early for this use case, are we going to do special things for touch pens?

cjpillsbury: just wanted to make sure it's accounted for somewhere since it's the only other well-defined pointerType, but happy to remove it for now if you're concerned?

src/js/media-display-gesture-receiver.js
```diff
@@ -0,0 +1,142 @@
...
+      this._pointerType = e.pointerType;
+    };
+
+    this.addEventListener('pointerdown', pointerDownHandler);
+
+    this.addEventListener('click', (event) => {
+      if (!event.pointerType) {
+        event.pointerType = this._pointerType;
+      }
+      const { pointerType } = event;
```
gkatsev, line 63: nit: we can set a default value via destructuring so that we don't modify the native event object.
```suggestion
      const { pointerType = this._pointerType } = event;
```

gkatsev (changes requested):

src/js/media-display-gesture-receiver.js
cjpillsbury: I was debating that. I thought it would be good to pass the "expected" pointerType to the handler. Thoughts?

src/js/media-display-gesture-receiver.js
luwes: concern is YAGNI

cjpillsbury: sold. Removing.

src/js/media-display-gesture-receiver.js
```diff
@@ -0,0 +1,134 @@
...
+        ? MediaUIEvents.MEDIA_PLAY_REQUEST
+        : MediaUIEvents.MEDIA_PAUSE_REQUEST;
+    this.dispatchEvent(
+      new window.CustomEvent(eventName, { composed: true, bubbles: true })
+    );
+  }
+}
+
+defineCustomElement(
+  'media-display-gesture-receiver',
```
luwes, line 130: any chance this element name could be made shorter? 

I know the display here is the video display but it could potentially be read as this element is displaying a gesture receiver which is a bit strange :)
> Use `-display` when the primary use is displaying a state or data detail. e.g. `media-duration-display`.

maybe just `media-gesture-receiver`? or should it be more specific?

---

I was looking up some synonyms and stuff for receiver and `media-gesture-control` came up. 
it puts it more in the category of one of the media controls.
Receiver for me feels more like it implies a sender and a time where something is transmitting but might be overthinking it 😅

cjpillsbury: This was based on an "off the dome" name proposal from Heff. Since we're not really advertising it just yet, I'm cool with renaming it while it's still tucked away as an internal if you feel strongly. `media-gesture-receiver` and `media-gesture-control` both sound like reasonable names to me. Preference?

luwes (approved):
👍 LGTM

gkatsev (approved):

PR #213, opened 2022-05-03
feat: re-enable media click to play toggle

gkatsev:
This enables clicking on the media element to toggle playback. This
works on both desktop and mobile/touch devices. Eventually, we'd probably want
to make mobile/touch devices to toggle controls showing on taps (except
for the first one).

One thing that's worth noting, in the mobile example, if during
playback, you click the play toggle, the controls will stay showing.
This is actually an existing issue relating to us keeping controls
showing when focus is on a control bar control (like the
`media-play-button`).

### DIFF
--- a/src/js/media-container.js
+++ b/src/js/media-container.js
@@ -232,11 +232,11 @@ class MediaContainer extends window.HTMLElement {
         new window.CustomEvent(eventName, { composed: true, bubbles: true })
       );
     };
+    media.addEventListener('click', this._mediaClickPlayToggle, false);
   }
 
   handleMediaUpdated(media) {
     const resolveMediaPromise = (media) => {
-      // media.addEventListener('click', this._mediaClickPlayToggle, false);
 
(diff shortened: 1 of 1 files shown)


src/js/media-container.js
```diff
@@ -232,11 +232,11 @@ class MediaContainer extends window.HTMLElement {
         new window.CustomEvent(eventName, { composed: true, bubbles: true })
       );
     };
+    media.addEventListener('click', this._mediaClickPlayToggle, false);
   }
 
   handleMediaUpdated(media) {
     const resolveMediaPromise = (media) => {
-      // media.addEventListener('click', this._mediaClickPlayToggle, false);
```
gkatsev, line 239: `mediaSetCallback` is called after `handleMediaUpdated`, so, `this._mediaClickPlayToggle` was never set. Instead, we can move this listener into `mediaSetCallback` to match it being removed from `mediaUnsetCallback`

PR #214, opened 2022-05-03
Fix isFullscreen check for deep shadow nesting

luwes:
This change fixes an issue where if the media-controller was contained in multi-level shadow DOM the isFullscreen check would fail. 

Safari doesn't support ShadowRoot.fullscreenElement, the solution here uses document.fullscreenElement to check if any element in the document is fullscreen and then event.target to check if the target is contained or is itself the media-controller.

This PR also improves the contains utility function `containsComposedNode` to be more efficient.

### DIFF
--- a/src/js/utils/element-utils.js
+++ b/src/js/utils/element-utils.js
@@ -15,10 +15,8 @@ export const getAllSlotted = (el, name) => {
 
 export const getSlotted = (el, name) => getAllSlotted(el, name)[0];
 
-export const containsWithShadow = (refNode, otherNode) => {
-  if (!refNode || !otherNode) return false;
-  if (refNode.contains(otherNode)) return true;
-  return [refNode, ...refNode.querySelectorAll("*")].some((el) => {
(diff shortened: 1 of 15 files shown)


src/js/utils/element-utils.js
```diff
@@ -15,10 +15,8 @@ export const getAllSlotted = (el, name) => {
...
-  if (!refNode || !otherNode) return false;
-  if (refNode.contains(otherNode)) return true;
-  return [refNode, ...refNode.querySelectorAll("*")].some((el) => {
-    return containsWithShadow(el.shadowRoot, otherNode);
-  });
-}
+export const containsComposedNode = (rootNode, childNode) => {
+  if (!rootNode || !childNode) return false;
+  if (rootNode.contains(childNode)) return true;
+  return containsComposedNode(rootNode, childNode.getRootNode().host);
```
cjpillsbury, line 21: cool I like this approach (recursively working outward instead of inward)!

cjpillsbury (commented):
🚢 🇮🇹

PR #216, opened 2022-05-10
Add custom translate CSS var for input range track

luwes:
Also added a `range` part attribute to target the input range element if needed.
In the player we might need this to set the input to display block.

The separation of the translate coords in x and y is so we don't have to use javascript.

- The CSS `translate` prop expects coords separated with a space.
- The transform: translate() function expects commas.

Also a restriction to only `translate` for now, `scale` and `rotate` would make it more complex because Firefox has the thumb not relative to the track. It's a sibling element instead of a child element which makes it a lot harder to do scaling or rotating and keeping that consistent between Firefox and the rest of the browsers.


src/js/media-chrome-range.js
```diff
@@ -121,9 +127,8 @@ template.innerHTML = `
     input[type=range]:disabled::-webkit-slider-runnable-track {
       background-color: #777;
     }
-
   </style>
-  <input id="range" type="range" min="0" max="1000" step="1" value="0">
+  <input id="range" type="range" part="range" min="0" max="1000" step="1" value="0">
```
heff, line 131: We probably don't want to do this right now. @cjpillsbury believes we'll need to switch to a non-native range implementation at some point, and this would make that a breaking change compared to exposing css vars we can translate between implementations.

heff (commented):
Other than the part concern this looks good. 👍

src/js/media-chrome-range.js
luwes: makes sense, I'll remove this part attribute.

cjpillsbury (approved):
May want to account for this technically being a "breaking change" for anyone relying on the renamed css var in the next release. Alternatively, we could keep the old var as one that's deprecated. Fine with either approach, just flagging this.

PR #217, opened 2022-05-12
fix: Account for cases where the media-controller instance is part of…

cjpillsbury:
… a custom element's shadow DOM in fullscreen mode.


cjpillsbury:
Latest not released. False alarm. Nothing to see here.

PR #220, opened 2022-05-17
feat: add cast button and castable video elements

luwes:
This PR adds a new `media-cast-button` and logic to handle show/hide the button.

---

Requires https://github.com/muxinc/castable-video 

An enhanced video element with casting logic.
Most of the casting logic lives in this custom element built-in extending the `HTMLVideoElement`.
It's called `CastableVideoElement` w/ tag name `castable-video` open to other suggestions. 

It's inspired by this implementation https://github.com/GoogleWebComponents/google-castable-video
but the API is cleaner and made more uniform to the native `HTMLVideoElement` API making it Media Chrome friendly.
Also looked at many other libraries, Shaka Player, Castjs, videojs Chromecast, etc.

It's a custom element built-in because Chrome browsers support it and it doesn't break anything on Safari.
Using the CustomVideoElement approach would have required more code and potential for bugs.

While building this I took the view of if casting would be a native API, how would that look like and held the Airplay API next to it. Result is a pretty simple and small API that should work for most of the use cases.

The video element has a few extra attributes, props, method, event;

- `castchange`: event that fires when cast devices are available
- `entercast`
- `leavecast`
- `cast-src`: attribute which could be needed if MediaSource is used and the video[src] is a blob. If not set it will take the currentSrc for casting.
- `stream-type`: needed for determining live vs on-demand content. defaults to on-demand
- `content-type`: might be needed to determine the video content-type but file extensions so far seem to do the trick.
- `video.requestCast(options)`: opens the browser cast menu to start casting to a device.
- `CastableVideoElement.castElement`: currently casted media element 
- `CastableVideoElement.castEnabled`: is set to true when the casting library is loaded.
- `CastableVideoElement.exitCast()`: exit the current cast 

These last 3 API's could potentially be set on the `document`, it's worth a discussion. It would remove the variable of a named class `CastableVideoElement` in Media Chrome code.

The rest of the video methods/getters/setters/events like play(), paused, currentTime, volume, duration, etc are seamlessly switched between local and remote playback in the enhanced video element.

Of course this change will require a non-standard API to be added to Media Chrome, there's no way around that I think.

### TODO

- [x] Test live playback and make it work
- [x] Add support for text tracks


src/js/media-controller.js
```diff
@@ -325,6 +347,21 @@ class MediaController extends MediaContainer {
...
         this.propagateMediaState(MediaUIAttributes.MEDIA_IS_PIP, isPip);
       },
+      // Note this relies on a customized video[is=castable-video] element.
+      'entercast,leavecast': (e) => {
+        const isCast = this.media && globalThis.CastableVideo?.castElement === this.media;
+
+        // Prevent auto hiding controls while casting.
+        if (isCast && this.autohide >= 0) {
+          this._prevAutohide = this.autohide;
+          this.setAttribute('autohide', '-1');
```
cjpillsbury, line 357: blocker: Our pre-existing convention is to handle this stuff in media-container's code for activity monitoring & related css, since `autohide` may be set by implementors. See, e.g. the way we handle checking for paused state.

src/js/media-controller.js
```diff
@@ -325,6 +347,21 @@ class MediaController extends MediaContainer {
...
+          this.removeAttribute('user-inactive');
+        } else if (this._prevAutohide >= 0) {
+          this.setAttribute('autohide', this._prevAutohide);
+        }
+
+        this.propagateMediaState(MediaUIAttributes.MEDIA_IS_CAST, isCast);
```
cjpillsbury, line 363: non-blocker: consider refactoring this to handle multiple well-defined states instead of on/off to support "fancier" UIs in the future per Google design recommendations (see e.g. https://developers.google.com/cast/docs/design_checklist/cast-button#sender-cast-icon-states).

src/js/media-controller.js
```diff
@@ -171,6 +174,25 @@ class MediaController extends MediaContainer {
...
+          document[fullscreenApi.exit]();
+        }
+
+        // Open the browser cast menu.
+        // Note this relies on a customized video[is=castable-video] element.
+        media.requestCast();
+      },
+      MEDIA_EXIT_CAST_REQUEST: async () => {
+        if (CastableVideo.castElement) {
+          CastableVideo.exitCast();
```
cjpillsbury, line 193: potential blocker: confirm this won't have issues with multiple `CastableVideo`s on the same page.

cjpillsbury (changes requested):
A few callouts, but overall this is looking great.

src/js/media-controller.js
luwes: you're right here I think but we're doing this for pip, fullscreen, etc all like this.
it's fairly safe because these exit requests are only made from a button that is in the exit state.
we should revisit it for all but this will be the same difficulty with the nested shadow dom etc.
prefer to wait a little because it's unlikely to reproduce this bug

src/js/media-controller.js
```diff
@@ -377,6 +414,22 @@ class MediaController extends MediaContainer {
         airplaySupporHandler;
     }
 
+    if (this._castUnavailable !== AvailabilityStates.UNSUPPORTED) {
+      const castSupportHandler = (event) => {
+        if (event?.detail !== 'NO_DEVICES_AVAILABLE') {
+          this._castUnavailable = undefined;
+        } else if (event?.detail === 'NO_DEVICES_AVAILABLE') {
```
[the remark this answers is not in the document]
luwes: changed this to a if / else

src/js/media-controller.js
luwes: good call, worked with just css in media-container

src/js/media-controller.js
luwes: I added the "connecting" value to media-is-cast for media state receivers and added the original Google svg's with classes for the wifi arches in the icon. only in my testing this connecting state only took a split second. I tested the <google-cast-launcher> button and it also didn't show the animation on my end. maybe my wifi is too good :P 

I'll look into animating the SVG today, in a 3rd icon <slot> I imagine for it to be replaceable

src/js/media-controller.js
cjpillsbury: ok I'm cool with treating this as a fast follow 👌. Let's make sure we get an issue filed for this.

src/js/media-cast-button.js
```diff
@@ -0,0 +1,80 @@
...
+    display: none;
+  }
+  </style>
+
+  <slot name="enter">${enterIcon}</slot>
+  <slot name="exit">${exitIcon}</slot>
+`;
+
+const updateAriaLabel = (el) => {
+  const isFullScreen =
```
cjpillsbury, line 34: nit: copypasta - `isCast` 😅

src/js/media-controller.js
```diff
@@ -171,6 +174,25 @@ class MediaController extends MediaContainer {
           document.exitPictureInPicture();
         }
       },
+      MEDIA_ENTER_CAST_REQUEST: () => {
+        const media = this.media;
+
+        if (!CastableVideo?.castEnabled) return;
+
+        // Exit fullscreen if needed
+        if (document[fullscreenApi.element]) {
```
cjpillsbury, line 183: 👌

src/js/media-controller.js
```diff
@@ -171,6 +174,25 @@ class MediaController extends MediaContainer {
...
+          document[fullscreenApi.exit]();
+        }
+
+        // Open the browser cast menu.
+        // Note this relies on a customized video[is=castable-video] element.
```
cjpillsbury, line 188: 👌

cjpillsbury (approved):
Pre-approved but with one nit blocker. Feel free to merge once it's ready!

src/js/constants.js
```diff
@@ -32,6 +34,7 @@ export const MediaStateChangeEvents = {
   MEDIA_VOLUME: 'mediavolumechange',
   MEDIA_VOLUME_UNAVAILABLE: 'mediavolumeunavailablechange',
   MEDIA_IS_PIP: 'mediaispipchange',
+  MEDIA_IS_CAST: 'mediaiscastchange',
```
heff, line 37: I see how we got here with IS_PIP, but IS_CAST is awkward phrasing. I think we need to consider _casting_ or _casted_ or something similar. I'm not sure what's best off the top of my head.

src/js/labels/labels.js
```diff
@@ -18,6 +18,8 @@ export const verbs = {
   MUTE: () => 'mute',
   UNMUTE: () => 'unmute',
   AIRPLAY: () => 'air play',
+  ENTER_CAST: () => 'enter cast mode',
```
heff, line 21: Similar comment here. Is "enter cast" the right phrasing? Something like "start casting" sounds a little better to me.

heff (commented):
This is really cool. 👍  My big question here is what happens when we want to add cast-ability to other custom media? Like mux-video and hls-video. It feels limiting to require a custom media element, but maybe it's not.

The `stream-type` and `content-type` names feel prone for future conflicts if we're extending the video element API with them. If we feel super confident in their definitions that could be fine, but otherwise we might consider prefixing them, like `cast-stream-type`.

src/js/constants.js
luwes: renamed it to MEDIA_IS_CASTING

src/js/labels/labels.js
luwes: renamed to start / stop casting

luwes:
> My big question here is what happens when we want to add cast-ability to other custom media? Like mux-video and hls-video. It feels limiting to require a custom media element, but maybe it's not.

@heff could you elaborate on this one? 

how I saw this working was to have the `<video>` element in those CustomVideoElement's like `mux-video` and `hls-video` be a `<video is="castable-video">` when needed. Something like

```html
<hls-video>
  <video is="castable-video" slot="media">
</hls-video>
```

A bit like you explained here: https://github.com/muxinc/custom-video-element/issues/6
We should definitely support a media slot in CustomVideoElement, also for progressive enhancement.

`cast-stream-type`, `cast-content-type` makes sense. I'll change that.

luwes:
thanks! I'll do the docs in a follow up PR to first verify things work in mux-player as expected

PR #221, opened 2022-05-18
feat(seekable-ranges): Add media-seekable/media.seekable state suppor…

cjpillsbury:
…t. Update media-time-display and media-time-range to handle seekable.

- Add state management for seekable (start & end) to `<media-controller/>`
- Add initial support for seekable start/end to `<media-time-range/>` and `<media-time-display/>` (note: no support for *both* seekable.end and (expected) duration (yet!)
- Update range and related math to no longer assume input.min will always be 0.

relates to #218


src/js/media-controller.js
```diff
@@ -682,9 +693,17 @@ const getCurrentTime = (controller) => {
 };
 
 const getDuration = (controller) => {
-  const media = controller.media;
+  const media = controller?.media;
+  if (!Number.isFinite(media?.duration)) return NaN;
```
luwes, line 697: is this to handle `Infinity`?

src/js/media-time-range.js
```diff
@@ -137,61 +145,58 @@ class MediaTimeRange extends MediaChromeRange {
...
+    const [start] = this.mediaSeekable ?? [];
+    return start;
   }
 
   /* Add a buffered progress bar */
   getBarColors() {
     let colorsArray = super.getBarColors();
+    const { range } = this;
+    // Use mediaCurrentTime instead of range.value for precision (CJP)
+    const value = this.mediaCurrentTime;
```
luwes, line 186: can we use `currentTime` instead of `value` for naming?

luwes (approved):
a small nit but LGTM

src/js/media-controller.js
cjpillsbury: yup, `Infinity` and any non-numerics all in one fell swoop.

PR #222, opened 2022-05-19
fix: conslidate and simplfy selectors dealing with pointer-events

gkatsev:
Should be merged via a squash merge, probably.


cjpillsbury (approved):
👌

PR #224, opened 2022-05-20
Fix/default UI times

cjpillsbury:
Since refactoring to support seekable ranges, we can no longer "default" the duration values. Per suggestion from @luwes , for now we can just assign default values for less confusing visualization.


luwes (approved):
👍

PR #225, opened 2022-05-20
fix: mouseout timerange for deeply nest shadowdom

luwes:
Fixes an issue where the off range logic was called too early for Media Chrome in a shadow dom of another element.


cjpillsbury (approved):
🏄‍♂️

PR #227, opened 2022-05-23
refactor: delegate all media ui attributes

gkatsev:
This implements the TODO to automatically propagate all state. Uses a delegate object that matches the MediaUIAttributes in case we need to do something special, otherwise, it'll use the `default` method which returns the attribute's value.

### DIFF
--- a/src/js/media-controller.js
+++ b/src/js/media-controller.js
@@ -732,6 +732,9 @@ class MediaController extends MediaContainer {
 
     if (this.media) {
       Object.keys(MediaUIAttributes).forEach((attribute) => {
+        // skip availability delegates as they were completed above
+        if (attribute.includes('UNAVAILABLE')) return;
+
         propagateMediaState(
           [el],
           MediaUIAttributes[attribute],
--- a/src/js/media-controller.js
+++ b/src/js/media-controller.js
@@ -730,67 +730,14 @@ class MediaController extends MediaContainer {
       this._pipUnavailable,
     );
 
-    // TODO: Update to propagate all states when registered
     if (this.media) {
-      propagateMediaState(
-        [el],
-        MediaUIAttributes.MEDIA_CAPTIONS_LIST,
-        stringifyTextTrackList(getCaptionTracks(this)) || undefined
-      );
-      propagateMediaState(
-        [el],
-        MediaUIAttributes.MEDIA_SUBTITLES_LIST,
-        stringifyTextTrackList(getSubtitleTracks(this)) || undefined
-      );
-      propagateMediaState(
-        [el],
-        MediaUIAttributes.MEDIA_CAPTIONS_SHOWING,
-        stringifyTextTrackList(getShowingCaptionTracks(this)) || undefined
-      );
-      propagateMediaState(
-        [el],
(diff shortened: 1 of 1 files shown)


src/js/media-controller.js
```diff
@@ -580,62 +580,14 @@ class MediaController extends MediaContainer {
...
-      propagateMediaState(
-        [el],
-        MediaUIAttributes.MEDIA_PLAYBACK_RATE,
-        getPlaybackRate(this)
-      );
+      Object.keys(MediaUIAttributes).forEach((attribute) => {
+        propagateMediaState(
+          [el],
+          MediaUIAttributes[attribute],
+          Delegates[attribute] ? Delegeates[attribute](this) : Delegates.default(this, MediaUIAttributes[attribute])
```
luwes, line 588: `Delegeates` typo, I think this repo really needs eslint. would help a bit

gkatsev: oops. Yeah, linting would be helpful to catch silly mistakes like this.

src/js/media-controller.js
```diff
@@ -580,62 +582,14 @@ class MediaController extends MediaContainer {
...
-      propagateMediaState(
-        [el],
-        MediaUIAttributes.MEDIA_PLAYBACK_RATE,
-        getPlaybackRate(this)
-      );
+      Object.keys(MediaUIAttributes).forEach((attribute) => {
+        propagateMediaState(
+          [el],
+          MediaUIAttributes[attribute],
+          Delegates[attribute] ? Delegates[attribute](this) : Delegates.default(this, MediaUIAttributes[attribute])
```
gkatsev, line 590: an alternative here is to loop through MediaUIAttributes and if there isn't one in Delegates, we set it equal to `Delegates.default` then this line just becomes `Delegates[attribute](this)` (with default changes to accept the `attribute` as a first value:
```
Object.keys(MediaUIAttribute).forEach((attribute) => {
  if (Delegates.hasOwnProperty(attribute)) {
    Delegates[attribute] = Delegates.default.bind(attribute);
  }
});
```

cjpillsbury (approved):
Approved with some thoughts on followups.

src/js/media-controller.js
```diff
@@ -730,67 +730,14 @@ class MediaController extends MediaContainer {
...
-        [el],
-        MediaUIAttributes.MEDIA_SEEKABLE,
-        getSeekable(this)?.join(':')
-      );
-      propagateMediaState(
-        [el],
-        MediaUIAttributes.MEDIA_PLAYBACK_RATE,
-        getPlaybackRate(this)
-      );
+      Object.keys(MediaUIAttributes).forEach((attribute) => {
```
heff, line 734: Minor nit: we're now double propagating all the attrs like MEDIA_PIP_UNAVAILABLE (above this diff).

heff (approved):
This looks fine to me. 👍

src/js/media-controller.js
gkatsev: Good call. I guess I should skip those 🤔

cjpillsbury (approved):
works for me. Could also `.filter()` but your version is slightly more performant.

PR #231, opened 2022-05-26
feat: add exports package.json field

gkatsev:
This means that folks using CJS and a new enough node/bundler will
automatically get the CJS builds without requiring them to know that
there is a CJS build and doing a deep import.

The default remains the ESM build.

~This was originally a PR against https://github.com/muxinc/media-chrome/pull/195 but didn't make it in.~

~This makes this module a hybrid module with the main being a cjs file.~

~We should also update the react build to generate a cjs subfolder and have a package.json that splits imports between the esm and cjs files. (Also, if we generate it to the root rather than dist, the require/import could be `media-chrome/react` rather than `media-chrome/dist/react`~

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,11 @@
   "version": "0.10.4",
   "description": "Custom elements (web components) for making audio and video player controls that look great in your website or app.",
   "main": "dist/index.js",
-  "type": "module",
+  "exports": {
+    "require": "./dist/cjs/index.js",
+    "import": "./dist/index.js",
+    "default": "./dist/index.js"
+  },
   "files": [
     "dist/*",
     "README.md",


heff:
Can we get this one merged or close it? Based on comments from the other PR I think this is waiting on @luwes?

gkatsev:
What about keeping the `exports` piece? That way, CJS users don't have to know how this repo is set up, assuming they're using a new-enough node version that supports `exports`. Most bundlers will be fine with esm as `main`.
I do think that users of modules shouldn't need to know how a package was authored or built, and that it should just work.

luwes:
yes, the exports prop would be a plus 👍

gkatsev:
cool, I'll rebase and update this PR with that in mind.

gkatsev:
Ah, since we're removing `type: module` (node won't let you import cjs in that case) we probably need to switch to `.mjs` filetype.

gkatsev:
Changing to `.mjs` isn't too bad, but the react build is completely broken and not obvious how to fix. So, to use exports, I think we'd probably want to move ESM code to mjs files but then it gets tricky unless we switch to bundling because the imports/requires reference the files directly which then becomes tricky with updated extensions.

I'm going to close this PR since I don't have the time to look at it right now, but I think it's worth investigating. While ESM is the future, I do think we should still be supporting people *now* and there are those that still use CJS.

PR #232, opened 2022-05-26
fix: unhide controls on touch & autohide on timer

gkatsev:
This keeps desktop behavior for mobile/touch devices. When
tapping, it'll unhide the controls and then setup the timer to autohide.

`mousemove` triggers even when a user only does a single tap.
`pointermove` doesn't occur with just a single tap, so, we switch
`mousemove` to `pointermove`. Then, we schedule the inactivity timeout
on `pointerup` when `pointerType` is `touch`, since other pointer types
are covered.

### DIFF
--- a/src/js/media-container.js
+++ b/src/js/media-container.js
@@ -304,25 +304,35 @@ class MediaContainer extends window.HTMLElement {
     // This allows things like autoplay and programmatic playing to also initiate hiding controls (CJP)
     this.setAttribute('user-inactive', 'user-inactive');
 
-    const scheduleInactive = () => {
+    const setInactive = () => {
+      if (this.autohide < 0) return;
+      this.setAttribute('user-inactive', 'user-inactive');
+      const evt = new window.CustomEvent(
+        MediaStateChangeEvents.USER_INACTIVE,
+        { composed: true, bubbles: true, detail: true }
+      );
+      this.dispatchEvent(evt);
+    };
+
+    const setActive = () => {
       this.removeAttribute('user-inactive');
       const evt = new window.CustomEvent(
-        MediaStateChangeEvents.USER_INACTIVE, 
+        MediaStateChangeEvents.USER_INACTIVE,
         { composed: true, bubbles: true, detail: false }
       );
       this.dispatchEvent(evt);
+    }
+
+    const scheduleInactive = () => {
+      setActive();
+
       window.clearTimeout(this._inactiveTimeout);
 
       // Setting autohide to -1 turns off autohide
       if (this.autohide < 0) return;
 
       this._inactiveTimeout = window.setTimeout(() => {
-        this.setAttribute('user-inactive', 'user-inactive');
-        const evt = new window.CustomEvent(
-          MediaStateChangeEvents.USER_INACTIVE, 
-          { composed: true, bubbles: true, detail: true }
-        );
-        this.dispatchEvent(evt);
+        setInactive();
       }, this.autohide * 1000);
     };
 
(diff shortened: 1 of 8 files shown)


gkatsev:
This PR does the simplest change to fix autohiding on mobile.
One alternative solution is to implement toggling of the controls on taps with the following behavior:
- on initial tap (which plays the video), autohide on the timer
- on subsequent taps, toggle controls and do not autohide.

cjpillsbury (dismissed):
Cool this seems reasonable to me and smoke testing looked 👌.

src/js/media-container.js
```diff
@@ -304,51 +304,66 @@ class MediaContainer extends window.HTMLElement {
     // This allows things like autoplay and programmatic playing to also initiate hiding controls (CJP)
     this.setAttribute('user-inactive', 'user-inactive');
 
-    const scheduleInactive = () => {
+    const setInactive = () => {
```
cjpillsbury, line 307: nit (non-blocking): If we're going to model this as a property, we should probably just make it an actual property of the media-container.

src/js/media-container.js
```diff
@@ -304,51 +304,66 @@ class MediaContainer extends window.HTMLElement {
...
       this.removeAttribute('user-inactive');
       const evt = new window.CustomEvent(
-        MediaStateChangeEvents.USER_INACTIVE, 
+        MediaStateChangeEvents.USER_INACTIVE,
         { composed: true, bubbles: true, detail: false }
       );
       this.dispatchEvent(evt);
+    }
+
+    const autohide = () => {
```
cjpillsbury, line 326: nit (blocking): Rename. Having a function called `autohide` that refers to a numeric property called `autohide` will be confusing.

src/js/media-container.js
```diff
@@ -359,13 +374,15 @@ class MediaContainer extends window.HTMLElement {
...
-        MediaStateChangeEvents.USER_INACTIVE, 
-        { composed: true, bubbles: true, detail: true }
-      );
-      this.dispatchEvent(evt);
+      setInactive();
+    });
+
+    // Allow for focus styles only when using the keyboard to navigate
+    this.addEventListener('keyup', (e) => {
+      this.setAttribute('media-keyboard-control', 'media-keyboard-control');
```
luwes, line 382:
```suggestion
      this.setAttribute('media-keyboard-control', '');
```

for attr booleans this is the platform way

cjpillsbury (changes requested):
a couple of non-blocking callouts and one minor blocking ask.

cjpillsbury (changes requested):
Current implementation will initiate hiding controls immediately as a result of interacting with a control element, at least on touch screens (e.g. tapping the seek backward button).

src/js/media-container.js
```diff
@@ -304,51 +304,66 @@ class MediaContainer extends window.HTMLElement {
...
-          MediaStateChangeEvents.USER_INACTIVE, 
-          { composed: true, bubbles: true, detail: true }
-        );
-        this.dispatchEvent(evt);
+        setInactive();
       }, this.autohide * 1000);
+    }
+
+    const scheduleInactive = () => {
+      setActive();
```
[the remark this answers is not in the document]
gkatsev: `scheduleInactive` was an existing method that when an event happened like a mousemove, keyup, or touch would make the controls active and then schedule hiding the controls on the timer.
I moved setActive into a separate function because there were two places where the same code was being called.

src/js/media-container.js
gkatsev: Going to move autohide back into scheduleInactive as it's no longer needed as a separate function.

src/js/media-container.js
gkatsev: Sounds good. As an FYI, this is how it was previously.

cjpillsbury (approved):
😎

PR #235, opened 2022-05-31
fix: cast unavailable bug

luwes:
fixes a bug where the cast button wouldn't hide because the unavailable attribute was not set.


gkatsev (approved):

cjpillsbury (approved):
👌

PR #236, opened 2022-06-02
Text track renderer

gkatsev:


src/js/text-track-renderer/index.js
```diff
@@ -0,0 +1,137 @@
+import {
```
gkatsev, line 1: this file would eventually land into `extras` or could be another project, like in open-elements or something.

src/js/text-track-renderer/index.js
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  get mediaController() {
+    return this._mc;
+  }
+
+  setupCustomRendering() {
+    const mc = this._mc;
+
+    const [track] = getTextTracksList(mc.media, {kind: 'captions', mode: 'showing'});
```
heff, line 114: Instead of needing a direct reference to media controller and media here, could we instead push the cue details to the custom renderer in the same we we pass state to other components?

luwes:
closing as stale and has conflicts

PR #241, opened 2022-06-09
ci: add canary command

luwes:
Discussed in https://github.com/muxinc/media-chrome/discussions/239
Adds auto publish on the main branch with canary versions (`0.6.9-canary.0+17dac0e`) on the NPM `canary` release channel.

SCR-20220609-hvt

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,17 +1,21 @@
 name: CI
 
-on: [pull_request]
+on: [push, pull_request]
 
 jobs:
   build:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v3
         with:
           node-version: ${{ matrix.node-version }}
       - run: yarn install --frozen-lockfile
       - run: yarn test
       - run: yarn build
+      - run: yarn publish-canary
+        if: ${{ github.ref == 'refs/heads/main' }}
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,6 +12,8 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version: ${{ matrix.node-version }}
+          # this line is required for the setup-node action to be able to run the npm publish below.
+          registry-url: 'https://registry.npmjs.org'
       - run: yarn install --frozen-lockfile
       - run: yarn test
       - run: yarn build
--- a/package.json
+++ b/package.json
@@ -20,7 +20,8 @@
     "dev": "node ./scripts/esbuild.js",
     "start": "yarn dev",
     "test": "web-test-runner --coverage --config test/web-test-runner.config.js",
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -20,7 +20,8 @@
     "dev": "node ./scripts/esbuild.js",
     "start": "yarn dev",
     "test": "web-test-runner --coverage --config test/web-test-runner.config.js",
-    "publish-patch": "yarn run build && np patch --no-tests --any-branch"
+    "publish-patch": "yarn run build && np patch --no-tests --any-branch",
+    "publish-canary": "yarn run build && np $(semver $(npm pkg get version | sed 's/\"//g') -i prerelease --preid canary)+$(git rev-parse --short HEAD) --tag canary --no-release-draft --any-branch"
```
dylanjha, line 24: I have no idea if this will work 😆 has this been tested yet? Not seeing any canary release cannel on npmjs, so probably only one way to know for sure which is to merge and let it run.

I'm guessing you've run the components of these commands to verify the output?

```
$(npm pkg get version | sed 's/\"//g')
```

```
$(git rev-parse --short HEAD)
```

```
semver {version} --tag canary --no-release-draft --any-branch
```

.github/workflows/ci.yml
```diff
@@ -1,17 +1,21 @@
...
       - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v3
         with:
           node-version: ${{ matrix.node-version }}
       - run: yarn install --frozen-lockfile
       - run: yarn test
       - run: yarn build
+      - run: yarn publish-canary
+        if: ${{ github.ref == 'refs/heads/main' }}
```
dylanjha, line 19: instead of having this `if` gate, my first instinct was that we should break this into `ci.yml` and `cd.yml`, but after thinking about it, I actually think this is better. I'm on board with having it all in `ci.yml` for consistency 👍🏼

dylanjha (approved):
Approved from my side, but would like to have someone else like @gkatsev take a closer look

package.json
gkatsev: npm dist tags can be created on the fly.

luwes: yup all tested locally:
- npm pkg get version => "0.6.8"
- git rev-parse --short HEAD => c682c73
- semver is the node module NPM uses for versions, used to bump the version without editing package.json
- np is still used to publish without GH release notes

.github/workflows/ci.yml
luwes: we can still change when adding the manual NPM release work flow. I'm fine with either

luwes:
on hold until NPM_TOKEN is set

.github/workflows/ci.yml
```diff
@@ -1,17 +1,21 @@
...
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v3
         with:
           node-version: ${{ matrix.node-version }}
```
gkatsev, line 14: I think we need to add the `registry-url` here, like mentioned in the docs https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#publish-to-npmjs-and-gpr-with-npm

gkatsev (approved):

PR #247, opened 2022-06-12
Add media playlist example

heff:
[Updated media-playlist](https://github.com/muxinc/media-playlist/pull/1) based on [the playlist discussion](https://github.com/muxinc/media-chrome/discussions/97). Adding an example to media-chrome.

### DIFF
--- a/examples/media-elements/media-playlist.html
+++ b/examples/media-elements/media-playlist.html
@@ -28,7 +28,7 @@
             <track label="thumbnails" default kind="metadata" src="https://image.mux.com/DS00Spx1CV902MCtPj5WknGlR102V5HFkDe/storyboard.vtt" />
           </media-playlist-item>
           <media-playlist-item type="audio" src="https://dl.espressif.com/dl/audio/ff-16b-2c-44100hz.mp3"></media-playlist-item>
-          <media-playlist-item type="video" src="https://www.learningcontainer.com/wp-content/uploads/2020/05/sample-mp4-file.mp4"></media-playlist-item>
+          <media-playlist-item type="video" src="https://muxed.s3.amazonaws.com/ink.mp4"></media-playlist-item>
         </media-playlist>
         <media-control-bar>
           <media-play-button></media-play-button>
--- a/examples/media-elements/media-playlist.html
+++ b/examples/media-elements/media-playlist.html
@@ -1,0 +1,52 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta name="viewport" content="width=device-width" />
+    <title>Media Chrome Media Playlist Example</title>
+    <script type="module" src="https://unpkg.com/media-playlist@0.1"></script>
+    <script type="module" src="https://unpkg.com/hls-video-element@0"></script>
+    <script type="module" src="../../dist/index.js"></script>
+    <style>
+      /** add styles to prevent CLS (Cumulative Layout Shift) */
+      media-controller:not([audio]) {
+        display: block;         /* expands the container if preload=none */
+        max-width: 720px;       /* allows the container to shrink if small */
+        aspect-ratio: 16 / 9;   /* set container aspect ratio if preload=none */
+      }
+
+      .examples {
+        margin-top: 20px;
+      }
+    </style>
+  </head>
+  <body>
+    <main>
+      <h1>Media Chrome Media Playlist Example</h1>
+      <media-controller>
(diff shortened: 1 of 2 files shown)


cjpillsbury (approved):
Since this doesn't impact any fundamental architecture of media chrome, I don't see any reason not to merge this.

luwes:
bug? on my end the first clip that is played doesn't update media chrome state like playing and progress

SCR-20220613-d4t

heff:
> on my end the first clip that is played doesn't update media chrome state
Interesting. Can I see your test? It's working for me at least.
I did have issues with `<youtube-video>`, but I wasn't sure if that was a media-playlist issue or related to the youtube player load timing.

> ideally media-playlist-item's should also support <source> and <track> elements but could be supported later.

[They do!](https://github.com/muxinc/media-playlist/pull/1/files#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346R248)

luwes:
sweet! sorry, didn't check the code source.

Mac / Chrome 102
https://media-chrome-git-media-playlist-example-mux.vercel.app/examples/media-elements/media-playlist.html
https://stream.new/v/Y576wlRA3xRgjz63Lpr00isKn16OKdBWuHicD4fTrdFg

examples/media-elements/media-playlist.html
```diff
@@ -0,0 +1,52 @@
...
+      }
+    </style>
+  </head>
+  <body>
+    <main>
+      <h1>Media Chrome Media Playlist Example</h1>
+      <media-controller>
+        <media-playlist slot="media" loop>
+          <media-playlist-item type="hls-video" crossorigin src="https://stream.mux.com/DS00Spx1CV902MCtPj5WknGlR102V5HFkDe.m3u8">
+            <track label="thumbnails" default kind="metadata" src="https://image.mux.com/DS00Spx1CV902MCtPj5WknGlR102V5HFkDe/storyboard.vtt" />
```
luwes, line 28: this is really cool! `type="hls-video"` nice!

luwes:
I'm more leaning to keeping the media elements (source) out of the Media Chrome repo. Examples in, for sure.

[`media-playlist`](https://github.com/muxinc/media-playlist) is not really a video engine either, more like a source / media selector
[`responsive-source`](https://github.com/muxinc/responsive-source) would also fall in this category 

Since it's not UI my s1 is to not include it but I'm not 100% sure what the extras can include.

--- 

If it's for reducing friction w/ releasing etc, there's still some things we could do for the standalone pkgs like auto releasing on PR merges etc.

heff:
> If we add those, do they go in that project? Or here?

Here (media-chrome), because media-playlist won't be the only media element with playlist functionality. e.g. [Youtube](https://developers.google.com/youtube/iframe_api_reference) with `player.nextVideo()`. That makes it a [media-ui-extensions](https://github.com/video-dev/media-ui-extensions/issues) detail to solve, but otherwise we've been pretty clear and consistent with the line between "chrome" and "media". While I feel like this is a _special_ or _supportive_ media element, it also feels like a slippery slope introducing media elements into media chrome. But the discoverability of the whole ecosystem, including other media elements, is something that's important to solve well.

heff:
> Must be a custom element loading issue. I'll dig into when I get a minute.

Confirmed, the load order of media-playlist vs media-chrome scripts matters for some reason. Not clear why yet. If media playlist is loaded _before_ media chrome, the problem happens. That feels backwards from what I'd expect.

luwes:
could it be before this line `document.createElement(playlistItem.getAttribute('type'));` if it's a custom element we need to wait until it's defined? `customElements.whenDefined()` 

or possibly upgrade it right after so the attributeCallback will be triggered when setting the attributes  
https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry/upgrade

also make sure you attach the events before setting the attributes, though less important, `loadstart` will be delayed by a tick when setting the src.

luwes:
I confirmed this is caused by https://github.com/muxinc/hls-video-element
if you add in my fork `<script type="module" src="https://unpkg.com/@luwes/hls-video-element@0"></script>` it works

maybe temporarily use just plain video and audio tags to get this example in?

(https://github.com/muxinc/media-playlist has a small bug where `media.muted` doesn't seem to work)

luwes (approved):
LGTM, small hls-video fix needed but don't want to block this from merging

PR #248, opened 2022-06-13
fix: hide gesture-layer when in audio mode

gkatsev:
The gestures layer, depending on configuration, may add an extra 5
pixels above or below the control bar, which is unwanted. Instead, when
we're playing back audio, we should hide the gestures display, which
involves making it a block element so that we can set the height of it
to zero.

Setting the MC's background color to red, we can see that there's some extra pixels above or on the side.
Screen Shot 2022-06-13 at 17 41 19
Screen Shot 2022-06-13 at 17 41 08
One thought I had been to set the layers's display to be `inline-block`, but that causes some extra sizing issues:
Screen Shot 2022-06-13 at 17 41 45

So, decided to go with a more surgical approach and only target the issue at hand, which works like a charm:
Screen Shot 2022-06-13 at 17 41 27

### DIFF
--- a/src/js/media-container.js
+++ b/src/js/media-container.js
@@ -44,6 +44,15 @@ template.innerHTML = `
     }
 
     /*
+     * when in audio mode, hide the gesture-layer which causes media-controller to be taller than the control bar
+     *
+     */
+    :host([audio]) [part~=layer][part~=gesture-layer] {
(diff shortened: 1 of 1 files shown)


cjpillsbury (approved):
🏄‍♂️

PR #250, opened 2022-06-15
chore: add publish workflow

luwes:
this change adds a CD workflow to publish a release via GH actions. can choose between conventional, patch, minor, major.

already working in https://github.com/muxinc/castable-video

this is also more in line with Mux elements where we started using [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)

- [x] first run `conventional-changelog -p angular -i CHANGELOG.md -s -r 0` to generate a CHANGELOG.md from scratch
  see https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-cli

### DIFF
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -1,0 +1,36 @@
+#!/usr/bin/env bash
+
+# npm publish with goodies
+# inspired by https://gist.github.com/stevemao/280ef22ee861323993a0
+#
+# prerequisites:
+# `npm install -D conventional-recommended-bump \
+#   conventional-changelog-cli conventional-github-releaser`
+# `release` with optional argument `patch`/`minor`/`major`/`<version>`
+# defaults to conventional-recommended-bump
+
+# `np` is not suitable for CI/CD
+# https://github.com/sindresorhus/np/issues/619#issuecomment-994493179
+
+release() {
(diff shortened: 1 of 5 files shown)


scripts/publish.sh
```diff
@@ -0,0 +1,36 @@
...
+
+# `np` is not suitable for CI/CD
+# https://github.com/sindresorhus/np/issues/619#issuecomment-994493179
+
+release() {
+  BUMP=$(conventional-recommended-bump -p angular)
+  VERSION=$(npm --no-git-tag-version version ${1:-$BUMP})
+  conventional-changelog -p angular -i CHANGELOG.md -s
+  git add CHANGELOG.md
+  git commit -m "docs(CHANGELOG): $VERSION"
```
gkatsev, line 20: if we move this into the `version` script, we can avoid the extra commit, it'll get committed as part of the version.
Example https://github.com/videojs/video.js/blob/main/package.json#L78
https://github.com/videojs/video.js/commit/fb43a69e707be22a57e3048738afd03202258960

luwes: I did experiment with that hook but there's 2 times the script calls `npm version` so I thought best to do it manually

PR #255, opened 2022-06-27
feat: improve time range behavior

luwes:
- [x] fixes hiding bug when hovering on preview thumbnail 
- [x] fixes time range progress not center of drag thumb bug
- [x] makes the preview thumbnail not go out of the player bounds
- [x] refactor `<media-thumbnail-preview>` to be able to use `max-width`, `max-height`, `min-width`, `min-height` on it
- [x] add `<media-time-range>` **preview** and **current** boxes/slots that move on pointer hover and on playhead / current time 
	this allows users to add anything in the slots that they like to have moved with the boxes along side the time range.
- [x] adds a [`<media-theme-youtube>`](https://media-chrome-git-fork-luwes-improve-time-range-mux.vercel.app/examples/themes/youtube-theme.html)
- [x] add [`<media-preview-thumbnail>` example](https://media-chrome-git-fork-luwes-improve-time-range-mux.vercel.app/examples/control-elements/media-preview-thumbnail.html) 
- [x] add `<div>` range hover element that expands the hover area vertically to help on thin time ranges
- [x] add preview time code
- [x] add preview highlight on the time range
- [x] time range steps are too coarse (minimum is step value of 1?)

---

New media-time-range nesting options with `preview` and `current` slot:

```html
  <media-time-range>
    <media-preview-thumbnail slot="preview"></media-preview-thumbnail>
    <media-preview-time-display slot="preview"></media-preview-time-display>
    <media-current-time-display slot="current"></media-current-time-display>
  </media-time-range>
```

--- 

TODO: update docs with all the new CSS vars

PR #258, opened 2022-06-29
test: add eslint

luwes:
just adds eslint so it can be merged separate from https://github.com/muxinc/media-chrome/pull/255

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,10 +49,28 @@
     "conventional-github-releaser": "^3.1.5",
     "conventional-recommended-bump": "^6.1.0",
     "esbuild": "^0.14.23",
+    "eslint": "^8.18.0",
     "prettier": "^2.5.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -49,10 +49,28 @@
...
     "sinon": "^11.1.2"
+  },
+  "eslintConfig": {
+    "globals": {
+      "globalThis": "writable"
+    },
+    "env": {
+      "browser": true,
+      "es6": true,
+      "mocha": true
```
cjpillsbury, line 66: nit(non-blocking): is this right? we don't use mocha.

cjpillsbury (approved):

package.json
luwes: good catch, test-runner uses Mocha under the hood actually. should be fine

PR #263, opened 2022-07-11
feat: initial keyboard shortcut support

gkatsev:
Kicking off the work for https://github.com/muxinc/media-chrome/discussions/257

### DIFF
--- a/docs/media-controller.md
+++ b/docs/media-controller.md
@@ -308,10 +308,10 @@ The following controls are supported:
 | ⬅       | Seek back 10s     |
 | ➡       | Seek forward 10s  |
 
-If you are implementing an interactive element that uses any of these keys, you can stopPropagation in your `keyup` handler. Alternatively, you can add a `keysUsed` property on the element or a `keysused` attribute.
+If you are implementing an interactive element that uses any of these keys, you can stopPropagation in your `keyup` handler. Alternatively, you can add a `keysUsed` property on the element or a `keysused` attribute. The values are those that match the `key` property on the KeyboardEvent. You can find a list of those values [on mdn](https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values). Additionally, since the DOM list can't have the Space key represented as `" "`, we will accept `Space` as an alternative name for it.
 Example (`keysused` attribute):
 ```html
-   <media-time-range keysused="ArrowLeft ArrowRight"></media-time-range>
+   <media-time-range keysused="ArrowLeft ArrowRight Space"></media-time-range>
 ```
 
 Example (`keysUsed` property):
--- a/docs/media-controller.md
+++ b/docs/media-controller.md
@@ -82,6 +82,28 @@ Example (disabling gestures via `gestures-disabled`):
 </media-controller>
 ```
 
+- `nohotkeys` - Use this to turn off keyboard shortcuts.
+
+Example (hotkeys disabled):
+
+```html
+<media-controller nohotkeys>
+  <video
+    slot="media"
+    src="https://stream.mux.com/DS00Spx1CV902MCtPj5WknGlR102V5HFkDe/high.mp4"
+  ></video>
+  <media-control-bar>
+    <media-seek-backward-button></media-seek-backward-button>
+    <media-play-button></media-play-button>
+    <media-seek-forward-button></media-seek-forward-button>
+    <media-mute-button></media-mute-button>
+    <media-volume-range></media-volume-range>
+    <media-time-range></media-time-range>
+    <media-time-display></media-time-display>
+  </media-control-bar>
+</media-controller>
+```
+
 # Styling
 
 - `aspect-ratio` - While this is [a standard CSS style](https://css-tricks.com/almanac/properties/a/aspect-ratio/), it's fairly new, and you're likely to want to use it frequently on `<media-controller/>`, at least for `video` use cases. Most often, you'll want the `aspect-ratio` to match your video content's aspect ratio.
@@ -272,3 +294,31 @@ As we work through other common use cases, both internally and with the communit
 - Example use cases:
   - [Basic example](https://media-chrome.mux.dev/examples/basic.html) ([view source](https://github.com/muxinc/media-chrome/blob/main/examples/basic.html))
   - [Spotify theme](https://media-chrome.mux.dev/examples/themes/spotify-theme.html) ([view source](https://github.com/muxinc/media-chrome/blob/main/examples/themes/spotify-theme.html))
+
+## Keyboard Shortcuts
+
+By default, Media Controller has keyboard shortcuts that will trigger behavior when specific keys are pressed when the focus is inside the Media Controller.
+The following controls are supported:
+| Key     | Behavior          |
+|---------|-------------------|
+| Space   | Toggle Playback   |
+| `k`     | Toggle Playback   |
+| `m`     | Toggle mute       |
+| `f`     | Toggle fullscreen |
+| ⬅       | Seek back 10s     |
+| ➡       | Seek forward 10s  |
+
+If you are implementing an interactive element that uses any of these keys, you can stopPropagation in your `keyup` handler. Alternatively, you can add a `keysUsed` property on the element or a `keysused` attribute.
+Example (`keysused` attribute):
+```html
+   <media-time-range keysused="ArrowLeft ArrowRight"></media-time-range>
+```
+
+Example (`keysUsed` property):
+```js
+class MyInteractiveElement extends window.HTMLElement {
+  get keysUsed() {
+    return ['Enter', ' '];
+  }
+}
+```
--- a/src/js/media-container.js
+++ b/src/js/media-container.js
@@ -17,8 +17,6 @@ import { nouns } from './labels/labels.js';
 import { containsComposedNode } from './utils/element-utils.js';
 // Guarantee that `<media-gesture-receiver/>` is available for use in the template
 import './media-gesture-receiver.js';
-import MediaChromeRange from './media-chrome-range.js';
-import MediaChromeButton from './media-chrome-button.js';
 
 const template = document.createElement('template');
 
@@ -350,8 +348,6 @@ class MediaContainer extends window.HTMLElement {
     // Unhide for keyboard controlling
     this.addEventListener('keyup', (e) => {
       scheduleInactive();
-
-      this.handleKeyupEvent(e);
     });
 
     // when we get a tap, we want to unhide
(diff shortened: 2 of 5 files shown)


src/js/media-container.js
```diff
@@ -392,6 +396,90 @@ class MediaContainer extends window.HTMLElement {
...
+        return;
+      }
+
+      let eventName, currentTimeStr, currentTime, detail, evt;
+      const seekOffset = 10;
+
+      switch (e.key) {
+        case ' ':
+        case 'k':
+          eventName =
```
gkatsev, line 414: all of these were copied from the corresponding buttons. It would be nice if we can share the code, but I'm not positive there's a great way to do it. We could do something like `PlayButton.prototype.handleClick.call(this, e)` or something. Thoughts?

src/js/media-container.js
```diff
@@ -348,6 +350,8 @@ class MediaContainer extends window.HTMLElement {
     // Unhide for keyboard controlling
     this.addEventListener('keyup', (e) => {
       scheduleInactive();
+
+      this.handleKeyupEvent(e);
```
cjpillsbury, line 354: suggestion(blocker): We should use the keydown + keyup strategy here or similar to filter out false positives while meta keys are pressed https://github.com/muxinc/media-chrome/blob/main/src/js/media-chrome-button.js#L108-L126 . If/when we add custom mappings, we'd still likely want "exact matches" even for multi-key support

cjpillsbury, line 354: suggestion(blocker): Before merging/releasing, let's make sure there's some way for implementors to disable this functionality. Can be a simple attribute like `disableshortcuts` (or whatever name is appropriate). Should also be an `observedAttribute` to ensure that the relevant event handler(s) are added/removed.

cjpillsbury (changes requested):
a couple of asks, but this looks close to done for a good mvp spike.

src/js/media-container.js
cjpillsbury: Glad you're thinking of that. For now I'd lean towards "duplicate code" for this PR and KISS functional-like approaches (some module that "owns" a bucket of functions that take an element and return an appropriate `CustomEvent` (imo the "`dispatchEvent`" side effect should be externalized, but I could be convinced otherwise)

src/js/media-container.js
```diff
@@ -392,6 +396,90 @@ class MediaContainer extends window.HTMLElement {
...
+    if (this.contains(e.target) || this.shadowRoot.contains(e.target)) {
+      // ignore space bar and enter
+      if (e.target instanceof MediaChromeButton && [' ', 'Enter'].includes(e.key)) {
+        return;
+      } else if (e.target instanceof MediaChromeRange && ['ArrowLeft', 'ArrowRight', 'ArrowUp', 'ArrowDown'].includes(e.key)) {
+        return;
+      }
+
+      let eventName, currentTimeStr, currentTime, detail, evt;
+      const seekOffset = 10;
```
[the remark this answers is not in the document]
gkatsev: yeah, this makes sense. I think also worth calling it out in https://github.com/muxinc/media-chrome/discussions/257.
This actually brings up an interesting point in that the code is currently in `media-container` and not `media-controller` and an attribute and default value should probably live in `media-controller`?

I guess one way of handling that is that we call the `this.handleKeyupEvent` or whatever method on `this` but only implement it in `media-controller` with an no-op method in `media-container`.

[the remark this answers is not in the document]
cjpillsbury: yeah that's a good point. These interactions are more about "state management"/"state change requests," so they probably belong in `media-controller`. I also think it would be fine to add a second event handler there for keyup, fwiw (which may make more sense anyway if you do key down + key up pairs similar to the components).

src/js/media-container.js
```diff
@@ -17,6 +17,8 @@ import { nouns } from './labels/labels.js';
 import { containsComposedNode } from './utils/element-utils.js';
 // Guarantee that `<media-gesture-receiver/>` is available for use in the template
 import './media-gesture-receiver.js';
+import MediaChromeRange from './media-chrome-range.js';
```
heff, line 20: This makes me cringe a little. Partly for the added dependency complexity, but partly because it means we've made media chrome's button and range more special somehow in this ecosystem. I'm not gonna call this a blocker but want to encourage us to consider an approach that removes this dependency, maybe as a fast follow.

gkatsev: Yeah, this is mostly for to get it working. I think the proper solution is what we discussed in https://github.com/muxinc/media-chrome/discussions/257, where the elements can expose a property like `keysUsed` and then below, we can do
```js
// skip handler is already used by the target
if (evt.target.keysUsed ?? []).includes(evt.key)) {
  return;
}
```
And we won't need to import these directly.

The alternative is that each component that handles keyup will call `stopPropagation` properly.

Originally, I was thinking we can punt on this, but this affects a component that's currently in Mux Player, so, we probably should come to a decision sooner rather than earlier. I lean towards a properly like `keysUsed` (or another name).

src/js/media-container.js
gkatsev: There doesn't seem to be a very good precedent for this, but I would lean towards `nokeyboardshortcuts` or `noshortcuts`, potentially with `globalkeyboardhosrtcuts` or `globalshortcuts` for later attaching to `window`.

src/js/media-container.js
heff: Ok cool, makes sense. 👍 

Without understanding all the complexity, I still feel better about something where control-specific complexity is kept in the controls, or the super classes of controls, rather than making media container/controller smarter here so controls don't have to be. Probably worth a call to discuss.

src/js/media-controller.js
```diff
@@ -477,6 +487,28 @@ class MediaController extends MediaContainer {
...
+
+    // copied from media-chrome-button
+    const keyUpHandler = (e) => {
+      const { key } = e;
+      if (!ButtonPressedKeys.includes(key)) {
+        this.removeEventListener('keyup', keyUpHandler);
+        return;
+      }
+
+      if (!this.hasAttribute('noshortcuts')) {
```
gkatsev, line 499: should this go into `constants.js` somewhere? MediaUIAttributes doesn't seem correct.
Also, should it be `media-noshortcuts`?

src/js/media-controller.js
```diff
@@ -45,6 +48,13 @@ const {
   MEDIA_PLAYBACK_RATE_REQUEST,
 } = MediaUIEvents;
 
+const ButtonPressedKeys = ['Enter', ' ', 'f', 'm', 'k', 'ArrowLeft', 'ArrowRight'];
+const DEFAULT_SEEK_OFFSET = 10;
+
+/**
+ * Media Controller should not mimic the HTMLMediaElement API.
+ * @see https://github.com/muxinc/media-chrome/pull/182#issuecomment-1067370339
+ */
```
gkatsev, line 57: moved from the bottom of the class here, which seems more appropriate.

src/js/media-controller.js
```diff
@@ -17,6 +17,9 @@ import { fullscreenApi } from './utils/fullscreenApi.js';
 import { constToCamel } from './utils/stringUtils.js';
 import { containsComposedNode } from './utils/element-utils.js';
 
+import MediaChromeRange from './media-chrome-range.js';
+import MediaChromeButton from './media-chrome-button.js';
```
gkatsev, line 21: The `keysUsed` vs `stopPropagation` should be resolved before this PR can be merged.

src/js/media-container.js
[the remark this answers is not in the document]
gkatsev: added a note here https://github.com/muxinc/media-chrome/discussions/257#discussioncomment-3132937

src/js/media-container.js
gkatsev: https://github.com/muxinc/media-chrome/pull/263#discussion_r919492637

src/js/media-container.js
gkatsev: Added a new comment so that this doesn't get lost here https://github.com/muxinc/media-chrome/pull/263#discussion_r919493069

src/js/media-controller.js
gkatsev: I created two PRs with examples of `keysUsed` and `stopPropagation`:
- keysUsed https://github.com/gkatsev/media-chrome/pull/1
- stopPropagation https://github.com/gkatsev/media-chrome/pull/2

To me, keysUsed is actually a bit less presumptuous since it requires less additional code from components, and it doesn't prevent users from still being able to listen to the key events if they should want.
In comparison, by default, with the stopPropagation thing, we either need to have special handling for native elements like buttons and inputs with hardcoded keys that are used, or our wrapper needs to specifically call stopPropagation, otherwise, both the native behavior and the keyboard shortcut would end up triggering. And now, we're also limiting the ability of users to be able to listen to keyup with these as targets.

gkatsev: follow-up from chat:
- write some docs on how this would be used by folks
- should this be an attribute instead of a property? Or should it be both?
- better name than keysUsed?
- keysUsed doesn't preclude from media-chrome users using stopPropagation in their keyboard handling to prevent multiple behaviors from occurring.

src/js/media-controller.js
luwes: I like the compactness of `noshortcuts`, but not super clear to me right off the bat. why not `nokeyboard`?

src/js/media-controller.js
luwes: I'm leaning more towards stopPropagation, seems this is exactly meant for this and no extra API is added. less coupling. the little extra code in the components is worth that I think.

src/js/media-controller.js
gkatsev: Alternative name: `nohotkeys`

src/js/media-controller.js
gkatsev: Agreed on both property `keysUsed` and attribute `keyused`.

src/js/media-controller.js
```diff
@@ -710,10 +739,95 @@ class MediaController extends MediaContainer {
     els.splice(index, 1);
   }
 
-  /**
-   * Media Controller should not mimic the HTMLMediaElement API.
-   * @see https://github.com/muxinc/media-chrome/pull/182#issuecomment-1067370339
-   */
+  keyboardShortcutHandler(e) {
+    if (this.contains(e.target) || this.shadowRoot.contains(e.target)) {
```
cjpillsbury, line 743: question(non-blocking): Do we need these checks?

src/js/media-controller.js
```diff
@@ -710,10 +739,95 @@ class MediaController extends MediaContainer {
...
+      // if the event's key is already handled by the target, skip keyboard shortcuts
+      // keysUsed is either an attribute or a property.
+      // The attribute is a DOM array and the property is a JS array
+      // In the attribute Space represents the space key and gets convered to ' '
+      const keysUsed = (e.target.getAttribute('keysused')?.split(' ') ?? e.target?.keysUsed ?? [])
```
cjpillsbury, line 748: thought(non-blocking): As a followup, we'll need to figure out how this should work for non-nested components based on the current architecture (See code/architecture for `associatedElements` and how event (re)propagation works).

src/js/media-controller.js
```diff
@@ -477,6 +484,28 @@ class MediaController extends MediaContainer {
...
+
+    // copied from media-chrome-button
+    const keyUpHandler = (e) => {
+      const { key } = e;
+      if (!ButtonPressedKeys.includes(key)) {
+        this.removeEventListener('keyup', keyUpHandler);
+        return;
+      }
+
+      if (!this.hasAttribute('nohotkeys')) {
```
cjpillsbury, line 496: suggestion(non-blocking): Instead of continuing to listen to irrelevant keyboard presses when `"nohotkeys"` is set, we should probably:

1. add `"nohotkeys"` to `observedAttributes`.
2. add/remove event handlers for `"keyup"`/`"keydown"` based on `attributeChangedCallback`
    1. This entails moving the event callbacks to methods instead of inline.

docs/media-controller.md
```diff
@@ -272,3 +294,31 @@ As we work through other common use cases, both internally and with the communit
...
+| Key     | Behavior          |
+|---------|-------------------|
+| Space   | Toggle Playback   |
+| `k`     | Toggle Playback   |
+| `m`     | Toggle mute       |
+| `f`     | Toggle fullscreen |
+| ⬅       | Seek back 10s     |
+| ➡       | Seek forward 10s  |
+
+If you are implementing an interactive element that uses any of these keys, you can stopPropagation in your `keyup` handler. Alternatively, you can add a `keysUsed` property on the element or a `keysused` attribute.
```
cjpillsbury, line 311: chore(blocking):

1. Discuss and link to https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values
2. Callout the `"Space"` keyword mapping

cjpillsbury (changes requested):
One docs chore and a handful of non-blocking callouts. Otherwise, 👌. Did basic smoke testing across available browsers.

src/js/media-controller.js
gkatsev: Maybe not, since I guess the event bubbling should guarantee that it's coming from a child element?

src/js/media-controller.js
gkatsev: yup, this got called out in the discussion #257

src/js/media-controller.js
gkatsev: updated

docs/media-controller.md
gkatsev: updated

docs/media-controller.md
```diff
@@ -272,3 +294,31 @@ As we work through other common use cases, both internally and with the communit
...
+| `k`     | Toggle Playback   |
+| `m`     | Toggle mute       |
+| `f`     | Toggle fullscreen |
+| ⬅       | Seek back 10s     |
+| ➡       | Seek forward 10s  |
+
+If you are implementing an interactive element that uses any of these keys, you can stopPropagation in your `keyup` handler. Alternatively, you can add a `keysUsed` property on the element or a `keysused` attribute. The values are those that match the `key` property on the KeyboardEvent. You can find a list of those values [on mdn](https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values). Additionally, since the DOM list can't have the Space key represented as `" "`, we will accept `Space` as an alternative name for it.
+Example (`keysused` attribute):
+```html
+   <media-time-range keysused="ArrowLeft ArrowRight Space"></media-time-range>
```
cjpillsbury, line 314: 😎

src/js/media-controller.js
cjpillsbury: yup that's what I was thinking. We may also get into some weirdness with `contains` checks for nested web component use cases.

src/js/media-controller.js
```diff
@@ -710,10 +735,123 @@ class MediaController extends MediaContainer {
...
   }
 
-  /**
-   * Media Controller should not mimic the HTMLMediaElement API.
-   * @see https://github.com/muxinc/media-chrome/pull/182#issuecomment-1067370339
-   */
+  #keyUpHandler(e) {
+    const { key } = e;
+    if (!ButtonPressedKeys.includes(key)) {
+      this.removeEventListener('keyup', keyUpHandler);
```
cjpillsbury, line 741: issue(blocker): missed `this.#keyUpHandler` refactor.

cjpillsbury (approved):
pre-approving with the minor callout and recommend to remove the contains et al. check.

docs/media-controller.md
```diff
@@ -82,6 +82,28 @@ Example (disabling gestures via `gestures-disabled`):
 </media-controller>
 ```
 
+- `nohotkeys` - Use this to turn off keyboard shortcuts.
+
+Example (hotkeys disabled):
+
+```html
+<media-controller nohotkeys>
```
luwes, line 90: 👍

luwes (approved):
LGTM! 🎉

PR #270, opened 2022-07-16
docs: add Chromecast / Google Cast example

luwes:
Adds an Google casting example to the examples site 
Requires https://github.com/muxinc/castable-video

I chose to add BBB here because somehow Chromecast didn't like the Mad Max progressive mp4 🤷‍♂️ (stuttering)


cjpillsbury (approved):
👍

PR #271, opened 2022-07-18
fix: timerange progress jumpy w/out playback rate

luwes:
just noticed this issue where the smooth progress becomes jumpy with a playback rate other than 1.

### DIFF
--- a/src/js/media-time-range.js
+++ b/src/js/media-time-range.js
@@ -130,6 +130,7 @@ class MediaTimeRange extends MediaChromeRange {
       MediaUIAttributes.MEDIA_PREVIEW_IMAGE,
       MediaUIAttributes.MEDIA_PREVIEW_TIME,
       MediaUIAttributes.MEDIA_BUFFERED,
+      MediaUIAttributes.MEDIA_PLAYBACK_RATE,
     ];
   }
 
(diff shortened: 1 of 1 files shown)


src/js/media-time-range.js
```diff
@@ -235,6 +236,11 @@ class MediaTimeRange extends MediaChromeRange {
     return attrVal != null ? +attrVal : undefined;
   }
 
+  get mediaPlaybackRate() {
+    const attrVal = this.getAttribute(MediaUIAttributes.MEDIA_PLAYBACK_RATE);
```
cjpillsbury, line 240: nit(non-blocking): We can use "new-fangled" js (should probably start using this everywhere):
```js
return this.getAttribute(MediaUIAttributes.MEDIA_PLAYBACK_RATE) ?? 1;
````

File: src/js/media-time-range.js [hunk 2]
```diff
@@ -161,7 +162,7 @@ class MediaTimeRange extends MediaChromeRange {
 
     this._refreshBar = () => {
       const delta = (performance.now() - this._updateTimestamp) / 1000;
-      this.range.value = this.mediaCurrentTime + delta;
+      this.range.value = this.mediaCurrentTime + delta * this.mediaPlaybackRate;
```
cjpillsbury, line 165: nit(non-blocking): Consider adding parens for clarity of order of operations (even though this is mathematically correct)

cjpillsbury (approved):
a couple of non-blocking nits. Otherwise, 👌

PR #273, opened 2022-07-19
feat: icons refresh + sizing by height

### LINKED ISSUE #86: CC button default height isn't consistent with other buttons
Follow up on #74 

image

https://media-chrome-err0093s8-mux.vercel.app/examples/standalone-controls.html

### LINKED ISSUE #89: Height of buttons with text content isn't consistent with icons
After #74 the height of buttons that use text isn't consistent with the height of buttons that use an icon.

Pasted_Image_10_29_21__5_00_PM

Before: https://media-chrome-djevl3pei-mux.vercel.app/examples/control-elements/media-seek-backward-button.html
After: https://media-chrome-err0093s8-mux.vercel.app/examples/control-elements/media-seek-backward-button.html

### LINKED ISSUE #144: Update Icons
See #139 for reference icons and goals.

- Icons should still respect all anticipated styling from Media Chrome (from css vars and well-defined component styles)
- Examples should be updated as needed based on changes to icon dimensions.

luwes:
Closes #146, fixes #144, fixes #86, fixes #89 and builds on this PR to refresh the SVG icons.

This fixes an issue which is pretty noticeable once you see the difference from before & after.

Currently icons are set by width but this can easily change depending the icon e.g. PiP is a rectangle, Fullscreen is a square.
Having a fixed width will result in the height being changed of the icon and the the icon will not be shown as intended.

(Also for the designer it's easier to design icons with a fixed height than a fixed width I think, also the extra horizontal space for achieving a fixed width icon slice would look strange)

- Updates the SVG files and cleaned up SVG markup. `xmlns="http://www.w3.org/2000/svg"` is not required for inline SVG's.
- Updated the themes to look okay with the new SVG's.

thanks @tomByrer, he noticed this here as well: https://github.com/muxinc/media-chrome/pull/146#issuecomment-1086900783

This is also an issue in mux-player but less noticeable because the icons are of more similar width.

### before

SCR-20220718-rxz

### ~after~

SCR-20220718-rxq

### after last iteration with 24px height + padding

SCR-20220803-dcv


cjpillsbury:
> Does this make sense? happy to discuss

@heff and I have discussed this switch (using height as the primary dimension instead of width) for awhile, so I'm completely on board (and so is Heff, iirc?). This will plausibly have a notable impact on pre-existing use cases, so we may want to consider doing a minor version bump for the release that includes this.

heff:
image

image

In general this looks good. 👍 

The icons are feeling a lot more crowded in the control bar after the change, especially the play and full screen buttons against the left and right side. I'd suggest reducing the default size of the icons to closer to where they were. Hopefully we can avoid adding left/right padding to the control bar, but it might come to that.

luwes:
that's a good point, yes pretty hard to see a solution that wouldn't require a left/right padding to the controlbar.

heff:
The previous version didn't have control bar padding, but it did have more intentional spacing which helped icons not feel cramped. What can we do to bring that back?

image

The 43px might actually be including some of the video letter box. But the main detail is the balance between the horizontal and vertical spacing. I'm not against icon-specific padding settings, assuming it's easily overridable.

cjpillsbury:
> The previous version didn't have control bar padding, but it did have more intentional spacing which helped icons not feel cramped. What can we do to bring that back?

@heff @luwes since this was a resuscitatation of a branch I initiated, the icons we're using in Mux Player did get updated after a similar discussion that included Jason. We may already have assets that have been updated with the whitespace, either because the ones used by Mux Player are identical, or I may have them available elsewhere. Let's coordinate offline.

luwes:
I think we'll have to figure out if it's better devex to add padding programmatically to the icons or should the icons come with padding. maybe we can look at the most popular icon sets and see if they are delivered with padding or almost touching the edges of the icon paths.

luwes:
most popular open source icons come with some padding, 24x24 is most popular format.
only fontawesome didn't have padding added to its icons.

what do you think of taking the average of these icon sets and make that look good in Media Chrome?
icons like the 1st, 3rd, 4th, ... the 7 last icons in the image below counting from the top.

SCR-20220722-jmg

luwes:
Updated with padding height 24px
SCR-20220726-g23

tomByrer:
Have the changes been manually tested on touchscreens, specifically smaller phones (cheap/older/emerging-markets)?
That is the scenario where a careful balance between legibility & usability.

dylanjha:
Here's a URL for testing on various devices:

https://media-chrome-hzll703jz-mux.vercel.app/examples/basic.html

luwes:
last iteration with 24px height + padding

SCR-20220803-dcv

luwes:
before:

SCR-20220803-g42

after:

SCR-20220803-g3s

luwes:
new in this PR (not documented yet):
`--media-control-height`: change the height of all the control elements at once
`--media-control-padding`: change the padding of all the controls in one go

docs/styling.md
```diff
@@ -23,8 +23,8 @@ Our current styling architecture is still quite nascent and is very likely to un
 | ---------------------------------- | ------------ | --------------------- | ------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `--media-control-background`       | `background` | `rgba(20,20,30, 0.7)` | background color of the component                                                                 | Applies to other components as well ([See notes below \*\*](#notes))                                                                                                 |
 | `--media-control-hover-background` | `background` | `rgba(50,50,70, 0.7)` | background color of the button when hovered                                                       | Applied by the `:hover` [pseudo-class](https://developer.mozilla.org/en-US/docs/Web/CSS/:hover) Applies to other components as well ([See notes below \*\*](#notes)) |
```
gkatsev, line 25: should add `--media-control-padding` and `--media-control-height` to the docs here

gkatsev (approved):
Yeah, this looks much nicer!
image
vs
image

Approving, but would be good to get the docs updated with the new CSS vars.

Also, could this potentially be a breaking change for folks? Wonder if when we version, we should do it as a "major" 0.x bump rather than the "patch" 0.x.y bump.

cjpillsbury (approved):
🎉

PR #282, opened 2022-08-08
Feat/list components

cjpillsbury:
Will add callouts for discussion here/in inline comments. tl;dr - this is a draft to discuss the broad architecture for a list component that is generic and also conforms to WAI recommendations/implementation details (aka "a11y"). Implemented the captions list to see how a component would "use" this architecture, also for discussion.


src/js/media-captions-list.js
```diff
@@ -0,0 +1,119 @@
...
+
+  attributeChangedCallback(attrName, oldValue, newValue) {
+    if (
+      [
+        MediaUIAttributes.MEDIA_SUBTITLES_LIST,
+        MediaUIAttributes.MEDIA_CAPTIONS_LIST,
+      ].includes(attrName) &&
+      oldValue !== newValue
+    ) {
+      this.data = [
```
cjpillsbury, line 62: given that this data is treated separately in the media controller state management architecture, we need some way to "normalize" to a single list (on the assumption we want our list component to be generic).

src/js/media-chrome-listbox.js
```diff
@@ -0,0 +1,416 @@
...
+  );
+};
+const identity = (x) => x;
+
+const getClassName = (x) => {
+  return x?.constructor?.name ?? 'UNKNOWN';
+};
+
+const DEFAULT_FORMAT = identity;
+const DEFAULT_ITEM_RENDERER = 'li';
```
cjpillsbury, line 69: currently defining an "itemRenderer" as an element (which could be a custom element) that is expected to have innerText + attributes set on it (See, e.g., `render()` below)

src/js/media-chrome-listbox.js
```diff
@@ -0,0 +1,416 @@
...
+
+  render() {
+    const { selectedValue } = this;
+    const listItemEls = Array.from(this.data).map((value) => {
+      const listItemEl = document.createElement(this.itemRenderer);
+      listItemEl.setAttribute('role', 'option');
+      listItemEl.id = value;
+      listItemEl.innerText = this.format(value);
+      if (value === selectedValue) {
+        this.selectedElement = listItemEl;
```
cjpillsbury, line 279: this is currently here to handle cases where data is updated/items are re-rendered after an item has already been selected.

src/js/media-chrome-listbox.js
```diff
@@ -0,0 +1,416 @@
...
+  handleItemClick(evt) {
+    if (evt.target.getAttribute('role') !== 'option') return;
+    this.selectedElement = evt.target;
+    this.activeElement = evt.target;
+  }
+
+  /**
+   * Check if the selected option is in view, and scroll if not
+   */
+  updateScroll() {
```
cjpillsbury, line 316: this is to automatically scroll so the activeElement if the list items don't fit.

src/js/media-chrome-listbox.js
```diff
@@ -0,0 +1,416 @@
...
+        this.activeIndex = length - 1;
+        break;
+      case ' ':
+        evt.preventDefault();
+        this.selectedElement = this.activeElement;
+        break;
+      // fall through
+      default:
+        var itemToFocus = this.findItemToFocus(key);
+        if (itemToFocus) {
```
cjpillsbury, line 367: this is to support search via typed string, recommended by WAI for lists.

src/js/media-chrome-listbox.js
```diff
@@ -0,0 +1,416 @@
...
+    return this.shadowRoot?.getElementById('listbox');
+  }
+
+  get listItemElements() {
+    return (
+      Array.from(this.listboxElement.querySelectorAll('[role="option"]')) ?? []
+    );
+  }
+
+  get activeDescendant() {
```
cjpillsbury, line 186: `activeDescendant` is the "source of truth" for all related "active" properties, since it bottoms out in the ARIA requirements for identifying active ("focused") items via `aria-activedescendant`, defined on the listbox element itself.

src/js/media-chrome-listbox.js
```diff
@@ -0,0 +1,416 @@
...
+        )} activeIndex property. Must be within the index range of data (0-${
+          listItemElements.length - 1
+        })`
+      );
+      return;
+    }
+    this.activeElement = listItemElements[value];
+  }
+
+  get selectedElement() {
```
cjpillsbury, line 252: for the implementation of "selection", there are two related pieces:
- `aria-selected` is used for A11Y and identifies the "source of truth" for what is selected
- this works because (currently), the `id` is currently used for identifying the `value` of an element, so e.g. the `selectedValue` would be `this.selectedElement.id`

src/js/media-chrome-listbox.js
```diff
@@ -0,0 +1,416 @@
...
+  <style>
+    [role="listbox"] {
+        margin: 1em 0 0;
+        padding: 0;
+        min-height: 18em;
+        border: 1px solid #aaa;
+        background: white;
+    }
+    
+    [role="listbox"]#listbox {
```
luwes, line 18: `[role="listbox"]` shouldn't be needed here.
best to use one of them, all `#listbox` or remove `id="listbox"` and use only `[role="listbox"]`

cjpillsbury: agree. This is an artifact of previous iterations, including exploring groupings, but we should defer to `[role="listbox"]` and can always refactor if/when necessary

gkatsev (commented):
A couple of minor comments. I don't see an issue with the high level direction of these components.
Though, I did find this [blogpost](https://zellwk.com/blog/element-focus-vs-aria-activedescendant/) that claimed that activedescendent doesn't really work with VoiceOver (and potentially other screen readers)

src/js/media-chrome-listbox.js
cjpillsbury: nope! This was derived from the WAI design example, but I'd much rather use this. Will refactor and test. 🎉

heff:
I need...so much more context. Can you writeup an example guide for how a media chrome user would use this feature?

src/js/media-chrome-listbox.js
cjpillsbury: > I need...so much more context. Can you writeup an example guide for how a media chrome user would use this feature?

This is unlikely to be used directly by any non-advanced Media Chrome user. Rather, it'll be the core component for things like a settings overlay/dialog and captions flyaway menus (though there's a bit of discovery we'll have to do on whether or not we'll need to instrumentalize two different aria role configurations. It's possible we'd have a settings overlay/dialog implementation with slots that would result in less advanced users using this, but that's out of scope for this discussion imo.

cjpillsbury:
> I need...so much more context. Can you writeup an example guide for how a media chrome user would use this feature?

@heff This is unlikely to be used directly by any non-advanced Media Chrome user. Rather, it'll be the core component for things like a settings overlay/dialog and captions flyaway menus (though there's a bit of discovery we'll have to do on whether or not we'll need to instrumentalize two different aria role configurations. It's possible we'd have a settings overlay/dialog implementation with slots that would result in less advanced users using this, but that's out of scope for this discussion imo.

src/js/media-chrome-listbox.js
cjpillsbury: Per some investigation/callouts by @gkatsev , we may want to rely on item focus instead of `aria-activedescendant` based on sub-par support across different screen readers. I'll need to do some followup investigation, but wanted that noted here in case this comes up in discussion. Here's a very exhaustive post about this (albeit a couple of years old):
https://zellwk.com/blog/element-focus-vs-aria-activedescendant/

heff:
I think I have my head around the context now, thanks.

A major design decision I'm seeing is that lists auto-render themselves, compared to an example like  `<select>` or `<ul>` where it's still the responsibility of the _application_ to build the list markup. Are there specific reasons why these lists need to build themselves? With it all in the shadow dom it does then require things like custom renderers, whereas if the elements were in the light dom they could be customized as you would other elements.

I assume these aren't the only menus media chrome users will want to build in the context of the player, and not all of the menus may be driven by media data, so it does feel like we should start with the lower level components (list/menu container, list/menu item) and build up from there.

cjpillsbury:
> I think I have my head around the context now, thanks.
> 
> A major design decision I'm seeing is that lists auto-render themselves, compared to an example like `<select>` or `<ul>` where it's still the responsibility of the _application_ to build the list markup. Are there specific reasons why these lists need to build themselves? With it all in the shadow dom it does then require things like custom renderers, whereas if the elements were in the light dom they could be customized as you would other elements.
> 

The particular cases this is trying to solve generically would require a non-trivial amount of effort on the part of the integrator. Consider the three callout examples provided in the description:

- Subtitles/Captions Selection (included in the example)
- Rendition "Quality" Selection
- Alternate Audio (e.g. "dubs") Selection

When working with HTTP Adaptive Streaming ("HAS") standards like HLS/MMPEG-DASH, these are going to be described in and provided as part of the media source via the manifest/multivariant playlist. This means that, even in the easy case:

- They won't be provided in advance of the src, and are asynchronously added to the media
- They come in as "data"
- "Selecting" between them is tightly coupled to their definition and is also tied to data.

While it's possible this could be managed at the application level,  these feel like good candidates for a data driven approach. If you can provide an example that wouldn't be especially complicated where an implementor can solve these problems reliably and generically, I could certainly be convinced otherwise. I'm just not seeing any myself.

> I assume these aren't the only menus media chrome users will want to build in the context of the player, and not all of the menus may be driven by media data, so it does feel like we should start with the lower level components (list/menu container, list/menu item) and build up from there.

Here, I didn't want to "muddy the waters." By far the biggest ask we've gotten related to these things in both mux player and media chrome are a "quality selector" of some sort and a "captions/subtitles" selector. These are also on our player roadmap. So, rather than trying to solve both (1) the declarative/DOM use case (which is how I'd describe your callout here and I agree we definitely will want) and (2) the "data-driven" use case (as described above and what this draft PR spike is exploring), I went with trying to solve the narrower problem of just (2) based on the priorities.

I _**believe**_ my current spike solution could be updated without too much difficulty so that the it could share a "common core" architecture/component set for _**both**_ (1) and (2), but I wanted to avoid implementing to that goal (while still anticipating that in how the components are built). This is partly because was less obvious that we'd need custom web components for that just yet (rather than simply using e.g. `<ul>` directly, partly because it isn't obvious to me that we'd want them to be built on the same components, and partly because that would slow down progress on the goals here.

That said, this PR is intentionally a draft for discussion, and it intentionally hasn't solved most of the "nitty gritty UI" details. I can definitely see the final non-draft PR benefiting from custom components that solve those problems, and in a way that would at least bring this closer to solving (parts of?) (1).

src/js/media-chrome-listbox.js
```diff
@@ -0,0 +1,394 @@
...
+
+const isNil = (x) => x == undefined;
+const isIterable = (value) => {
+  return Symbol.iterator in Object(value);
+};
+const isArraylike = (value) => {
+  return (
+    typeof value?.length === 'number' &&
+    value.length >= 0 &&
+    value.length <= 4294967295
```
cjpillsbury, line 47: Dylan ask: make this "Magic number" a constant (e.g. `MAX_ARRAY_LENGTH`)

src/js/media-captions-list.js
```diff
@@ -0,0 +1,130 @@
...
+    super({ format });
+  }
+
+  attributeChangedCallback(attrName, oldValue, newValue) {
+    if (
+      [
+        MediaUIAttributes.MEDIA_SUBTITLES_LIST,
+        MediaUIAttributes.MEDIA_CAPTIONS_LIST,
+      ].includes(attrName) &&
+      oldValue !== newValue
```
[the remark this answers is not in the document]
cjpillsbury: Unfortunately, `attributeChangedCallback` is invoked whenever an attribute is set, even when the attribute has not, in fact changed. E.g. this will cause `attributeChangedCallback` to get invoked:

```js
const mclEl = document.querySelector('media-captions-list');
mclEl.setAttribute('media-subtitles-list', mclEl.getAttribute('media-subtitles-list'));
```

Since we decided (while implementing the "change" events) to not make media controller only propagate state if that state has, in fact, changed, this is a realistic possibility in our current architecture.

src/js/media-captions-list.js
```diff
@@ -0,0 +1,130 @@
...
+      [
+        MediaUIAttributes.MEDIA_SUBTITLES_SHOWING,
+        MediaUIAttributes.MEDIA_CAPTIONS_SHOWING,
+      ].includes(attrName) &&
+      oldValue !== newValue
+      && !this.stateUpdaterId
+    ) {
+      // Discuss this simple scheduling as a means of avoiding edge case updates as a result of 
+      // unsetting + resetting via distinct events + separation of captions vs. subtitles (CJP)
+      this.stateUpdaterId = setTimeout(() => {
```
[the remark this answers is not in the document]
cjpillsbury: @heff This is likely very safe, since it's going to defer to whatever the _**attribute**_ is (not the value passed in via the `attributeChangedCallback`. That said, I agree that its use case here is both a code sniff and, in my opinion, a workaround for some decisions earlier on where we made a "left" instead of a "right." These are related less to the generic "list" functionality and more to how we're modeling state for captions & subtitles. For example, even if we keep all of our existing attributes and events, I believe this code would be much cleaner/clearer if we had:

1. just one attribute for the list of both subtitles **_and_** captions
2. just one attribute for the currently selected subtitle **_or_** caption
3. just one event (figuring out the right name may be difficult) that will both "show" a given subtitle _**or**_ caption _**and**_ disable any currently showing subtitles/captions.

All of the acrobatics in `media-captions-list` are a result of the data coming from multiple attributes and the event to disable + the event to enable happening sequentially and in the same call stack, each of which will change the showing state of either captions or subtitles, so consecutively may change both.

I wanted to keep our existing larger architecture in media chrome "as is" with this spike effort, but I actually think it may be valuable to revisit some of those decisions now, given what this implementation effort is exposing.

heff:
> So, rather than trying to solve both (1) the declarative/DOM use case...I went with trying to solve the narrower problem of just (2) based on the priorities

I definitely appreciate the desire to narrow the scope, but inherent in this task is creating the first "menu item" in the world of media chrome. At some point someone is going to want to set a lot of variables like `--menu-item-background: #ccc`, and have it impact every menu, regardless of how it's generated. And non-data-driven menus are going to need all the same a11y. So while we don't have to jump right to solving every use case with this first project, we should have the understanding that the declarative (or custom menu) use case will ultimately use the same components for building a menu, not be a completely separate thing.

So then I do question why not solve the declarative use case as part of this? How much harder is it to define a "menu"-like component and "menu-item"-like component that can be used independently, and that the data-driven list generation is built on top of?

> If you can provide an example that wouldn't be especially complicated where an implementor can solve these problems reliably and generically, I could certainly be convinced otherwise.

I'm not exactly following your concern, so I'm just gonna jump to how I'd expect this to be implemented otherwise, and you can poke holes in it. I'm gonna use the term "menu" because it feels clearest for the example, but I know you have concerns with that term. 

Starting with 3 formal components:
* media-captions-menu-button
* media-chrome-menu
* media-chrome-menu-item

With those three components, there's not really a need for a special MediaCaptionsList component, the captions menu button can internally build the menu from the generic menu components. That would at least prevent some duplicate processing of the captions/subtitles attributes by both the button and a special menu. Starting there, what issues do you see? 

I think I'm seeing it like we have these needs:
1. We need the core UI components to build a menu in media chrome (i.e. menu box, menu item)
2. Depending on how those work, we might need a way to make it easy to generate the items of a menu from an array of data. (It's not particularly hard to generate a list of `<option>`s for a `<select>` menu. Can we make our menu components that easy?)

It feels like we're jumping to number 2, but we're not actually avoiding number 1, and if we're intentional about the primitives of number 1 it could change what we need in number 2. Does that makes sense?

cjpillsbury:
@gkatsev closing this since it's only here for reference at this point. Will keep branch alive though. @ me if you think it would be more convenient for it to continue living and I'll re-open.

PR #286, opened 2022-08-23
Add hotkeys blocklist

gkatsev:


docs/media-controller.md
```diff
@@ -322,3 +346,14 @@ class MyInteractiveElement extends window.HTMLElement {
...
+
+### hotKeys property
+
+A `hootKeys` property is available on the Media Controller. It is an [AttributeTokenList](../src/js/utils/attribute-token-list.js), which is based on the [DOMTokenList API](https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList) (like `classList`). This allows you to add and remove which shortcuts are allowed.
+
+Since, by default all the hotkeys listed above work, if you only wanted `Space` to toggle playback and `f` to toggle fullscreen, and didn't want the other functionality, you can add `Space` and `f` to the `hotKeys`. This will disable the other hotkeys.
+```js
+const mc = document.querySelector('media-controller')
+mc.hotKeys.add('f');
+mc.hotKeys.adD('Space');
```
cjpillsbury, line 358: typo(blocking): `adD`

docs/media-controller.md
```diff
@@ -322,3 +346,14 @@ class MyInteractiveElement extends window.HTMLElement {
   }
 }
 ```
+
+### hotKeys property
+
+A `hootKeys` property is available on the Media Controller. It is an [AttributeTokenList](../src/js/utils/attribute-token-list.js), which is based on the [DOMTokenList API](https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList) (like `classList`). This allows you to add and remove which shortcuts are allowed.
```
cjpillsbury, line 352: typo(blocking): `hootKeys` (🦉🎹)

cjpillsbury (changes requested):
haven't reviewed code yet, but caught a couple of docs nits/typos.

src/js/media-controller.js
```diff
@@ -40,12 +41,15 @@ const DEFAULT_SEEK_OFFSET = 10;
...
-    return super.observedAttributes.concat('nohotkeys');
+    return super.observedAttributes.concat('nohotkeys', 'hotkeys');
   }
 
+  #hotKeys = new AttributeTokenList(this, 'hotkeys');
+
   constructor() {
     super();
 
+
```
cjpillsbury, line 52: nit: extra whitespace

src/js/media-controller.js
```diff
@@ -775,6 +792,12 @@ class MediaController extends MediaContainer {
     switch (e.key) {
       case ' ':
       case 'k':
+        if (!this.#hotkeyEnabled('Space') &&
```
cjpillsbury, line 795: suggestion(non-blocking): Should we just check:

```js
if (!this.#hotkeyEnabled(e.key) return;
```
once up top?

cjpillsbury (changes requested):
A few non-blocking callouts in the code.

src/js/media-controller.js
```diff
@@ -487,10 +491,15 @@ class MediaController extends MediaContainer {
   attributeChangedCallback(attrName, oldValue, newValue) {
     if (attrName === 'nohotkeys') {
       if (newValue !== oldValue && newValue === '') {
+        if (this.hasAttribute('hotkeys')) {
+          console.warn('Both `hotkeys` and `nohotkeys` have been enabled. No hotkeys will trigger functionality.');
```
luwes, line 495: `No hotkeys will trigger functionality` read a bit strange to me. 
maybe something like "all hotkeys are disabled / deactivated" would be more clear.

src/js/media-controller.js
luwes: yes, would be neat

docs/media-controller.md
gkatsev: 🦉

src/js/utils/attribute-token-list.js
```diff
@@ -0,0 +1,95 @@
...
+    return this.#tokens.keys();
+  }
+
+  forEach(callback) {
+    this.#tokens.forEach(callback);
+  }
+
+  add(...tokens) {
+    tokens.forEach((t) => {
+      if (!this.contains(t)) this.#tokens.push(t);
```
gkatsev, line 51: Should we be case-insensitive? Perhaps with a `t.toLowerCase()` here?

Thinking about `<media-controller hotkeys="noarrowleft">`, since `event.key` would be `ArrowLeft`, and might make it nicer if we allowed `noF` for clarity.

cjpillsbury: we're effectively case-insensitive based on your "early bail" logic in the handler.

cjpillsbury (approved):

src/js/utils/attribute-token-list.js
gkatsev: The difference is if I do `mc.hotKeys.add('noArrowLeft')`. Then, it won't turn things off.

gkatsev:
Also, thanks to @luwes for writing the AttributeTokenList that I borrowed here. Made it super easy to implement.

src/js/utils/attribute-token-list.js
cjpillsbury: I don't feel strongly on this generally, but I think we should "bias for action" and ship what we have, since we can add your suggestion as a followup in a non-breaking way.

PR #291, opened 2022-08-31
POC of handling media errors in media chrome

heff:
The first pass is how I'd expect to handle PiP-not-ready errors.

Assume where `alert()`s happen, it means:
1. log to the console for the media chrome developer
3. propagate a [currently undefined] error state that an error display component could use to inform the viewer what's happening

There's a few decisions baked in:
1. Don't assume the media has the needed APIs to avoid the error (readyState in this case)
2. When it does have the API, use it, and make viewer-friendly error messaging possible
4. If the error can't be avoided (e.g. readyState isn't supported), handle known errors triggered by the viewer (InvalidStateError in this case) and make viewer friendly error messaging possible. 
5. Don't silence unknown errors from the media element

Thoughts on all that?

### DIFF
--- a/src/js/media-controller.js
+++ b/src/js/media-controller.js
@@ -159,14 +159,31 @@ class MediaController extends MediaContainer {
       MEDIA_ENTER_PIP_REQUEST: () => {
         const media = this.media;
 
-        if (!document.pictureInPictureEnabled) return;
+        if (!document.pictureInPictureEnabled) {
+          alert('Picture-in-picture is not supported.');
+        };
+
+        if (!media.requestPictureInPicture) {
+          alert('This media does not support picture-in-picture.');
+        }
+
+        if (typeof media.readyState === 'number' && media.readyState == 0 ) {
+          alert('The video is not ready yet.');
+        }
 
         // Exit fullscreen if needed
         if (document[fullscreenApi.element]) {
           document[fullscreenApi.exit]();
         }
 
-        media.requestPictureInPicture();
+        media.requestPictureInPicture().catch(err => {
+          if (err.code === 11) {
+            // InvalidStateError
+            alert('The video is not ready yet.');
+          } else {
+            throw err;
+          }
+        });
       },
       MEDIA_EXIT_PIP_REQUEST: () => {
         if (document.pictureInPictureElement) {
--- a/src/js/media-controller.js
+++ b/src/js/media-controller.js
@@ -187,7 +187,7 @@ class MediaController extends MediaContainer {
             // if we can rely on readyState and preload
             // Only works in Chrome currently. Safari doesn't allow triggering
             // in an event listener. Also requires readyState == 4.
-            // Firefox doesn't do PiP yet.
+            // Firefox doesn't have the PiP API yet.
             if (media.readyState === 0 && media.preload === 'none') {
               function cleanup() {
                 media.removeEventListener('loadedmetadata', tryPip);
--- a/src/js/media-controller.js
+++ b/src/js/media-controller.js
@@ -191,6 +191,7 @@ class MediaController extends MediaContainer {
             if (media.readyState === 0 && media.preload === 'none') {
               function cleanup() {
                 media.removeEventListener('loadedmetadata', tryPip);
+                media.preload = 'none';
               }
 
               function tryPip() {
@@ -202,7 +203,7 @@ class MediaController extends MediaContainer {
               media.preload = 'metadata';
 
               // No easy way to know if this failed and we should clean up
-              // quickly if it doesn't to prevent other issues
+              // quickly if it doesn't to prevent an awkward delay for the user
               setTimeout(()=>{
                 if (media.readyState === 0) warnNotReady();
                 cleanup();
--- a/src/js/media-controller.js
+++ b/src/js/media-controller.js
@@ -160,15 +160,15 @@ class MediaController extends MediaContainer {
         const media = this.media;
 
         if (!document.pictureInPictureEnabled) {
-          alert('Picture-in-picture is not supported.');
+          console.warn('MediaChrome: Picture-in-picture is not enabled');
+          // Placeholder for emitting a user-facing warning
+          return;
         };
 
         if (!media.requestPictureInPicture) {
-          alert('This media does not support picture-in-picture.');
-        }
-
-        if (typeof media.readyState === 'number' && media.readyState == 0 ) {
-          alert('The video is not ready yet.');
+          console.warn('MediaChrome: The current media does not support picture-in-picture');
+          // Placeholder for emitting a user-facing warning
+          return;
         }
 
         // Exit fullscreen if needed
@@ -176,11 +176,43 @@ class MediaController extends MediaContainer {
           document[fullscreenApi.exit]();
         }
 
+        const warnNotReady = (err) => {
+          console.warn('MediaChrome: The media is not ready for picture-in-picture. It must have a readyState > 0.');
+        };
+
         media.requestPictureInPicture().catch(err => {
+          // InvalidStateError, readyState == 0 (Not ready)
           if (err.code === 11) {
-            // InvalidStateError
-            alert('The video is not ready yet.');
+            // We can assume the viewer wants the video to load, so attempt to
+            // if we can rely on readyState and preload
+            // Only works in Chrome currently. Safari doesn't allow triggering
+            // in an event listener. Also requires readyState == 4.
+            // Firefox doesn't do PiP yet.
+            if (media.readyState === 0 && media.preload === 'none') {
+              function cleanup() {
+                media.removeEventListener('loadedmetadata', tryPip);
+              }
+
+              function tryPip() {
+                media.requestPictureInPicture().catch(warnNotReady);
+                cleanup();
+              }
+
+              media.addEventListener('loadedmetadata', tryPip);
+              media.preload = 'metadata';
+
+              // No easy way to know if this failed and we should clean up
+              // quickly if it doesn't to prevent other issues
+              setTimeout(()=>{
+                if (media.readyState === 0) warnNotReady();
+                cleanup();
+              }, 1000);
+            } else {
+              // Rethrow if unknown context
+              throw err;
+            }
           } else {
+            // Rethrow if unknown context
             throw err;
           }
         });


cjpillsbury:
This seems reasonable to me to solve the one off case in a way that also puts some consideration on forward-looking usage. Below are my thoughts for immediate & longer term. tl;dr - For right now, I think this would be 👌 (with one minor callout). However, we may want to consider some of the longer term things, insofar as they would impact the API/expectations. I personally don't think any of them should be blockers for the current (narrower) proposed solution, but wanted visibility on them just in case others disagree.

> The first pass is how I'd expect to handle PiP-not-ready errors.
> 
> Assume where `alert()`s happen, it means:
> 
> 1. log to the console for the media chrome developer

👍 Let's explicitly say we'll `console.error()` or `console.warn()`. I can see arguments for either, so I defer to others on which is better.

> 2. propagate a [currently undefined] error state that an error display component could use to inform the viewer what's happening

Should we add an `error` getter to `MediaController`, similar to `HTMLMediaElement`, to avoid any "if a tree falls in the woods and yadda yadda" problems? There would still be things to work out (when to reset, if it should also represent any slotted `media.error` cases, etc.) if we did want to add it. On the plus side, this would be a step in the right direction for the error dialog support/implementation we have in Mux Player. However, since it would increase the footprint of things to resolve, I don't think this should be a blocker for this first pass (even if it does entail a plausible API change longer term to handle cases + UIs like Mux Player supports).

> 
> There's a few decisions baked in:
> 
> 1. Don't assume the media has the needed APIs to avoid the error (readyState in this case)

👍 If we're not going to assume `media.readState` on the slotted media, should we also not assume `media.requestPictureInPicture()`? If not, should we also provide an error in this case?

> 2. When it does have the API, use it, and make viewer-friendly error messaging possible

This is forward more looking, correct? If yes, I think we should
1. Update the text.
2. Minimally move it to a constant or better a function to anticipate i18n.

> 3. If the error can't be avoided (e.g. readyState isn't supported), handle known errors triggered by the viewer (InvalidStateError in this case) and make viewer friendly error messaging possible.

We may want to consider dispatching an `error` event (with a potential `ErrorEvent` if we don't implement an `error` getter, per above) for both known and unknown errors instead of re-throwing the error. This would allow generic error monitoring from developers without causing an actual crash (though there are some benefits to unexpected crashes). Here again, I don't think this should be a blocker for this effort, but it's food for thought.

> 4. Don't silence unknown errors from the media element
> 

👍 agree, either with my proposal under (3) or in the current impl.

> Thoughts on all that?

See above. 😄

luwes:
I like this approach.

@cjpillsbury I would initially hold off on an error handling property and event until it comes up. Is it really needed?

Isn't it quite easy for a developer to disable pointer-events on the MC buttons if the `src` is not set and `media-ready-state` is not sufficient. (add `media-ready-state` as a media-controller attribute?) 

For PiP specifically if the src is set I would potentially just go ahead and set `preload=auto` and call the API once ready state is good. They clearly want to enter PiP if they press that button.

cjpillsbury:
> @cjpillsbury I would initially hold off on an error handling property and event until it comes up. Is it really needed?

I wouldn't say it's "needed" and definitely shouldn't block anything here. The only cases it would be handling are ones where a media state receiver is registered after the an error occurs (or is unregistered and then re-registered). This is what I meant by "if a tree falls in the woods" cases. With the exception of the thumbnail/preview attributes (https://github.com/muxinc/media-chrome/blob/main/src/js/media-controller.js#L250-L258), our state is modeled as actual state (see, e.g. https://github.com/muxinc/media-chrome/blob/main/src/js/media-controller.js#L635). This would be modeling errors more like an event. That may be fine for errors specifically, since they're a pretty different beast than other things, and modeling them as state has its own complexities. But it does mean it will behave differently than our other state.

> Isn't it quite easy for a developer to disable pointer-events on the MC buttons if the `src` is not set and `media-ready-state` is not sufficient. (add `media-ready-state` as a media-controller attribute?)

Yup this was discussed (or some variation of this). I'm on board with adding additional state for folks to use for CSS selector behavior. Not to put words into @heff's mouth, but I believe his take here is "that's a different problem that maybe we don't need to solve now (or possibly ever)." Whether that's a flesh and blood heff or one hanging out in my imagination, I think that's also right (per his push back to me in the discussion thread [here](#288))

> For PiP specifically if the src is set I would potentially just go ahead and set `preload=auto` and call the API once ready state is good. They clearly want to enter PiP if they press that button.

This one actually hadn't come up in our prior discussion, but I like where your head's at. If we can be fairly confident about a user **_intention_** (they made a media enter pip **_request_**), then we should try to meet that intention (aka start the load for them in this case). Probably a better way to meet that intention would be to use `media.load()`, but it's the same general idea as you're proposing.

luwes:
Sounds good, would be a nice stepping stone to have those extra attributes like `media-ready-state` or similar and make it very easy on the user to disable buttons based on that. 

Not sure how it would work if the components are not children of the media-controller though? 🤔 

--- 

In regard to preloading for PiP `video.load()` is not the right API for that I believe. If you check the spec it's really a reset of the src. https://html.spec.whatwg.org/multipage/media.html#loading-the-media-resource

Setting `video.preload` to `metadata` or `auto` should be sufficient I think.

heff:
> console.error() or console.warn()

I don't have a strong opinion here so defer to anyone who does. Are there definitions for those functions that can guide us?

> Should we add an error getter to MediaController, similar to HTMLMediaElement, to avoid any "if a tree falls in the woods and yadda yadda" problems?

Yeah, I was imagining that whatever state was propagated to UI components would also be reflected at least as an attribute media controller

> There would still be things to work out (when to reset, if it should also represent any slotted media.error cases, etc.)

Definitely. Also, I think the errors we're describing are very different from media errors. At minimum media errors tend to fatal, whereas "video not ready" is more ephemeral. e.g. For a media error I'd expect a big, uncloseable "This video has failed" error message. Clicking the PiP button too soon I'd expect more of a warning, that probably goes away on its own after a few seconds. So more to design there for sure, and if people have examples of that difference being defined well, call them out. My gut is new `media-error` and `media-warning`  attributes.

> should we also not assume media.requestPictureInPicture()

Yep. 👍 I added a check and error for it. And that obviously has implications for media element functions.

> This is forward more looking, correct? If yes, I think we should
> Update the text.
> Minimally move it to a constant or better a function to anticipate i18n.

Yeah, all the text is just for POC purposes. All that sounds good.

> We may want to consider dispatching an error event ... instead of re-throwing the error

Definitely don't want to cause unhelpful crashes. What's the discoverability of an error event? Would it still show up in the console?

> For PiP specifically if the src is set I would potentially just go ahead and set preload=auto

I like this general design direction. It's similar to [this old note I just remembered](https://github.com/muxinc/media-chrome/blob/main/src/js/media-time-range.js#L156), which addresses the ready state and seeking issue called out in the other discussion. Framed as a principle might be: "When necessary and possible, take the extra steps to achieve the user's intent for an action/request, in cases where the media element API adds additional, unnecessary friction."

heff:
Cleaned up the POC to make it ready to merge if approved, removing forward looking things.
Added an attempt to load the video if it just wasn't ready for PiP. Didn't work in Safari because it denied triggering PiP in an event listener.

src/js/media-controller.js
```diff
@@ -159,14 +159,63 @@ class MediaController extends MediaContainer {
...
+                media.removeEventListener('loadedmetadata', tryPip);
+              }
+
+              function tryPip() {
+                media.requestPictureInPicture().catch(warnNotReady);
+                cleanup();
+              }
+
+              media.addEventListener('loadedmetadata', tryPip);
+              media.preload = 'metadata';
```
cjpillsbury, line 202: nitpick(blocking): If we're going to use `media.preload`, we should reset `media.preload = 'none';` in `cleanup()` to not cause permanent side effects.

src/js/media-controller.js
```diff
@@ -159,14 +159,63 @@ class MediaController extends MediaContainer {
       MEDIA_ENTER_PIP_REQUEST: () => {
         const media = this.media;
 
-        if (!document.pictureInPictureEnabled) return;
+        if (!document.pictureInPictureEnabled) {
+          console.warn('MediaChrome: Picture-in-picture is not enabled');
```
cjpillsbury, line 163: nitpick(non-blocking): Per prior comment, we should consider move these inline warning strings to constants to anticipate i18n.

src/js/media-controller.js
```diff
@@ -159,14 +159,63 @@ class MediaController extends MediaContainer {
...
+
+              media.addEventListener('loadedmetadata', tryPip);
+              media.preload = 'metadata';
+
+              // No easy way to know if this failed and we should clean up
+              // quickly if it doesn't to prevent other issues
+              setTimeout(()=>{
+                if (media.readyState === 0) warnNotReady();
+                cleanup();
+              }, 1000);
```
cjpillsbury, line 209: question(non-blocking): Is there anything we can do to be less reliant on this? Have we done any testing to see if 1 second is long enough when e.g. simulating poor network conditions?

src/js/media-controller.js
```diff
@@ -159,14 +159,63 @@ class MediaController extends MediaContainer {
...
+
+        media.requestPictureInPicture().catch(err => {
+          // InvalidStateError, readyState == 0 (Not ready)
+          if (err.code === 11) {
+            // We can assume the viewer wants the video to load, so attempt to
+            // if we can rely on readyState and preload
+            // Only works in Chrome currently. Safari doesn't allow triggering
+            // in an event listener. Also requires readyState == 4.
+            // Firefox doesn't do PiP yet.
+            if (media.readyState === 0 && media.preload === 'none') {
```
cjpillsbury, line 191: question(non-blocking): Is there a reason we're only retrying if `media.preload === 'none'`? Will this edge case not arise if someone requests PIP when `media.readyState === 0` and `media.preload` is `"metadata"` or `"auto"` (aka attempted to enter PIP faster than the src's metadata load)?

src/js/media-controller.js
```diff
@@ -159,14 +159,63 @@ class MediaController extends MediaContainer {
...
+        };
+
+        media.requestPictureInPicture().catch(err => {
+          // InvalidStateError, readyState == 0 (Not ready)
+          if (err.code === 11) {
+            // We can assume the viewer wants the video to load, so attempt to
+            // if we can rely on readyState and preload
+            // Only works in Chrome currently. Safari doesn't allow triggering
+            // in an event listener. Also requires readyState == 4.
+            // Firefox doesn't do PiP yet.
```
cjpillsbury, line 190: > Firefox doesn't do PiP yet.

polish(non-blocking) This isn't quite accurate. See https://support.mozilla.org/en-US/kb/about-picture-picture-firefox

cjpillsbury (changes requested):
one minor blocking ask and a few non-blocking callouts. Otherwise LGTM.

src/js/media-controller.js
```diff
@@ -159,14 +159,63 @@ class MediaController extends MediaContainer {
...
+
+              // No easy way to know if this failed and we should clean up
+              // quickly if it doesn't to prevent other issues
+              setTimeout(()=>{
+                if (media.readyState === 0) warnNotReady();
+                cleanup();
+              }, 1000);
+            } else {
+              // Rethrow if unknown context
+              throw err;
```
cjpillsbury, line 212: thought(non-blocking): As per before, I don't think this discussion should block this PR, but it's worth noting that, since these errors are a result of async behavior, they will be impossible for an implementor to `catch`.

src/js/media-controller.js
gkatsev: could change to "Firefox doesn't do the PiP *API* yet".

src/js/media-controller.js
heff: Is i18n needed/realistic for dev-facing console messages or is that more for future user-facing messages?

src/js/media-controller.js
heff: I'm open to other ideas on how to handle this, because I couldn't think of anything better. The 1s limit is as much about responsiveness after the user action as it is about being enough time to load. If we wait too much longer it could feel awkwardly delayed when the pip finally happens. But otherwise, loadedmetadata does tend to happen very fast, and falling back to a warning in slower load cases isn't the worst.

src/js/media-controller.js
heff: 👍 done

src/js/media-controller.js
heff: > attempted to enter PIP faster than the src's metadata load

I see how the logic is open for that but honestly that feels like such a rare situation that I don't think it's worth complicating this any further to support it. If it actually comes up I'll update it then.

src/js/media-controller.js
heff: 👍 done

src/js/media-controller.js
heff: Before this we were just calling `media.requestPictureInPicture();` and letting any promise error happen as it does, so there's at least nothing new here I can see. Unless there's a better way to just rethrow the promise error. But otherwise yeah, let's group this into a conversation around formal warnings.

heff:
I think this could be merged now if there's no other followup comments.

cjpillsbury (approved):
Ship it (🚢 🇮🇹)

luwes (approved):
👍

gkatsev:
FYI, this PR didn't get merged with a conventional changelog commit, so, it'll end up missing from the generated docs.

PR #293, opened 2022-09-06
Stop progress bar when media is loading

MattiasBuelens:
When playback stalls (because the backing media player fails to buffer quickly enough), the progress bar would incorrectly keep advancing. This fixes that, stopping the progress bar from advancing while the media is loading.


dylanjha:
Thanks for the PR @MattiasBuelens! We'll take a look

luwes (approved):
Good catch, ziet er goed uit! Merci @MattiasBuelens!

PR #296, opened 2022-09-08
fix: prevent hotkeys from scrolling page

### LINKED ISSUE #295: Spacebar press scrolls page
When pressing the space bar to pause video play the page scrolls until only the control bar with space above for the storyboard images is visible, obscuring part of the video. This is demonstrated at the codepen https://codepen.io/checkmark/pen/dyeGmGQ. The expected behavior is that only the video toggles playback on spacebar press. The page scroll is unexpected.

gkatsev:
If a key will scroll the page, such as Space, ArrowLeft, and ArrowRight, we want to prevent the default action so the page won't scroll.

Browsers scroll on keydown rather than keyup, so, we have to do the preventDefault separately from where we do the action. This has the side-effect that should someone choose to cancel an action -- for example, holding down space and then hitting escape to prevent playback toggling from happening -- then the default behavior, in the example, scrolling down a page, will still be prevented.

Fixes #295

### DIFF
--- a/src/js/media-controller.js
+++ b/src/js/media-controller.js
@@ -751,6 +751,18 @@ class MediaController extends MediaContainer {
       this.removeEventListener('keyup', this.#keyUpHandler);
       return;
     }
+
+    // if the pressed key might move the page, we need to preventDefault on keydown
+    // because doing so on keyup is too late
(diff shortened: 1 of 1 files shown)


cjpillsbury (approved):
LGTM

PR #298, opened 2022-09-09
feat: fix order of layers + add chrome CSS parts

luwes:
This change adds the correct order of the chrome containers so a11y tab order improves going from the top chrome elements to the center chrome elements and then to the bottom chrome elements.

Before it went from top -> bottom -> center 

This change makes use of disabling the CSS `display: contents` property on some `<slot>` elements essentially making them visible containers.

In addition some CSS parts were added for the different chrome sections which we'll be able to use in Mux Player.

### DIFF
--- a/src/js/media-container.js
+++ b/src/js/media-container.js
@@ -360,7 +360,7 @@ class MediaContainer extends window.HTMLElement {
       // so, only run pointermove for mouse
       if (e.pointerType !== 'mouse') return;
 
-      if (!containsComposedNode(this, e.target)) return;
+      if (e.target === this) return;
 
       setActive();
       // Stay visible if hovered over control bar
--- a/src/js/media-container.js
+++ b/src/js/media-container.js
@@ -69,6 +69,8 @@ template.innerHTML = `
     }
 
     :host(:not([audio])) *[part~=layer][part~=centered-layer] {
+      display: flex;
+      flex-grow: 1;
       align-items: center;
       justify-content: center;
     }
@@ -79,7 +81,9 @@ template.innerHTML = `
       flex-grow: 1;
     }
 
-    .spacer {
+    slot[name=middle-chrome] {
+      display: inline;
+      flex-grow: 1;
       pointer-events: none;
       background: none;
     }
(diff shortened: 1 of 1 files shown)


src/js/media-container.js
```diff
@@ -120,25 +124,17 @@ template.innerHTML = `
...
+  <slot name="poster" part="layer poster-layer"></slot>
+  <slot name="gestures-chrome" part="layer gesture-layer">
+    <media-gesture-receiver slot="gestures-chrome"></media-gesture-receiver>
+  </slot>
   <span part="layer vertical-layer">
-    <slot name="top-chrome"></slot>
-    <span class="spacer"><slot name="middle-chrome"></slot></span>
+    <slot name="top-chrome" part="top chrome"></slot>
+    <slot name="middle-chrome" part="middle chrome"></slot>
+    <slot name="centered-chrome" part="layer centered-layer center centered chrome"></slot>
```
gkatsev, line 135: question(blocking): any concerns over nesting the layers here?

gkatsev (commented):
The slots demo example page ended up being super useful for testing this out. Definitely helps order things in a more expected way and doesn't rely on tabindex which is great.

this PR: https://media-chrome-git-fork-luwes-fix-tab-order-mux.vercel.app/examples/slots-demo.html
main: https://media-chrome-mux.vercel.app/examples/slots-demo.html

src/js/media-container.js
luwes: initially it doesn't seem an issue. 
it could become an issue if a user would like the vertical layer to be a different size than 100% dimensions maybe or starting with backgrounds and opacity etc.

@heff @cjpillsbury was also wondering. was there a clear use for the middle-chrome slot? 

we could possibly rename the centered-layer to something more specific / local for this case and remove `layer centered-layer`. something like `part="center centered chrome overlay"`

cjpillsbury: I have some concerns with this DOM restructure, but since this is mostly advanced usage/"under the hood," I don't think these concerns should block the PR. 
tl;dr - centered-chrome is intended to be used as an independent layer/layout from the vertical layer/slots, so things may get weird around the edges for advanced use cases and potential future theme use cases. If those come up, we can always revist.

luwes:
I discovered a little issue with `user-inactive` not working correctly on pointermove. working on a fix

src/js/media-container.js
```diff
@@ -366,7 +360,7 @@ class MediaContainer extends window.HTMLElement {
       // so, only run pointermove for mouse
       if (e.pointerType !== 'mouse') return;
 
-      if (!containsComposedNode(this, e.target)) return;
+      if (e.target === this) return;
```
luwes, line 363: revert to this 
https://github.com/muxinc/media-chrome/commit/ac767acdc4218d42199e4abc6e1ebae2030c5f8a#diff-a093df03191bf45d0036001fefb7ca7217096f130642377d56e0cd90312b3e23R341

`containsComposedNode` did not work if `mux-video` was the target while it was not a direct child of media-controller but a "virtual" child via slots

cjpillsbury (approved):
🚢 🇮🇹

PR #301, opened 2022-09-13
fix: fix #299 right click on media element

### LINKED ISSUE #299: Gesture receiver element breaks <video> right click
We shouldn't prevent showing the extra options like loop, PiP, controls, etc. that are shown on a right click on the slotted media element. They are useful and expected by many users.

Smaller reason for devs, right clicking to go straight to the video DOM element in web inspector.

luwes:
prevents the media-gesture-receiver from capturing all the click events. instead the pointer events are attached to window.

still uses the bounding box of the gesture receiver elements to check if the click event was on it or not allowing for potential positioning of other gesture receiver elements.

### DIFF
--- a/src/js/media-gesture-receiver.js
+++ b/src/js/media-gesture-receiver.js
@@ -11,11 +11,7 @@ template.innerHTML = `
 <style>
   :host {
     display: inline-block;
-    width: auto;
-    height: auto;
     box-sizing: border-box;
-
-    pointer-events: auto;
   }
 </style>
 `;
@@ -24,7 +20,7 @@ class MediaGestureReceiver extends window.HTMLElement {
   // NOTE: Currently "baking in" actions + attrs until we come up with
   // a more robust architecture (CJP)
   static get observedAttributes() {
-    return [MediaUIAttributes.MEDIA_CONTROLLER, MediaUIAttributes.MEDIA_PAUSED];
+    return [MediaUIAttributes.MEDIA_CONTROLLER, MediaUIAttributes.MEDIA_PAUSED, MediaUIAttributes.MEDIA_CLICK];
   }
 
   constructor(options = {}) {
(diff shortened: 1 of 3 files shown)


src/js/media-gesture-receiver.js
```diff
@@ -54,9 +50,19 @@ class MediaGestureReceiver extends window.HTMLElement {
...
     };
 
-    this.addEventListener('pointerdown', pointerDownHandler);
+    window.addEventListener('pointerdown', pointerDownHandler);
+
+    // Cannot use composedPath or target because this is a layer on top and pointer events are disabled.
+    // Attach to window and check if click is in this element's bounding box to keep <video> right-click menu.
+    window.addEventListener('click', (event) => {
+      const { clientX, clientY } = event;
+      const { left, top, width, height } = this.getBoundingClientRect();
```
cjpillsbury, line 59: idea: could we instead only `getBoundingClientRect()` on `resize` events to avoid recomputing on any page click?

gkatsev: Would this be helpful as an alternative as well? https://developer.mozilla.org/en-US/docs/Web/API/Document/elementFromPoint

src/js/media-gesture-receiver.js
```diff
@@ -54,9 +50,19 @@ class MediaGestureReceiver extends window.HTMLElement {
       this._pointerType = e.pointerType;
     };
 
-    this.addEventListener('pointerdown', pointerDownHandler);
+    window.addEventListener('pointerdown', pointerDownHandler);
```
cjpillsbury, line 53: issue(blocker): let's move the adding of the event handler to `connectedCallback` and add a corresponding remove to `disconnectedCallback` (can add a private method for `pointerDownHandler`)

cjpillsbury (changes requested):
a couple of thoughts, but this is looking like the way we'll have to go to not block right clicks.

src/js/media-gesture-receiver.js
gkatsev: ah, elementFromPoint might not work due to the point-events:none thing

cjpillsbury (approved):
😎 awesome.

src/js/media-gesture-receiver.js
```diff
@@ -16,6 +15,12 @@ template.innerHTML = `
 </style>
 `;
 
+const resizeObserver = new ResizeObserver((entries) => {
+  for (const entry of entries) {
+    entry.target.rect = entry.target.getBoundingClientRect();
+  }
+});
```
luwes, line 22: I saw this a while back 1 resizeobserver vs multiple per element performs a lot better

gkatsev (approved):

PR #304, opened 2022-09-14
fix: `gestures-disabled` attribute

luwes:
related #301

setting any element to `display: none` in the gesture-layer will disable gestures on that element.


cjpillsbury (approved):
makes sense to me.

gkatsev (approved):

PR #308, opened 2022-09-20
feat: add a hotkey for toggling captions: c

gkatsev:
Adds a `c` hotkey to toggle captions.

This uses method delegation to the `MediaCaptionsButton`'s `handleClick` method because there's a lot of code there.

~Needs a quick addition to docs still, but also, due to the recent media-gestures-receiver change, the player doesn't get focus when click on the video element and thus keyboard shortcuts don't work there. Though, might want a fix for that separately from this.~

### DIFF
--- a/src/js/media-controller.js
+++ b/src/js/media-controller.js
@@ -17,7 +17,7 @@ import { AttributeTokenList } from './utils/attribute-token-list.js';
 import { fullscreenApi } from './utils/fullscreenApi.js';
 import { constToCamel } from './utils/stringUtils.js';
 import { containsComposedNode } from './utils/element-utils.js';
-import MediaCaptionsButton from './media-captions-button.js';
+import { toggleSubsCaps } from './utils/captions.js';
 
 import {
   MediaUIEvents,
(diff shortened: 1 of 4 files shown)


gkatsev:
Opened an issue for the issue: https://github.com/muxinc/media-chrome/issues/309

src/js/media-controller.js
```diff
@@ -17,6 +17,7 @@ import { AttributeTokenList } from './utils/attribute-token-list.js';
 import { fullscreenApi } from './utils/fullscreenApi.js';
 import { constToCamel } from './utils/stringUtils.js';
 import { containsComposedNode } from './utils/element-utils.js';
+import MediaCaptionsButton from './media-captions-button.js';
```
cjpillsbury, line 20: issue(blocking): Per last round of hotkeys, we're trying to avoid imports of components here. Either duplicate the logic in `handleClick` or pull it out into a reusable function.

cjpillsbury (changes requested):
One ask, otherwise 👌

src/js/media-controller.js
gkatsev: updated

luwes (approved):
LGTM

cjpillsbury (approved):
LGTM! This does mean we can get into a mismatch of toggle behavior between hotkeys vs. button if they're using `no-subtitles-fallback`, but I think that should be saved for a larger discussion/potential rearchitecture anyway.

PR #311, opened 2022-09-20
Disable fullscreen button if fullscreen is disabled

gkatsev:
Some inspiration for how to disable the button from https://www.smashingmagazine.com/2021/08/frustrating-design-patterns-disabled-buttons/

### DIFF
--- a/examples/iframe.html
+++ b/examples/iframe.html
@@ -1,0 +1,6 @@
+<iframe
+    src="./advanced.html"
+    width="736"
+    height="531"
+    allow="fullscreen 'none'"
+></iframe>
--- a/src/js/media-controller.js
+++ b/src/js/media-controller.js
@@ -132,6 +132,10 @@ class MediaController extends MediaContainer {
       //   - Element.requestFullscreen()
       //
       MEDIA_ENTER_FULLSCREEN_REQUEST: () => {
+        if (!document.fullscreenEnabled) {
+          console.warn('Fullscreen support is disabled, not entering fullscreen');
+          return;
+        }
         const media = this.media;
 
         if (document.pictureInPictureElement) {
--- a/src/js/media-fullscreen-button.js
+++ b/src/js/media-fullscreen-button.js
@@ -26,15 +26,21 @@ const exitFullscreenIcon = `<svg aria-hidden="true" viewBox="0 0 26 24">
 const slotTemplate = document.createElement('template');
 slotTemplate.innerHTML = `
   <style>
-  :host([${MediaUIAttributes.MEDIA_IS_FULLSCREEN}]) slot:not([name=exit]) > *, 
+  :host([${MediaUIAttributes.MEDIA_IS_FULLSCREEN}]) slot:not([name=exit]) > *,
   :host([${MediaUIAttributes.MEDIA_IS_FULLSCREEN}]) ::slotted(:not([slot=exit])) {
     display: none !important;
   }
 
   /* Double negative, but safer if display doesn't equal 'block' */
-  :host(:not([${MediaUIAttributes.MEDIA_IS_FULLSCREEN}])) slot:not([name=enter]) > *, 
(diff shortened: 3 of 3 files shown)


gkatsev:
After making the PR, I remembered about https://github.com/muxinc/media-chrome/discussions/134.
The question is does this make fullscreen an unavailable thing? In some ways, I think that the behavior of showing the button but having it be disabled is the correct one for this. For example, if you're putting your player in an iframe, you can see that the button doesn't work and maybe you won't notice if the button isn't there altogether.

examples/iframe.html
```diff
@@ -0,0 +1,6 @@
+<iframe
+    src="./advanced.html"
+    width="736"
+    height="531"
+    allow="fullscreen 'none'"
```
gkatsev, line 5: this is the syntax for disallowing a particular feature.

src/js/media-controller.js
```diff
@@ -132,6 +132,10 @@ class MediaController extends MediaContainer {
       //   - Element.requestFullscreen()
       //
       MEDIA_ENTER_FULLSCREEN_REQUEST: () => {
+        if (!document.fullscreenEnabled) {
```
gkatsev, line 135: we don't want to try and fullscreen if it isn't enabled.

examples/iframe.html
gkatsev: I think I need to explicitly disallow it because otherwise the iframe is from the same domain and allows it by default?

src/js/media-fullscreen-button.js
```diff
@@ -26,16 +26,22 @@ const exitFullscreenIcon = `<svg aria-hidden="true" viewBox="0 0 26 24">
...
-  :host(:not([${MediaUIAttributes.MEDIA_IS_FULLSCREEN}])) slot:not([name=enter]) > *, 
+  :host(:not([${MediaUIAttributes.MEDIA_IS_FULLSCREEN}])) slot:not([name=enter]) > *,
   :host(:not([${MediaUIAttributes.MEDIA_IS_FULLSCREEN}])) ::slotted(:not([slot=enter])) {
     display: none !important;
   }
+
+  :host(:disabled),
+  :host([aria-disabled]) {
+    cursor: not-allowed;
+    opacity: 60%;
```
gkatsev, line 43: this one is tricky, since it might make the button not pass the contrast ratio req, but not sure if there's another good way of making it visually disabled. Of course, may not be necessary, if we decide that if `document.fullscreenEnabled === false` means the button shouldn't show up.

examples/iframe.html
dylanjha: Makes sense. Interesting how the syntax of this API is different from something like controlslist

luwes:
> After making the PR, I remembered about #134. The question is does this make fullscreen an unavailable thing? In some ways, I think that the behavior of showing the button but having it be disabled is the correct one for this. For example, if you're putting your player in an iframe, you can see that the button doesn't work and maybe you won't notice if the button isn't there altogether.

I can get on board with that but for this first change I would keep it consistent with pip, airplay, cast, etc.
the users still are free to do what they like with that `[media-fullscreen-unavailable]` attribute CSS selector.
I would say yes, it makes fullscreen unavailable in that context. it's disabled for that user.

gkatsev:
yeah, I'll update this PR to make a fullscreen-unavailable. Based on the discussion in #134 it seems like media-chrome doesn't do default styles for unavailable currently.

src/js/media-controller.js
```diff
@@ -132,6 +135,10 @@ class MediaController extends MediaContainer {
...
+          console.warn('Fullscreen support is disabled, not entering fullscreen');
```
gkatsev, line 139: can use help wordsmithing this.

cjpillsbury: suggestion(non-blocking): `"Fullscreen support is unavailable; not entering fullscreen"`? also is perfectly fine as is imo.

cjpillsbury (approved):
😎

PR #312, opened 2022-09-20
fix: set tabindex on media element, if not set

### LINKED ISSUE #309: Hotkeys don't work when clicking on the player outside of buttons
This was due to the recent update to media-gestures-receiver (#301). Now, the player doesn't get focus on clicked inside the player area, but not on a button. Instead, the focus stays on the `body`. This means that hotkeys don't work when "focusing on the player" since the keyboard handlers are on the media-controller.

A potential fix is adding `tabindex=-1` explicitly on the video element or the media-controller. This would make focus change to be inside the player.

Another potential fix, which is less than ideal is to check to see if the last mouse click was inside the player and then allow global keyup handlers to handle keys.

There might be other solutions but not sure right now.

gkatsev:
This is so that clicking inside the player gets focus changed to be inside the player and hotkeys work as expected. If a tabindex is already set, we shouldn't override it. Otherwise, we should set it to be -1 so that clicking would change focus but it wouldn't be added to the regular tab order.

Since the primary usecase of this is for hotkeys, there will be keyboard interactions with the player after the focus change. However, this does mean that an outline will be added to the media element. This probably isn't an issue when windowed, but in fullscreen, we definitely don't want an outline border on the media.

Fixes #309

### DIFF
--- a/src/js/media-container.js
+++ b/src/js/media-container.js
@@ -21,6 +21,16 @@ const template = document.createElement('template');
 
 template.innerHTML = `
   <style>
+
+    /*
+     * outline on media is turned off because it is allowed to get focus to faciliate hotkeys.
+     * However, on keyboard interactions, the focus outline is shown,
+     * which is particularly noticeable when going fullscreen via hotkeys.
+     */
+    :host([media-is-fullscreen])  ::slotted([slot=media]) {
+      outline: none;
+    }
+
     :host {
       box-sizing: border-box;
       position: relative;
--- a/src/js/media-controller.js
+++ b/src/js/media-controller.js
@@ -556,6 +556,11 @@ class MediaController extends MediaContainer {
 
   mediaSetCallback(media) {
     super.mediaSetCallback(media);
+
+    if (!media.hasAttribute('tabindex')) {
+      media.setAttribute('tabindex', -1);
+    }
+
     // Listen for media state changes and propagate them to children and associated els
     Object.keys(this._mediaStatePropagators).forEach((key) => {
       const events = key.split(',');


src/js/media-container.js
```diff
@@ -21,6 +21,16 @@ const template = document.createElement('template');
 
 template.innerHTML = `
   <style>
+
+    /*
+     * outline on media is turned off because it is allowed to get focus to faciliate hotkeys.
+     * However, on keyboard interactions, the focus outline is shown,
+     * which is particularly noticeable when going fullscreen via hotkeys.
+     */
+    :host([media-is-fullscreen])  ::slotted([slot=media]) {
```
gkatsev, line 30: This is a compromise, though, the case can be made that there's no reason to ever show the outline on the media element.

src/js/media-controller.js
```diff
@@ -556,6 +556,11 @@ class MediaController extends MediaContainer {
 
   mediaSetCallback(media) {
     super.mediaSetCallback(media);
+
+    if (!media.hasAttribute('tabindex')) {
```
gkatsev, line 560: this is a bit unusual, but I think alternatives for fixing #309 are probably a lot worse.

luwes (approved):
LGTM

cjpillsbury (approved):
👌

cjpillsbury:
@heff would be good to get your explicit stamp on this one in case you have any concerns from a "best practices" perspective.

heff (approved):
Seems like a smart fix for the issue. A few questions come to mind:

1. How do other players avoid/handle this? A lot seem to be one-player-per-page examples (watch page, iframe) in which case you're just catching any key board events at the window level. Do we know of other many-player examples?

2. Should the player container just be in the tab order anyway?

But approved. 👍 Ship it and let's see if we get any feedback.

gkatsev:
Video.js has the player be focusable with a click via `tabindex=-1`. Looks like YouTube has a mix of global key binding as well as player key-binding. YouTube also has `tabindex=-1` on the player element, but interestingly, tabbing through is getting the control bar elements. When I was testing this change, I originally was thinking of adding `tabindex=-1` to the media-controller, but it didn't let me tab into the control bar buttons without first focusing into the player. Though, it's possible that YouTube is doing some other weird things as well.
Vimeo seems to have global keybindings on the page.

I do think that ultimately `tabindex=-1` might be correct, but there's some accessibility/SR/taborder things that I didn't get a chance to iron out and didn't want to dive deep into it yet. If/when we revisit this, I don't think it'll be an issue to remove the `tabindex=-1` from the video element at the time (or even leave it there).

PR #314, opened 2022-09-22
Demuxed 2022 theme

wietsehage:
The goal of this pull request is to get the Demuxed 2022 theme included as an example. This PR contains the following:

- MediaTheme for ‘Demuxed 2022’ including custom inline SVG icons.
- New ‘Demuxed 2022’ entry under examples on index page
- Example page for ‘Demuxed 2022’ theme.


dylanjha:
I know this is early, so I just took a very light first pass on the functionality.

First off. This is a great start, visually. Some points of feedback I came across. I'm guessing some of this you already know / is on the radar, but figured I would mention it all anyway.

1. Can we name this name this `media-theme-demuxed-2022` (We plan on having a different theme every year)
2. `media-captions-button` should only show if the underlying media source has a text track. I forget exactly the best way to detect this, @luwes do you know?
3. We will need a way to specify the live UI player or the on demand UI player. Probably through a stream `stream-type` attribute on the theme either `"live"` or `"on-demand"`. Live is the same as VOD expect: live has fewer controls. No need for playback rate, no `<media-time-display>`, no `<media-time-range>`
4. We will need a Chromecast control w/ `<media-cast-button>` (also requires [castable-video](https://github.com/muxinc/castable-video)), we can go with the default icon for now, will get a proper icon later if necessary
5. Will also need Airplay control `<media-airplay-button>`

When adding controls for chromecast, airplay, fullscreen, pip, you should see that `<media-controller>` has attributes that reflect the state of weather the underlying media source supports the control. For example:

- Simply adding `<media-airplay-button>` will render the airplay button, regardless of weather or not the browser or underlying media supports
- `<media-controller>` will set an attribute on itself `media-airplay-unavailable` after doing feature detection
- It is therefore the responsibility of the theme to write some CSS to handle that state (A theme author may decide they want to show the button as disabled, but in Mux Player we choose to hide it completely, and I think we should do the same for this Demuxed theme). You'd have some CSS like:

```css
media-airplay-button[media-airplay-unavailable] {
  display: none
}
```

We do exactly [this in Mux Player](https://github.com/muxinc/elements/blob/1271fb4149a4749adab261dc84533a05d12a7348/packages/mux-player/src/media-theme-mux/styles.css#L72-L74)

This is the pattern we expect theme authors to follow. It makes it easy for development because you can easily show / hide things with CSS

I see we have a settings cog here, but Media Chrome currently has no concept of a overlay/flyout menu. (Some early work done here #282 on lists). It might be best to avoid the overlay/flyout menu replace it with inline controls for playback rate and captions. The alternative would be to have this theme implement it's own version of an overly/flyout.

src/js/themes/media-theme-demuxed.js
```diff
@@ -0,0 +1,296 @@
...
+const template = `
+<style>
+  :host {
+    --primary-color: black;
+    --secondary-color: transparent;
+    --tertiary-color: #7596CC;
+  }
+
+  video {
+    object-fit: initial;
```
heff, line 22: Is this needed? Feels like it shouldn't be.

src/js/themes/media-theme-demuxed.js
```diff
@@ -0,0 +1,296 @@
...
+
+  media-control-bar :first-child {
+    margin: 0 5px 0 0;
+  }
+  
+  media-control-bar :last-child {
+    margin: 0 0 0 5px;
+  }
+
+  media-play-button, media-mute-button, media-seek-forward-button, media-captions-button, media-pip-button, media-fullscreen-button, media-settings-button, media-cinema-button {
```
heff, line 59: In case it's helpful, since this is part of a theme you could add a class to all of these elements to make this simpler.

src/js/themes/media-theme-demuxed.js
wietsehage: This fixes a difference that shows up between the poster image and the first frame of a video on Safari 15.4+. For the specific case of Demuxed Live this shouldn't be a problem, so I've removed this attribute for now.

src/js/themes/media-theme-demuxed.js
wietsehage: Thanks for this suggestion, makes total sense and cleans things up nicely.

wietsehage:
I've changed the following:

- [ ] Renamed theme to Demuxed 2022
- [ ] Add class .small-button to target all small buttons at once
- [ ] Both Airplay and Cast buttons will show when functionality is available

davidvanleeuwen:
- [x] We've got the responsive view going using the [`resizeObserver`](https://media-chrome-docs.vercel.app/en/responsive-controls#observe-player-resizing):
image
image

- [x] You can switch between `< ... stream-type="on-demand">` and `< ... stream-type="live">`. When setting it to live, it will also cause the media slot to autoplay:
image
image

src/js/themes/media-theme-demuxed-2022.js
```diff
@@ -0,0 +1,566 @@
...
+                C13.9,8.1,13.4,7.9,13.1,8.2z"/>
+            </g>
+            <g>
+              <g>
+                <path d="M16.3,17.4c-0.2,0-0.3-0.1-0.4-0.2c-0.3-0.3-0.3-0.6,0-0.9c0.6-0.6,1-1.4,1-2.3s-0.4-1.6-1-2.3c-0.3-0.3-0.3-0.6,0-0.9
+                  c0.3-0.3,0.6-0.3,0.9,0c0.9,0.8,1.4,2,1.4,3.2s-0.5,2.3-1.4,3.2C16.6,17.4,16.4,17.4,16.3,17.4z"/>
+              </g>
+            </g>
+          </g>
+        </svg>
```
luwes, line 350: these SVG's look like they can be optimized. have you tried running them through a tool like SVGO?

dylanjha:
@wietsehage feel free to address any comments convert this to not-a-draft so we can get it merged

src/js/themes/media-theme-demuxed-2022.js
wietsehage: Great suggestion. I've just run all icons through SVGO and it saves quite some bytes (40% on average).

luwes (approved):
Beautiful! let's merge it in!

I might have a follow PR soon to change the base class, we did some updates to the theming functionality.

PR #318, opened 2022-09-23
feat: remove minify/srcmap esm/cjs, add esm-module

luwes:
- remove minify and sourcemap options from the esm and cjs bundles. these are all separate files (not bundled) so it makes less sense to minify and add sourcemaps for each file. these files are expected to be bundled by the consumer.
- adds a esm-module bundle, was mainly for getting a accurate gzip bundle size but just thought the iife could also be used for that?
- put the comments in expressions so these can be easily removed by bundlers. golfed down -1.3kB with this 🥳 

SCR-20220922-u5h

SCR-20220923-d1s

### DIFF
--- a/src/js/media-chrome-button.js
+++ b/src/js/media-chrome-button.js
@@ -19,7 +19,7 @@ template.innerHTML = `
 
     padding: var(--media-control-padding, 10px);
 
-    /* Vertically center any text */
+    ${/* Vertically center any text */''}
     font-size: 14px;
     line-height: var(--media-text-content-height, var(--media-control-height, 24px));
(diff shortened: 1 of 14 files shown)


src/js/media-chrome-button.js
```diff
@@ -32,17 +32,17 @@ template.innerHTML = `
     font-family: Arial, sans-serif;
   }
 
-  /*
+  ${/*
```
gkatsev, line 35: this feels weird but it does work

gkatsev (approved):
I've been wanting unminified media-chrome for a while 👍🏻

PR #320, opened 2022-09-23
feat: support being able to disable buttons

gkatsev:
The final piece I believe is to turn off thumbnails of the time-range when it's disabled. But wanted to get some 👀 to make sure that this direction is fine before making that change.

Another thing that I haven't been able to figure out is if I set up the following css:
```css
[disabled] {
  cursor: not-allowed;
}
```
When hovering over the `input` element itself, it would still show the hand cursor. Though, it will not do anything. I tried even adding a part on it and added `[disabled]::part(range)` to the selector above, and it still shows the hand cursor.

### DIFF
--- a/src/js/media-chrome-button.js
+++ b/src/js/media-chrome-button.js
@@ -94,40 +94,40 @@ class MediaChromeButton extends window.HTMLElement {
     this.enable();
   }
 
-  #clickHandler = (e) => {
+  #clickListener = (e) => {
     this.handleClick(e);
   }
 
   // NOTE: There are definitely some "false positive" cases with multi-key pressing,
   // but this should be good enough for most use cases.
-  #keyupHandler = (e) => {
+  #keyupListener = (e) => {
     const { key } = e;
     if (!this.keysUsed.includes(key)) {
-      this.removeEventListener('keyup', keyupHandler);
+      this.removeEventListener('keyup', keyupListener);
       return;
     }
 
     this.handleClick(e);
   }
 
-  #keydownHandler = (e) => {
+  #keydownListener = (e) => {
     const { metaKey, altKey, key } = e;
     if (metaKey || altKey || !this.keysUsed.includes(key)) {
-      this.removeEventListener('keyup', this.#keyupHandler);
+      this.removeEventListener('keyup', this.#keyupListener);
       return;
     }
-    this.addEventListener('keyup', this.#keyupHandler, {once: true});
+    this.addEventListener('keyup', this.#keyupListener, {once: true});
   }
 
   enable() {
-    this.addEventListener('click', this.#clickHandler);
-    this.addEventListener('keydown', this.#keydownHandler);
+    this.addEventListener('click', this.#clickListener);
+    this.addEventListener('keydown', this.#keydownListener);
     this.setAttribute('tabindex', 0);
   }
 
   disable() {
-    this.removeEventListener('click', this.#clickHandler);
-    this.removeEventListener('keyup', this.#keyupHandler);
+    this.removeEventListener('click', this.#clickListener);
+    this.removeEventListener('keyup', this.#keyupListener);
     this.removeAttribute('tabindex');
   }
 
--- a/src/js/media-chrome-button.js
+++ b/src/js/media-chrome-button.js
@@ -91,30 +91,42 @@ class MediaChromeButton extends window.HTMLElement {
 
     shadow.appendChild(buttonHTML);
 
-    this.addEventListener('click', (e) => {
-      this.handleClick(e);
-    });
+    this.enable();
+  }
 
-    // NOTE: There are definitely some "false positive" cases with multi-key pressing,
-    // but this should be good enough for most use cases.
-    const keyUpHandler = (e) => {
-      const { key } = e;
-      if (!this.keysUsed.includes(key)) {
-        this.removeEventListener('keyup', keyUpHandler);
-        return;
-      }
+  #clickHandler = (e) => {
+    this.handleClick(e);
+  }
 
-      this.handleClick(e);
-    };
+  // NOTE: There are definitely some "false positive" cases with multi-key pressing,
+  // but this should be good enough for most use cases.
+  #keyUpHandler = (e) => {
+    const { key } = e;
+    if (!this.keysUsed.includes(key)) {
+      this.removeEventListener('keyup', keyUpHandler);
+      return;
+    }
 
-    this.addEventListener('keydown', (e) => {
-      const { metaKey, altKey, key } = e;
-      if (metaKey || altKey || !this.keysUsed.includes(key)) {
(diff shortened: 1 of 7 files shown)


src/js/media-chrome-button.js
```diff
@@ -91,30 +91,44 @@ class MediaChromeButton extends window.HTMLElement {
...
+  enable() {
+    this.addEventListener('click', this.#clickHandler);
+    this.addEventListener('keydown', this.#keydownHandler);
+    this.setAttribute('tabindex', 0);
+  }
+
+  disable() {
+    this.removeEventListener('click', this.#clickHandler);
+    this.removeEventListener('keyup', this.#keyUpHandler);
+    this.removeAttribute('tabindex');
```
gkatsev, line 131: Removing `tabindex` makes our buttons not focusable when they are disabled.

src/js/media-chrome-range.js
```diff
@@ -213,6 +216,16 @@ class MediaChromeRange extends window.HTMLElement {
...
       }
+    } else if (
+      attrName === 'disabled' ||
+      attrName === 'aria-disabled' &&
+      oldValue !== newValue
+    ) {
+      if (newValue == null) {
+        this.range.removeAttribute(attrName);
+      } else {
+        this.range.setAttribute(attrName, newValue);
```
gkatsev, line 227: we want to copy over the `disabled` attributes onto the `input` to make sure that it gets disabled correctly.

luwes (approved):
👍

src/js/media-time-range.js
```diff
@@ -313,77 +327,84 @@ class MediaTimeRange extends MediaChromeRange {
...
+    );
+    if (!this.#rangeEntered && mediaDurationStr) {
+      this.#rangeEntered = true;
+      this.#trackMouse();
+
+      window.addEventListener('pointermove', this.#offRangeHandler, false);
+    }
+  }
+
+  _enableBoxes() {
```
luwes, line 399: feel free to change these to private class methods too.
at the time I didn't know we were adopting these in Media Chrome.

src/js/media-time-range.js
```diff
@@ -313,77 +327,84 @@ class MediaTimeRange extends MediaChromeRange {
...
+      !evt.composedPath().includes(this) ||
+      [...this.#boxes].some((b) => evt.composedPath().includes(b))
+    ) {
+      window.removeEventListener('pointermove', this.#offRangeHandler);
+      this.#rangeEntered = false;
+      this.#stopTrackingMouse();
+    }
+  }
+
+  #trackMouse = () => {
```
luwes, line 374: is this worth a separate method?

luwes (approved):
LGTM tested locally

src/js/media-time-range.js
gkatsev: I guess worth input from others. I'm ok with either, since esbuild does compile them away.

gkatsev: The other private props are more important to have the event handler methods be per instance to avoid wardresses around removing the event handler and all buttons stopping working.

luwes: it's fine I think, support is pretty good https://caniuse.com/mdn-javascript_classes_private_class_fields

gkatsev: yeah, we could consider having esbuild not copmile it away.

src/js/media-chrome-button.js
```diff
@@ -91,30 +91,44 @@ class MediaChromeButton extends window.HTMLElement {
...
-        this.removeEventListener('keyup', keyUpHandler);
-        return;
-      }
+    this.enable();
+  }
 
-      this.handleClick(e);
-    };
+  #clickHandler = (e) => {
+    this.handleClick(e);
```
heff, line 98: Two functions named basically the same thing isn't great. Maybe the addition of it being private is enough.

Now I'm wondering if it'd be better to have the child class set the click handler via `super()` instead of `.handleClick`.

src/js/media-chrome-button.js
```diff
@@ -91,30 +91,44 @@ class MediaChromeButton extends window.HTMLElement {
...
-
-    // NOTE: There are definitely some "false positive" cases with multi-key pressing,
-    // but this should be good enough for most use cases.
-    const keyUpHandler = (e) => {
-      const { key } = e;
-      if (!this.keysUsed.includes(key)) {
-        this.removeEventListener('keyup', keyUpHandler);
-        return;
-      }
+    this.enable();
```
heff, line 94: Just confirming, we don't need to check if disabled already here, yeah?

src/js/media-chrome-range.js
```diff
@@ -190,7 +190,10 @@ template.innerHTML = `
 
 class MediaChromeRange extends window.HTMLElement {
   static get observedAttributes() {
-    return [MediaUIAttributes.MEDIA_CONTROLLER];
+    return [
+      'disabled',
+      'aria-disabled',
```
heff, line 195: This is needed for `range` but not `button`?

src/js/media-time-range.js
```diff
@@ -313,77 +327,84 @@ class MediaTimeRange extends MediaChromeRange {
     style.transform = `translateX(${boxPos}px)`;
   }
 
-  _enableBoxes() {
```
heff, line 316: It's hard to follow the diff after this point. Was this a simple moving around of code or was there a refactor happening?

heff (approved):
Looks good to me. Did just using `pointer-events: none` turn out to not be a solution here?

When finishing this up we should add the disabled state to the example files.

gkatsev:
> Did just using `pointer-events: none` turn out to not be a solution here?

Yeah, `pointer-events: none` isn't enough here. First, we actually do want pointer interactions, at least partially. Since, we generally want to have `cursor: not-allowed` on disabled controls. Second, even if we set it and pointer events aren't allowed, keyboard and screen reader users would still be able to focus the element and interact with it which we don't want.

> When finishing this up we should add the disabled state to the example files.

Good call, I'll add something for that.

src/js/media-time-range.js
gkatsev: yes, this mostly moves the handlers out of `_enableBoxes()` into private properties so that a `_disableBoxes()` method could be created.

src/js/media-chrome-button.js
gkatsev: renamed to `#clickListener` along with `keyupListener` and `keydownListener`, for consistency. Hopefully, makes it clear enough that there's a difference.

heff (approved):

cjpillsbury (approved):
I think this is a reasonable initial implementation. Let's make sure we follow up on the Screen Reader/A11Y concerns/considerations/discussion we had out of band so it doesn't fall off completely. (tl;dr - Screen Reader-specific navigation vs. standard keyboard navigation and expectations/good experiences for folks relying on those)

PR #327, opened 2022-10-06
feat: add poster object fit and position css vars

luwes:
Related to https://github.com/muxinc/elements/issues/444

adds positioning and size / fit css vars for the `media-poster-image` element

### DIFF
--- a/src/js/media-poster-image.js
+++ b/src/js/media-poster-image.js
@@ -13,9 +13,6 @@ template.innerHTML = `
       pointer-events: none;
       display: inline-block;
       box-sizing: border-box;
-
-      width: auto;
-      height: auto;
     }
 
     img {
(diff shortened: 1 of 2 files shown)


src/js/media-poster-image.js
```diff
@@ -13,20 +13,18 @@ template.innerHTML = `
...
       min-width: 100%;
       min-height: 100%;
-      background-size: contain;
-      background-position: center;
       background-repeat: no-repeat;
-      object-fit: contain;
+      background-position: var(--media-background-position, var(--media-object-position, center));
+      background-size: var(--media-background-size, var(--media-object-fit, contain));
+      object-fit: var(--media-object-fit, contain);
+      object-position: var(--media-object-position, center);
```
gkatsev, line 27: default object position is `50% 50%`, not sure if there's a difference between that and center, though.

gkatsev: yeah, can't tell a difference between them, so, probably equivalent.

gkatsev (approved):

src/js/media-poster-image.js
luwes: yup it's same, checked it too

PR #330, opened 2022-10-09
feat: expose types

gkatsev:
Use tsc to automatically generated types from the JS and JSDoc comments.
Mostly minimal changes done, and it actually caught a few issues that I'll pick out of here into a separate PR, probably.
Some things are `@ts-ignore` because it's not simple fixes since I don't think you can cast values using jsdoc comments.


tsconfig.json
```diff
@@ -0,0 +1,107 @@
+{
```
gkatsev, line 1: this file can probably be trimmed. I used `tsc init` to create it.

src/js/media-chrome-button.js
```diff
@@ -103,7 +106,7 @@ class MediaChromeButton extends window.HTMLElement {
   #keyupListener = (e) => {
     const { key } = e;
     if (!this.keysUsed.includes(key)) {
-      this.removeEventListener('keyup', keyupListener);
+      this.removeEventListener('keyup', this.#keyupListener);
```
gkatsev, line 109: this is one of the fixes that we need in `main`

src/js/media-controller.js
```diff
@@ -35,6 +35,7 @@ import {
 
 const ButtonPressedKeys = ['ArrowLeft', 'ArrowRight', 'Enter', ' ', 'f', 'm', 'k', 'c'];
 const DEFAULT_SEEK_OFFSET = 10;
+const DEFAULT_TIME = 0;
```
gkatsev, line 38: this is one of the fixes that we need in `main`

src/js/media-controller.js
```diff
@@ -1284,7 +1287,34 @@ const airplaySupported = !!window.WebKitPlaybackTargetAvailabilityEvent;
 const castSupported = !!window.chrome;
 const fullscreenEnabled = document[fullscreenApi.enabled];
 
-function serializeTimeRanges(timeRanges = []) {
+/** @type {TimeRanges} */
+const emptyTimeRanges = Object.freeze({
```
gkatsev, line 1291: copied from elements, thanks @cjpillsbury.
We don't need it in main, though, it won't hurt either.

src/js/media-seek-backward-button.js
```diff
@@ -8,7 +8,7 @@ import { MediaUIEvents, MediaUIAttributes } from './constants.js';
 import { verbs } from './labels/labels.js';
 import { getSlotted, updateIconText } from './utils/element-utils.js';
 
-const DEFAULT_SEEK_OFFSET = 30;
+const DEFAULT_SEEK_OFFSET = '30';
```
gkatsev, line 11: we only use this in strings, tsc will complain otherwise.

src/js/extras/media-clip-selector/index.js
```diff
@@ -468,6 +487,10 @@ class MediaClipSelector extends window.HTMLElement {
   }
 
   disableThumbnails() {
+    /** @type {HTMLElement} */
+    const thumbnailContainer = this.shadowRoot.querySelector(
```
gkatsev, line 491: this is another fix that should go into main, quickly.

src/js/extras/media-clip-selector/index.js
```diff
@@ -19,6 +19,7 @@ function lockBetweenZeroAndOne(num) {
   return Math.max(0, Math.min(1, num));
 }
 
+// @ts-ignore
```
heff, line 22: I wouldn't expect to have to address all these native properties like this. Wondering if that's gonna get annoying.

heff (commented):
This has conflicts now, sorry 😞 . Want to try to get this in before the next merge? Controller changes will be the most complicated.

Looks good to me otherwise. How do others feel about this approach? I'm assuming this makes it way easier to use media chrome with typescript?

src/js/extras/media-clip-selector/index.js
gkatsev: another option is to turn off `checkJs` in tsc so that it doesn't spew any errors. It's possible there's a way to handling this, but not sure. If I remember correctly, adding `@type {HTMLTemplateElement}` didn't fix it like for the other things.

gkatsev: Also, there may be a way of fixing this, but unfortunately, I haven't figured it out yet. It basically comes down to not being able to cast vars like you can in native typescript using the jsdoc syntax.

heff: Cool. Not a blocker either way.

gkatsev: Apparently, it does support it https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#casts

However, I think that should be a separate PR not to delay this one much longer.

tsconfig.json
luwes: yes, can the comments be removed here?

luwes (changes requested):

luwes (approved):

PR #331, opened 2022-10-10
fix: a couple of fixes

gkatsev:
When creating https://github.com/muxinc/media-chrome/pull/330, it pointed out a few issues that should be fixed. Since I figured #330 might take might be a bit before getting merged, I'd create a separate PR for the specific fixes that should go in.


luwes (approved):
👍

PR #332, opened 2022-10-10
fix: properly check iphones for fullscreen unavailability

gkatsev:


src/js/media-controller.js
```diff
@@ -1283,7 +1283,18 @@ const volumeSupportPromise = hasVolumeSupportAsync().then((supported) => {
 
 const airplaySupported = !!window.WebKitPlaybackTargetAvailabilityEvent;
 const castSupported = !!window.chrome;
-const fullscreenEnabled = document[fullscreenApi.enabled];
+
+export const hasFullscreenSupport = (mediaEl = getTestMediaEl()) => {
+  let fullscreenEnabled = document[fullscreenApi.enabled];
+
+  if (!fullscreenEnabled) {
+    fullscreenEnabled = 'webkitDisplayingFullscreen' in mediaEl;
```
luwes, line 1291: is this the right way?
not https://developer.apple.com/documentation/webkitjs/htmlvideoelement/1628805-webkitsupportsfullscreen

luwes (approved):

gkatsev:
As a followup to this, we should have better handling for iPhone's fullscreen and PiP. For example webkitSupportsFullscreen can be false if the video hasn't loaded or for other reasons.

PR #337, opened 2022-10-12
feat: fade poster over placeholder on load

decepulis:
Just a first pass. Lemme know what you think!


dylanjha:
Wow, this is pretty dope

https://user-images.githubusercontent.com/764988/195941174-2cbd73ed-aef0-4472-ab85-bef419d56209.mov

src/js/media-poster-image.js
```diff
@@ -41,10 +53,19 @@ const unsetBackgroundImage = (el) => {
...
+  el.classList.remove('fade');
+}
+const loadFade = (el) => {
+  el.classList.add('loaded');
+}
 
 class MediaPosterImage extends window.HTMLElement {
   static get observedAttributes() {
-    return [MediaUIAttributes.MEDIA_HAS_PLAYED, 'placeholder-src', 'src'];
+    return [MediaUIAttributes.MEDIA_HAS_PLAYED, 'no-fade', 'placeholder-src', 'src'];
```
dylanjha, line 68: Is `no-fade` a new attribute? Assuming we go this direction, I don't *think* we need it, honestly

src/js/media-poster-image.js
```diff
@@ -53,7 +74,10 @@ class MediaPosterImage extends window.HTMLElement {
     this.attachShadow({ mode: 'open' });
     this.shadowRoot.appendChild(template.content.cloneNode(true));
 
+    this.container = this.shadowRoot.querySelector('#container');
     this.image = this.shadowRoot.querySelector('#image');
+
+    this.image.addEventListener('load', () => loadFade(this.image));
```
dylanjha, line 80: @cjpillsbury would probably know the answer here (I genuinely don't), do we need to remove event listeners like this that get attached in the constructor

gkatsev: yeah, they should get removed. Though, also can use `{once: true}` option to auto-remove it when `load` triggers.

src/js/media-poster-image.js
```diff
@@ -63,13 +87,23 @@ class MediaPosterImage extends window.HTMLElement {
       } else {
         this.image.setAttribute('src', newValue);
       }
-    }
+    }
```
gkatsev, line 90: nit: whitespace addition

src/js/media-poster-image.js
```diff
@@ -15,24 +15,36 @@ template.innerHTML = `
...
       object-fit: var(--media-object-fit, contain);
       object-position: var(--media-object-position, center);
+      opacity: 1;
+    }
+    img.fade {
+      opacity: 0;
+      transition: opacity 0.15s ease;
+    }
+    img.fade.loaded {
+      opacity: 1;
```
gkatsev, line 37: Can switch to use `img:not([no-fade])` instead and get rid of everything related to the `fade` class to simplify the code and the new additional JS.

src/js/media-poster-image.js
```diff
@@ -15,24 +15,36 @@ template.innerHTML = `
...
     }
 
     :host([${MediaUIAttributes.MEDIA_HAS_PLAYED}]) img {
       display: none;
     }
   </style>
 
-  <img aria-hidden="true" id="image"/>
+  <div aria-hidden="true" id="container">
```
gkatsev, line 45: Potentially, the host itself i.e. `this` below, can be used for the background image so we may not need this new container

src/js/media-poster-image.js
```diff
@@ -63,13 +87,23 @@ class MediaPosterImage extends window.HTMLElement {
...
 
     if (attrName === 'placeholder-src') {
       if (newValue == null) {
-        unsetBackgroundImage(this.image);
+        unsetBackgroundImage(this.container);
+      } else {
+        setBackgroundImage(this.container, newValue);
+      }
+    }
+    if (
```
gkatsev, line 99: if we use the attribute selector, this if shouldn't be needed.

src/js/media-poster-image.js
gkatsev, line 80: Christian was wondering whether this is necessary but I think it does, especially since I think we should expose the `loading` property in `media-poster-image` https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-loading

src/js/media-poster-image.js
gkatsev: Additionally, should we fade in only if a placeholder is provided? I'd probably expect it to show up asap unless placeholder (unless/when we expose the `loading` attribute)

src/js/media-poster-image.js
cjpillsbury: Can't use `{ once: true }` since it's valid to change the `media-poster-image` src.

src/js/media-poster-image.js
luwes: I agree, I'd prefer no added attribute but maybe possible to use a CSS var to override this fading functionality.

some users might prefer to see the image as soon as anything is loaded like if they use progressive jpeg format
https://medium.com/hd-pro/jpeg-formats-progressive-vs-baseline-73b3938c2339

src/js/media-poster-image.js
gkatsev: ah, that's true. I guess just need to make sure it's removed on disconnectedCallback then.

src/js/media-poster-image.js
gkatsev: maybe if we expose `loading=lazy/eager` as on the `img` element, we can use that property to decide whether to fade or not. Where, with lazy, you probably want to fade, but with eager, you probably don't.

luwes:
closing as stale

PR #338, opened 2022-10-14
fix: clicking in media-container should schedule inactive

### LINKED ISSUE #188: user inactivity tracking doesn't account for mouse clicks

gkatsev:
If a user presses play but doesn't move the mouse out of the control bounds, the controls never autohide. Instead, we should schedule inactive when a click happens. Then, on pointermove automatic hiding would be prevented as usual.

Mobile behavior of toggling controls is maintained as well.

Fixes #188

### DIFF
--- a/src/js/media-container.js
+++ b/src/js/media-container.js
@@ -359,6 +359,8 @@ class MediaContainer extends window.HTMLElement {
         } else {
           scheduleInactive();
         }
+      } else {
+        scheduleInactive();
       }
     });
 
--- a/src/js/media-container.js
+++ b/src/js/media-container.js
@@ -359,7 +359,7 @@ class MediaContainer extends window.HTMLElement {
         } else {
           scheduleInactive();
         }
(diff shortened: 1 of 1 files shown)


gkatsev:
We should consider whether any click on any control should trigger this behavior. It might only be wanted on a play toggle or a fullscreen toggle 🤔

gkatsev:
Updated to limit to media-play-button and media-fullscreen-button.

cjpillsbury (approved):
👌 good for quick current solution. As a followup (maybe a fast followup), we should solve this via media state change requests (`mediaplayrequest` & `mediaenterfullscreenrequest`)

PR #342, opened 2022-10-18
fix: preview bounds in shadow dom

luwes:
this fixes an issue when configuring a bounds for the previews in shadow dom

### DIFF
--- a/src/js/media-time-range.js
+++ b/src/js/media-time-range.js
@@ -7,7 +7,7 @@ import {
 import { MediaUIEvents, MediaUIAttributes } from './constants.js';
 import { nouns } from './labels/labels.js';
 import { formatAsTimePhrase } from './utils/time.js';
-import { getOrInsertCSSRule } from './utils/element-utils.js';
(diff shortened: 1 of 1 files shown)


src/js/media-time-range.js
```diff
@@ -428,7 +428,7 @@ function getBoxPosition(el, box, percent) {
   // Get the element that enforces the bounding box for the hover preview.
   const mediaBounds =
     (el.getAttribute('media-bounds')
-      ? document.getElementById(el.getAttribute('media-bounds'))
+      ? closestComposedNode(el, `#${el.getAttribute('media-bounds')}`)
```
gkatsev, line 431: this is because the `media-bounds` references element could be outside the `media-time-range`'s shadow dom?

luwes: yes, this fixes the case where it was in the shadow dom but `document.getElementById` doesn't have access to it
using closesComposedNode it can get all the ancestors

gkatsev (approved):

Issue #281, opened 2022-08-07
Time range Fixes & Improvements


luwes:
Wants and needs I ran into developing 
repo: https://github.com/luwes/media-theme-neo 
demo: https://media-theme-neo.vercel.app/?current-time=21.3

### Fixes:

- The current time box's position is not always correct on player load and on player resize, it should reposition on resize

![SCR-20220807-ji0](https://user-images.githubusercontent.com/360826/183306994-b4d6dee2-623a-4d4b-a465-65047e5eb593.png)

- When dragging on the time range the current time box doesn't move along smoothly, might have to do debounce here because it looks like the attribute is updated very quickly and render can't keep up -> missing frames.

- Setting a 1px solid border around the range input is glitchy because it's also set on the .background div, creates a 2 px border.

- The `media-bounds` attribute on the `media-time-range` element should not query an element via `document.getElementById(id)`. it should be via `this.getRootNode().querySelector(`#${id}`) because the parent defining the bounds is probably in a shadow DOM

### Improvements

- Being able to set a left and right padding for the current and preview boxes 

![SCR-20220807-jwv](https://user-images.githubusercontent.com/360826/183307520-a8832412-e00f-4774-8c68-07d7179561d6.png)

- Another cool thing would be to have the current time code show the duration for the video if the clip has not started yet and when playing starts it turns into the current time. also realized that `media-current-time-display` is not much different than `media-time-display`, maybe `media-current-time-display` can be removed. add the `media-has-played` attribute and this is all possible with `media-time-display` and CSS

luwes:
many of these were fixed in #345 #342

PR #345, opened 2022-10-20
fix: box percentage positioning & time range border glitch

luwes:
Related #281

- fixes a `--media-range-track-border` double border bug on the chrome range
- adds a `--media-range-track-outline` and `--media-range-track-outline-offset`
- fixes the time range current time box using percentages which repositions on player resize 
  that was a tricky one but it makes the code simpler overal though introduces 1 wrapper div#container

### DIFF
--- a/src/js/media-chrome-range.js
+++ b/src/js/media-chrome-range.js
@@ -46,6 +46,7 @@ template.innerHTML = `
       --media-range-padding: var(--media-control-padding, 10px);
       --_media-range-padding: var(--media-range-padding, 10px);
 
+      vertical-align: middle;
       box-sizing: border-box;
       display: inline-block;
       position: relative;
--- a/src/js/media-chrome-range.js
+++ b/src/js/media-chrome-range.js
@@ -28,6 +28,8 @@ const trackStyles = `
   min-width: 40px;
   height: var(--track-height);
   border: var(--media-range-track-border, none);
+  outline: var(--media-range-track-outline);
+  outline-offset: var(--media-range-track-outline-offset);
   border-radius: var(--media-range-track-border-radius, 0);
   background: var(--media-range-track-progress-internal, var(--media-range-track-background, #eee));
   box-shadow: var(--media-range-track-box-shadow, none);
@@ -115,7 +117,6 @@ template.innerHTML = `
       width: var(--media-range-track-width, 100%);
       height: var(--track-height);
       border-radius: var(--media-range-track-border-radius, 0);
-      box-sizing: border-box;
       position: absolute;
       top: 50%;
       transform: translate(var(--media-range-track-translate-x, 0px), calc(var(--media-range-track-translate-y, 0px) - 50%));
@@ -141,7 +142,6 @@ template.innerHTML = `
       ${/* Add z-index so it overlaps the top of the control buttons if they are right under. */''}
       z-index: 1;
       display: var(--media-time-range-hover-display, none);
-      box-sizing: border-box;
       position: absolute;
       width: 100%;
       bottom: var(--media-time-range-hover-bottom, -5px);
@@ -151,7 +151,6 @@ template.innerHTML = `
     #range {
       z-index: 2;
       position: relative;
-      height: var(--media-range-track-height, 4px);
     }
 
     ${/*
--- a/src/js/media-chrome-range.js
+++ b/src/js/media-chrome-range.js
@@ -151,6 +151,7 @@ template.innerHTML = `
     #range {
       z-index: 2;
       position: relative;
+      height: var(--media-range-track-height, 4px);
     }
 
     ${/*
--- a/src/js/media-chrome-range.js
+++ b/src/js/media-chrome-range.js
@@ -44,12 +44,9 @@ template.innerHTML = `
       --media-range-padding: var(--media-control-padding, 10px);
       --_media-range-padding: var(--media-range-padding, 10px);
 
-      position: relative;
-      display: inline-flex;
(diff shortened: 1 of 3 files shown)


src/js/media-chrome-range.js
```diff
@@ -182,10 +183,12 @@ template.innerHTML = `
       background-color: #777;
     }
   </style>
-  <div id="background"></div>
-  <div id="pointer"></div>
-  <div id="hoverzone"></div>
-  <input id="range" type="range" min="0" max="1000" step="any" value="0">
+  <div id="container">
```
luwes, line 186: added a container div here which makes it easier to take in account the padding that is added on media-time-range
#background, #pointer, etc which are absolutely positioned don't have to take in to account that padding anymore

src/js/media-time-range.js
```diff
@@ -34,13 +34,20 @@ template.innerHTML = `
       color: #fff;
     }
 
-    [part~="box"] {
-      display: flex;
-      flex-direction: column;
-      align-items: center;
+    #preview-rail,
+    #current-rail {
+      width: 100%;
```
luwes, line 39: this is a key part the "rails" are expanded to 100% width which equals the width of the parent and the input range.
then we set CSS transform translateX in percentages which are percentages based on the element itself not the parent. transforms are weird that way

src/js/media-time-range.js
```diff
@@ -408,37 +415,35 @@ class MediaTimeRange extends MediaChromeRange {
...
-function getBoxPosition(el, box, percent) {
-  const rect = el.getBoundingClientRect();
-  const rangeRect = el.range.getBoundingClientRect();
+function getBoxPosition(el, box, ratio) {
+  let position = `${ratio * 100}%`;
+
+  // Get the element that enforces the bounds for the time range boxes.
+  const bounds =
+    (el.getAttribute('media-bounds')
+      ? closestComposedNode(el, `#${el.getAttribute('media-bounds')}`)
```
luwes, line 424: should we rename this attribute `media-bounds`? 
I remember @heff saying `media-` prefixed attrs are propagated. this one is not.

open to suggestions.
(I don't think anyone has discovered this functionality yet)

src/js/media-time-range.js
```diff
@@ -408,37 +415,35 @@ class MediaTimeRange extends MediaChromeRange {
...
 }
 
-function getBoxPosition(el, box, percent) {
-  const rect = el.getBoundingClientRect();
-  const rangeRect = el.range.getBoundingClientRect();
+function getBoxPosition(el, box, ratio) {
+  let position = `${ratio * 100}%`;
+
+  // Get the element that enforces the bounds for the time range boxes.
+  const bounds =
```
gkatsev, line 422: it's probably worth moving this back down, if we return early on line 436 `if !boxWidth`, we won't need to do the work for these bounds.

src/js/media-time-range.js
gkatsev: Where is media-bounds coming from?

gkatsev (commented):
For whatever reason, this completely breaks thumbnails on firefox for me.

src/js/media-time-range.js
luwes: 👍 will do

src/js/media-time-range.js
luwes: it's used when the media-time-range has a left offset for example
https://media-theme-neo.vercel.app/

the attribute should be the `id` of the container that act as the bounding box for the time range boxes (preview thumb)

src/js/media-time-range.js
luwes: @gkatsev which example? 
on my end it seems to work.

gkatsev: @luwes with https://media-chrome-git-fork-luwes-fix-range-bugs-mux.vercel.app/examples/advanced.html, when hovering over the progress bar I get a bunch of these errors in the console:
```
Uncaught DOMException: CSSStyleSheet.cssRules getter: Not allowed to access cross-origin stylesheet
```
from `getOrInsertCSSRule` in `element-utils.js`
However, it works fine on https://media-chrome-mux.vercel.app/examples/advanced.html

luwes: what's your Firefox version? gonna try to repro. strange because I'm not doing much different than before.
maybe 1 thing is that 1 of the style tags is now in the container element instead of the shadowRoot

gkatsev: Seeing the same behavior in my Firefox for Windows, but a quick test in Browserstack with Firefox 105 and 106 (Beta) shows it working as expected 🤔

gkatsev: ah, I do have Dark Reader enabled, let me check. Though, it's weird that it's working in main but not with this fork

luwes: true yes weird, let me try install that too

gkatsev: @luwes that was it. I guess I had dark reader disabled on main, but not on your fork preview.

gkatsev: So, yeah, not a blocker for this PR. Probably worth investigating but as a low priority thing.

luwes: will add in a try / catch to workaround this issue

src/js/media-time-range.js
gkatsev: ah, I see. It's a bit like the thing we talked about for the fullscreen button/media-controller thing.

cjpillsbury (approved):

PR #349, opened 2022-10-21
docs: tweak docs themes ala media-chrome.org

luwes:
another pass on the docs theme to make it more look like https://www.media-chrome.org/
also made the header sticky and translucent.

SCR-20221021-jvj

SCR-20221021-jvt


gkatsev (approved):
Looks good, the only potential addition is updating the hover color on the left bar from that greyish color to some purple-ish color.

cjpillsbury (approved):
💅

luwes:
@gkatsev good call, will update that together with the new logo PR.

PR #351, opened 2022-10-21
Fix missing clip-selector in package builds

redoPop:
This PR restores `<media-clip-selector>` to package builds. It'd fallen out because the esbuild entry points [`src/js/*.js src/js/*/*.js`](https://github.com/muxinc/media-chrome/blob/e0b5fa98378a297042cc957acbc47e67f8acb582/package.json#L16) don't encompass `src/js/extras/media-clip-selector/index.js`.

Per discussion in #344 we'd prefer to use esbuild's command line interface, and extending the entry points to `src/js/*.js src/js/*/*.js src/js/*/*.js` is a simple way to fix the proximate issue.

Per evanw/esbuild#2508, the entry point can be simplified to `src/js/**/*.js` after the next major release of esbuild.


luwes (approved):
LGTM thank you!

gkatsev:
Published in 0.15.0. Unfortunately, it didn't make it into the changelog.

PR #353, opened 2022-10-24
docs: landing page spike

luwes:
ported over the landing page from Next.js based media-chrome.org to Astro


docs/package.json
```diff
@@ -13,24 +13,28 @@
...
+    "@types/react": "^18.0.22",
+    "astro": "1.5.2",
+    "clsx": "^1.2.1",
+    "hastscript": "^7.1.0",
     "html-escaper": "^3.0.3",
-    "media-chrome": "^0.11.1",
-    "preact": "^10.7.3",
-    "react": "^18.1.0",
-    "react-dom": "^18.1.0",
+    "media-chrome": "^0.14.1",
```
gkatsev, line 26: non-blocking: we should figure out how to keep this up-to-date. Maybe we can use yarn's `file:` here to point at the root of this project?
Alternatively, we can do a script as part of a publish that updates this version on publish.

docs/src/styles/theme.css
```diff
@@ -64,17 +64,17 @@
...
-  --theme-accent-opacity: 0.1;
   --theme-divider: hsla(var(--color-gray-95), 1);
   --theme-text: hsla(var(--color-gray-10), 1);
   --theme-text-light: hsla(var(--color-gray-40), 1);
   /* @@@: not used anywhere */
   --theme-text-lighter: hsla(var(--color-gray-80), 1);
   --theme-bg: hsla(var(--color-base-white), 100%, 1);
-  --theme-bg-hover: hsla(var(--color-gray-95), 1);
-  --theme-bg-offset: hsla(var(--color-gray-90), 1);
+  --theme-bg-hover: hsla(var(--color-magenta), var(--theme-accent-opacity));
```
gkatsev, line 76: 👍🏻

docs/package.json
gkatsev: If we can use `file:` then having the individual component examples in the docs site won't be a huge issue because then if we wanted to debug an individual example it would just be all linked together.

gkatsev (commented):
Super quickly skimmed through things, I assume there's a lot of just copy-pasta from the other site.
I'll do a review once vercel is back up and re-deployed the site.

docs/package.json
luwes: good call, will make this change

docs/src/pages/landing/components/media-chrome-list-item-renderer.tsx
```diff
@@ -0,0 +1,119 @@
...
+          return;
+        }
+        deselectItem(value);
+      }}
+      // NOTE: Semi-hacky way of ensuring that focus isn't set to element
+      // (should only be set to list itself) (CJP)
+      onMouseDown={(e) => e.preventDefault()}
+      onClick={() => selectItem(value)}
+      className={clsx("block w-full", {
+        "outline-black": active && showActive && !selected,
```
gkatsev, line 78: This is copied from media-chrome.org, but I noticed that the outline doesn't show up with regular keyboard use. We also want to add `outline` here.
```suggestion
        "outline-black": active && showActive && !selected,
        "outline": active && showActive && !selected,
```

docs/src/pages/landing/components/toggle.tsx
```diff
@@ -0,0 +1,26 @@
+/** @jsxImportSource react */
+const Toggle: React.FC<{
+  id: string;
+  label: string;
+  onSelectionChange?: (selected: boolean) => void;
+}> = ({ id, label, onSelectionChange = (_: boolean) => {} }) => {
+  return (
+    <div>
+      <span className="mr-4">{label}</span>
+      <div className="switch focus-within:outline-black">
```
gkatsev, line 10: we want to show the outline when necessary
```suggestion
      <div className="switch focus-within:outline-black focus-within:outline">
```

gkatsev (approved):
Site looks great!
I noticed that keyboard usage of the components list doesn't have outlines as expected, so, have a suggestion for a quickfix for it. It's present on the existing site, but we may as well get it fixed here.

PR #355, opened 2022-10-25
fix: use composedPath for checking target in schedule inactive event handler

gkatsev:


luwes (approved):
LGTM 
nit: `el?.nodeName?.toLowerCase()` could also use `el?.localName`

cjpillsbury (approved):
👍

gkatsev:
`localName` seems wrong to use since it comes from xml stuff.

PR #362, opened 2022-11-10
feat: Add HTML based Theme w/ minimal template language

luwes:
Related discussion: https://github.com/muxinc/media-chrome/discussions/323
This change adds a `<media-theme>` web component that can render HTML only based themes / templates.

I included a template parts implementation that has support for the `InnerTemplatePart` as explained in the proposal.
https://github.com/dy/template-parts It's also more in line with the newer DOM parts API proposal which makes the parts its own thing, it doesn't require a `TemplateInstance`. 
Conditionals and other constructs like loops could be handled by these inner templates. [Template-Instantiation](https://github.com/WICG/webcomponents/blob/gh-pages/proposals/Template-Instantiation.md#33-conditionals-and-loops-using-nested-templates) 
The power of template parts is shown very well in this library https://github.com/dy/templize by @dy Not that Media Chrome themes should go that far, we should prefer less specialized syntax as possible.

---

Questions for unblocking first iteration. 
Required for continuing Mux player work.

- [x] Is the naming of `<media-theme template=""></media-theme>` acceptable?
- [x] Should media-theme be included in the media-chrome bundle by default or accessible via a separate directory?
  What should the name be for this directory?
  e.g. `import { MediaThemeElement } from 'media-chrome/media-theme-element';`

The template parts are included in `media-chrome/media-theme-element` and would be exportable


gkatsev:
Is the idea here that MediaThemeTemplate be a declarate way of writing themes and if you wanted to do it otherwise, you can still extend MediaTheme directly?

I think ideally, we'd want a single base-class that would allow for both.

examples/themes/template-theme.html
```diff
@@ -0,0 +1,204 @@
...
+                <svg slot="enter" viewBox="0 0 24 24"><use xlink:href="#fullscreen-enter" /></svg>
+              </media-fullscreen-button>
+            </media-control-bar>
+          </template>
+
+          <!-- This could be a solution for when you'd like to re-use certain elements across different layouts -->
+          <!-- In this example the play and fullscreen button in the bottom control bar could be the same elements in extra small and large,
+                not working yet because they have to be cached in the template parts processor -->
+          <!-- It's quite verbose but this would be the foundation on which it is build, I could see we introduce non-HTML language
+                in the future but it would translate to what you see below at runtime if we chose template parts -->
```
heff, line 95: It does feel pretty verbose, but makes sense as a starting point and good to see the inner workings.

The proposals suggest mustache syntax, which has `{{> thing}}` for partials like this, if we should consider that. Though I'm not sure how you define a partial in the first place within a mustache doc. I only see how to do it in a different file.

src/js/themes/media-theme-template.js
```diff
@@ -0,0 +1,117 @@
...
+        console.log(part);
+        switch (part.directive) {
+          case 'set': {
+            // cache these so they return the same element across layouts
+            state[expression] = new TemplateInstance(part.template, state, mediaThemeProcessor);
+            break;
+          }
+          case 'layout': {
+            if (state.layout === part.expression) {
+              // todo: cache the `TemplateInstance` per layout
```
heff, line 60: What happens if these have to be recreated whenever needed?

src/js/themes/media-theme-template.js
```diff
@@ -0,0 +1,117 @@
...
+
+const mediaThemeProcessor = {
+  processCallback(instance, parts, state) {
+    if (!state) return;
+    for (const [expression, part] of parts) {
+      if (part instanceof InnerTemplatePart) {
+        console.log(part);
+        switch (part.directive) {
+          case 'set': {
+            // cache these so they return the same element across layouts
```
heff, line 54: That means the element can only be used once per layout, which is different from a _partial_ (or just any _template_) which could be used multiple times in the same layout. I can see value in sharing the element in some cases, but i'm not sure that should be the default behavior of a sub-template/partial. I think that's probably what you're getting at by using the term `set`, so I think I get it.

examples/themes/template-theme.html
```diff
@@ -0,0 +1,204 @@
...
+
+          <template directive="layout" expression="live extra-small">
+            <media-control-bar>
+              {{playButton}}
+              <div class="spacer"></div>
+              {{fullscreenButton}}
+            </media-control-bar>
+          </template>
+
+          <template directive="layout" expression="live large">
```
heff, line 116: Could this also work like a conditional? Like `directive="if" expression="?"`

heff (commented):
This is great. 🎉  I like that it's using forward-looking browser templating concepts, while still possible to achieve some of the simpler and well known markup for templating like this.

examples/themes/template-theme.html
```diff
@@ -0,0 +1,204 @@
...
+
+          <!-- This could be a solution for when you'd like to re-use certain elements across different layouts -->
+          <!-- In this example the play and fullscreen button in the bottom control bar could be the same elements in extra small and large,
+                not working yet because they have to be cached in the template parts processor -->
+          <!-- It's quite verbose but this would be the foundation on which it is build, I could see we introduce non-HTML language
+                in the future but it would translate to what you see below at runtime if we chose template parts -->
+          <template directive="set" expression="centerPlayButton">
+            <media-play-button part="center play button" disabled="{{disabled}}" aria-disabled="{{disabled | string}}"></media-play-button>
+          </template>
+          <template directive="set" expression="playButton">
```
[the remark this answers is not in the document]
luwes: indeed these are not standard, we're inventing them.

it's a bit like handlebars helpers, you can add custom ones as well.

src/js/themes/media-theme-template.js
luwes: right now each time the layout is changed it recreates the HTML elements from the templates, it makes a clone basically. I imagined it could be cached based on the layout id which should be unique.

src/js/themes/media-theme-template.js
luwes: yes maybe it's not the best idea, users will have to know the same element can't be in 2 places in the dom.
we can leave caching but it will clone the element each time, maybe it's fine

examples/themes/template-theme.html
luwes: yes, there's one like that for title. 
https://github.com/muxinc/media-chrome/pull/362/files#diff-57ae49d38159fbdf0cae4bb887922912ee746d210c62a5f148f34998c55ace6fR72-R76

expression can contain all kinds of characters.
this is all possible if we'd like
https://github.com/dy/templize#expressions

examples/themes/template-theme.html
luwes: `{{> }}` syntax is totally doable too, to make a difference between escaped and non escaped html

src/js/themes/media-theme-template.js
heff: If there's value in creating a reference to the same element I think we can do both, we just need to be real clear with the naming of the directives between what's meant to be used as a _template partial_ and what's meant to be a _reference_.

src/js/themes/media-theme-template.js
```diff
@@ -0,0 +1,117 @@
...
+    const observer = new MutationObserver(() => this.render());
+    observer.observe(this, { attributes: true });
+  }
+
+  get props() {
+    const props = {};
+    for (let attr of this.attributes) {
+      if (attr.value != null) {
+        props[camelCase(attr.name)] = attr.value === '' ? true : attr.value;
+      } else {
```
cjpillsbury, line 30: issue(blocking): removed attributes will not be members of the `attributes` object, so this case will never be met. One possible alternative is to instead make a the `props` object a `proxy` that evaluates on the key:value pairs on the fly.

src/js/themes/media-theme-template.js
```diff
@@ -0,0 +1,117 @@
...
+    this.shadowRoot.append(this.templateInstance);
+
+    const observer = new MutationObserver(() => this.render());
+    observer.observe(this, { attributes: true });
+  }
+
+  get props() {
+    const props = {};
+    for (let attr of this.attributes) {
+      if (attr.value != null) {
```
cjpillsbury, line 28: issue(blocking): This assumes that values are only ever a string or a boolean, and all based on the value, which also means e.g. `''` **_always_** means `true`, even in cases like `title=''`

luwes: yes, this is an issue that is solved with casting in the template language.
we talked a bit about yesterday, it looks like it will be a requirement to solve this use case cc @heff 

syntax is up for discussion. I added something like `{{ title | string }}` similar to Nunjucks I believe

src/js/themes/media-theme-template.js
luwes: yes, but the change in attributes will be detected and the parts will be checked. 
if `props[partexpression]` is nullish it will be be deleted or handled

luwes:
I vote yes for
1. `<media-theme template=""></media-theme>`
2. `import { MediaTheme } from 'media-chrome/theme';`

luwes:
heh looks familiar https://github.com/tastejs/todomvc/blob/gh-pages/examples/polymer/elements/elements.build.html

src/js/theme/template-processor.js
```diff
@@ -0,0 +1,249 @@
...
+    return parseFloat(raw);
+  }
+
+  // state property
+  return state[raw];
+}
+
+export function isNumeric(str) {
+  if (typeof str != 'string') return false; // we only process strings!
+  return !isNaN(str) && !isNaN(parseFloat(str));
```
gkatsev, line 194: nit: I think this could be simplified with `Number.isNaN`
```suggestion
  return !Number.isNaN(parseFloat(str));
```

src/js/theme/template-parts.js
```diff
@@ -0,0 +1,327 @@
+import { Window as window } from '../utils/server-safe-globals.js';
+
+/* Adapted from https://github.com/dy/template-parts - ISC - Dmitry Iv. */
+
+// Template Instance API
+// https://github.com/WICG/webcomponents/blob/gh-pages/proposals/Template-Instantiation.md
+
+const ELEMENT = 1,
+  STRING = 0,
+  PART = 1;
```
gkatsev, line 10: nit: having them be separate consts would make changing them easier, should we need to (less diff churn)
```suggestion
const ELEMENT = 1;
const STRING = 0;
const PART = 1;
```

gkatsev (commented):
Some mostly minor comments. I do think we're running into a bit of an issue with themes in that we're introducing a new templating language, more or less, which might hurt adoption. Especially, for those that are framework-land like react. That said, I'm not sure if there's necessarily a better alternative, and this new syntax is very htmly and not complicated. Probably, we'd need to make sure we have really good documentation and examples of *everything*.

Also, if we want the new theme stuff to be exported via `media-chrome/theme`, we'd need to add either a file at the root on publish (or always) or add it to the `exports` field in the package.json, though, I think that exports field isn't well-used by tooling yet.

heff:
> Should media-theme be included in the media-chrome bundle by default or accessible via a separate directory?

I don't see a reason to make it separate from the media chrome bundle.

> Is the naming of `<media-theme template=""></media-theme>` acceptable?

I think I like this, as opposed to `<media-theme-loader>` which I proposed elsewhere. In the same sense that `` isn't `<imgloader>`.

Re: `template=""` - I'm considering the different ways someone might _provide_ a theme to this element.
* Provide the ID for a local `<template>`
* Provide a `<template>` element ref (object, not ID) via a property like `mediaTheme.template=`
* Provide a URL to a remote theme
* While we're now defining _themes_ as advanced `<template>`s, there may still be benefit in registering themes as web components, where the web component is just a vehicle for the theme template (e.g. `myThemeElement.theme = <template>`). Web components come with global registration that is accessible in shadow doms.

The last two have implications for the attribute/property design, because we want to avoid potentially conflicting attributes if we can. What if the attribute were `theme` and we did:
* **Theme ID:** `<media-theme theme="#myTemplateId">` (query selector, must be a `<template>` element)
* **Theme URL:** `<media-theme theme="http://player.style/fancy-theme">` (detect URL)
* **Web component:** `<media-theme theme="media-theme-fancy">` (any element name other than `template` we can assume is a custom element name)

We can of course start with the first and ship that then build into the rest.

> `import { MediaTheme } from 'media-chrome/theme';`

I think we need to be more specific than just "theme" here. To keep it more inline with the other element naming I'd call it `media-theme-element`, with the tag still being just `<media-theme>`. The potential confusion is that this element is not a _theme_ itself, it is just loading the theme. Themes are `<template>`s that live elsewhere. In the same sense that the HTMLImageElement loads images from elsewhere.

heff:
Chatted with @luwes. 

I'm on board with starting with `template` as the attr, and deprecating that if needed once we better understand our remote theme loading approach.

So then confirming, I think we go with `media-theme` as the element name. @cjpillsbury brought up the concern of a naming conflict with a base class for a media theme web component, but with the definition of a theme now being ~"representable as a template", I don't think that's a risk. If we move forward with using web components as a vehicle for theme templates, the base class name for that could be MediaThemePackage or something like it.

examples/themes/netflix-theme.html
```diff
@@ -4,22 +4,149 @@
...
         height: 100vh;
       }
+
+      video {
+        width: 100%;      /* prevents video to expand beyond its container */
+      }
     </style>
 
     <main aria-label="Media Chrome Netflix Theme Example">
-      <media-theme-netflix>
```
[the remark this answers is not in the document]
heff: Here's what I think we should do:
* Take the old media-theme.js and rename it media-theme-package.js (open to other names)
* Have that base MediaThemePackage class require that a `<template>` be provided in the init args. (This will help avoid the prior confusion of if a theme is a template or a web component)

With that, you can still build a `<media-theme-netflix>`, which would be a dumb wrapper to the template. You can also then load the netflix theme from `<media-theme>`, like:
```js
const mediaThemeEl = document.querySelector('media-theme')
mediaThemeEl.template = mediaThemeNetflix.template
```

heff:
> I do think we're running into a bit of an issue with themes in that we're introducing a new templating language, more or less, which might hurt adoption. Especially, for those that are framework-land like react. That said, I'm not sure if there's necessarily a better alternative

@gkatsev yeah, I think the option is either going all in on a framework, or sticking as close to HTML as possible to work across frameworks. That said, we can look for extra steps we can take to make the framework-specific experiences better. Maybe there's a way to export a react project to a theme `<template>`, so react devs can build a theme in react? Maybe there's a way to convert an existing theme template into React, using the media chrome react components instead?

gkatsev:
@heff yeah, not sure. Maybe there's a way to translate it across react, but maybe it isn't necessary, especially when we get the pre-packaged themes. The first step is definitely great documentation, though.

src/js/theme/media-theme.js
```diff
@@ -0,0 +1,79 @@
...
+import { processor } from './template-processor.js';
+
+class MediaTheme extends HTMLElement {
+  static observedAttributes = ['template'];
+
+  constructor() {
+    super();
+    this.attachShadow({ mode: 'open' });
+
+    const observer = new MutationObserver(() => this.render());
```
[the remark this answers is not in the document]
luwes: good call, `MutationObserver`'s callback is called async and normally one would iterate through the changed attributes but here we call `render()` 1 time so it works pretty well as a scheduler for multiple attribute changes. they're batched basically

src/js/theme/media-theme.js
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  get mediaController() {
+    // Expose the media controller if API access is needed
+    return this.shadowRoot.querySelector('media-controller');
+  }
+
+  attributeChangedCallback(attrName, oldValue, newValue) {
+    if (attrName === 'template' && oldValue != newValue) {
+      this.template = this.getRootNode().querySelector(
```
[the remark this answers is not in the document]
luwes: will update this, also inline with @heff 's feedback

src/js/theme/media-theme.js
```diff
@@ -0,0 +1,79 @@
...
+
+            if (t.hasAttribute('partial')) directive = 'partial';
+
+            if (directive) {
+              t.setAttribute('directive', directive);
+              t.setAttribute('expression', t.getAttribute(directive));
+            }
+          });
+
+        this.templateInstance = new TemplateInstance(
```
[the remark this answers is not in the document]
luwes: like this idea, will make it more flexible

src/js/theme/template-processor.js
luwes: added some tests for this one.
changing it to only `!Number.isNaN(parseFloat(str));` failed this test because parseFloat tries returns the number part until it reaches the non-numeric part

```js
assert(!isNumericString('12px'), '12px is not numeric');
```

luwes:
Setting aside React and just looking at JSX as a templating language it's missing a few pieces that would make translation from JSX to template parts hard. it's all explained very detailed here by @WebReflection https://webreflection.medium.com/jsx-is-inefficient-by-default-but-d1122c992399

In short it's there is no way to know if a value in JSX was an interpolation or not, and know when the template fragment was started (what is the top of the tree). I also realized this after a while working on Sinuous and htm https://github.com/developit/htm/issues/143

Maybe with a custom Babel plugin it could be done. JSX needs a build step anyway.

WebReflection:
There’s a TC39 proposal around a better standard called ESX https://es.discourse.group/t/proposal-esx-as-core-js-feature/1511

it has a polyfill https://github.com/ungap/esxtoken and a Babel transformer already https://github.com/ungap/babel-plugin-transform-esx

src/js/media-theme-element.js
```diff
@@ -0,0 +1,99 @@
...
+    const templateId = this.getAttribute('template');
+    if (templateId) {
+      const template = this.getRootNode()?.getElementById(templateId);
+      if (template) return template;
+    }
+    return this.#template;
+  }
+
+  set template(element) {
+    this.#template = element;
```
cjpillsbury, line 40: idea(non-blocking): We should probably figure out some way of not requiring an actual DOM element instance of the template having to exist here, otherwise the definition _**has to**_ be explicitly declared on the DOM in the same context/page as the use of it (at least without also including some build step/additional work). This means we can't easily **_define_** a theme template in one place and _**use it**_ in another place. A simple solution for this would be to support a string or element for `set template` and do something like https://stackoverflow.com/questions/494143/creating-a-new-dom-element-from-an-html-string-using-built-in-dom-methods-or-pro/35385518#35385518 (I know, a template used to create a template)

src/js/media-theme-element.js
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  attributeChangedCallback(attrName, oldValue, newValue) {
+    if (attrName === 'template' && oldValue != newValue) {
+      if (this.template) {
+        // Transform short-hand if/partial templates to directive & expression.
+        this.template.content
+          .querySelectorAll('template[if],template[partial]')
```
cjpillsbury, line 48: suggestion(non-blocking): Let's make sure we break these directives into constants and rely on more programmatic stuff here, something like:
```js
this.template.content
  .querySelectorAll(DirectiveNames.map(name => `template[${name}]`).join(','))
  .forEach(templateEl => {
    const directive = DirectiveNames.find(name => templateEl.hasAttribute(name);
    if (directive) {
      templateEl.setAttribute('directive', directive);
      templateEl.setAttribute('expression', t.getAttribute(directive));
    }
  });
```

src/js/media-theme-element.js
```diff
@@ -0,0 +1,99 @@
...
+              t.setAttribute('expression', t.getAttribute(directive));
+            }
+          });
+
+        this.createRenderer();
+      }
+    }
+  }
+
+  get props() {
```
cjpillsbury, line 67: suggestion(non-blocking): Since we have an attributes mutation observer, we should be able to compute and store this as an optimization. Additionally, if we stored a `#props` object that we updated (similar to our `#state` in Mux Player), this is also where we could potentially support other "well-defined" props, a la passing in event handlers (or really any value, I suppose?)

src/js/utils/server-safe-globals.js
```diff
@@ -1,9 +1,14 @@
+class EventTarget {
```
cjpillsbury, line 1: 😎 👌

src/js/media-theme-element.js
```diff
@@ -0,0 +1,99 @@
...
+    this.renderRoot.textContent = '';
+    this.renderRoot.append(this.renderer);
+  }
+
+  render() {
+    this.renderer?.update(this.props);
+  }
+}
+
+function camelCase(name) {
```
cjpillsbury, line 95: suggestion(non-blocking) move to `/utils` somewhere?

src/js/utils/server-safe-globals.js
```diff
@@ -1,9 +1,14 @@
...
+}
+
 const windowShim = {
-  HTMLElement: function HTMLElement() {
-    this.addEventListener = () => {};
-    this.removeEventListener = () => {};
-    this.dispatchEvent = () => {};
-  },
+  HTMLElement: class HTMLElement extends EventTarget {},
+  DocumentFragment: class DocumentFragment extends EventTarget {},
```
cjpillsbury, line 11: 👍

cjpillsbury (approved):
Had a handful of callouts, none of which are blockers for this phase imo, though some of the callouts we should probably work through before we settle on an "official" implementation and not the POC. Overall, great work! 🎉 🚀

src/js/media-theme-element.js
luwes: hmm, I'm not fully convinced this is needed. this was mainly added as a secondary way to set a template element.
first being the `template` attribute that is a `id` to query the DOM.

they can always create a template element if they have the html string right.
maybe as a convenience method setting / getting `templateHTML`

PR #364, opened 2022-11-11
fix: improve CPU usage while playback

luwes:
this reduced CPU usage to around 5% (before ~10%) by removing the expensive `getComputedStyle()` calls during animation frame callbacks.

todo: improve mouse move hover thumb preview as well because it really affects the CPU, throttling the mouse events would work I believe


src/js/media-chrome-range.js
```diff
@@ -209,6 +211,8 @@ class MediaChromeRange extends window.HTMLElement {
     this.container = this.shadowRoot.querySelector('#container');
     this.range = this.shadowRoot.querySelector('#range');
     this.range.addEventListener('input', this.updateBar.bind(this));
+
+    this.#thumbWidth = parseInt(getComputedStyle(this).getPropertyValue('--media-range-thumb-width') || 10);
```
luwes, line 215: we discussed in chat that this would not be changeable after construction.

I think for now this is okay because even when this was called in the `getBarColors()` method changing the CSS var would not have any effect if the media was not playing.

I'd rather have this optimization in than cover the rare case of someone changing the thumb width after initialization. there's also still the option to use CSS transforms, scale

gkatsev, line 215: nit: should add a radix to not parse incorrectly
```suggestion
    this.#thumbWidth = parseInt(getComputedStyle(this).getPropertyValue('--media-range-thumb-width') || 10, 10);
```

gkatsev (approved):
One small nit, otherwise, 👍🏻

PR #365, opened 2022-11-12
feat: declarative listbox

gkatsev:
This defines `media-chrome-listbox`, with a role of `listbox`, and `media-chrome-listitem` with a role of `option`. These internally are basically a `ul` and `li`.
It uses element focus and "roving tabindex" for focus and selection.
`tabindex="0"` on the `media-chrome-listitem` means focused/active. `aria-selected="true"` means that the item is selected. Multi-select is supported by adding `aria-multiselectable="true"` on the `media-chrome-listbox`, which won't remove `aria-selected` from other options before applying for the current option.

TODO:
- [x] ignore disabled items
- [x] typeahead should start at current item
- [x] move example out of advanced.html
- [x] cache assignedElements/slot


examples/advanced.html
```diff
@@ -31,6 +38,13 @@
   <body>
     <main>
       <h1>Media Chrome Advanced Video Usage Example</h1>
+      <media-chrome-listbox aria-multiselectable="true">
```
gkatsev, line 41: this example can probably move into a separate file.

src/js/media-chrome-listbox.js
```diff
@@ -0,0 +1,266 @@
...
+
+    const composedPath = e.composedPath();
+    const slotIndex = composedPath.findIndex(el => el.nodeName === 'SLOT');
+
+    return composedPath[slotIndex - 1];
+  }
+
+  handleSelection(e) {
+    const item = this.#getItem(e);
+
```
luwes, line 175: nit (non-blocking): is the empty newline not too spacey?

src/js/media-chrome-listbox.js
```diff
@@ -0,0 +1,266 @@
...
+  }
+
+  handleMovement(e) {
+    const { key } = e;
+
+    let currentOption = this.#getItem(e);
+    if (!currentOption) {
+      const slot = this.shadowRoot.querySelector('slot');
+      currentOption = slot.assignedElements()
+        .filter(el => el.getAttribute('tabindex') === '0')[0];
```
luwes, line 200: could this be undefined? might need `...)?.[0]`

src/js/media-chrome-listbox.js
```diff
@@ -0,0 +1,266 @@
...
+    super();
+
+    const shadow = this.attachShadow({ mode: 'open' });
+
+    const listboxHTML = template.content.cloneNode(true);
+    this.nativeEl = listboxHTML;
+
+    shadow.appendChild(listboxHTML);
+
+    const slot = this.shadowRoot.querySelector('slot');
```
luwes, line 42: nit (nb): might wanna cache this query, `this.shadowRoot.querySelector('slot')` is used 8 times in this class

src/js/media-chrome-listbox.js
```diff
@@ -0,0 +1,266 @@
...
+
+    slot.addEventListener('slotchange', () => {
+      const els = slot.assignedElements();
+
+      const activeEls = els.some(el => el.getAttribute('tabindex') === '0');
+
+      console.log(els, activeEls);
```
luwes, line 49: B: don't forget console.log();

src/js/media-chrome-listbox.js
```diff
@@ -0,0 +1,266 @@
...
+      const slot = this.shadowRoot.querySelector('slot');
+      slot.assignedElements().forEach(el => el.setAttribute('tabindex', '-1'));
+      nextOption.setAttribute('tabindex', '0');
+      nextOption.focus();
+    }
+  }
+
+  handleClick(e) {
+    const item = this.#getItem(e);
+
```
luwes, line 233: nit (nb): newline

src/js/media-chrome-listbox.js
gkatsev: I think better to have extra whitespace than not enough :)
But I'll keep it in mind as this gets closer to completion.

src/js/media-chrome-listbox.js
gkatsev: Theoretically possible, but I think in practice unlikely.
However, yeah, we should still probably do a null check.

gkatsev:
this is good to review. One thing I'd like to re-implement https://github.com/muxinc/media-chrome/pull/282 on top of this PR before merging this PR in case there's things that come up there that might be helpful.

src/js/media-chrome-listbox.js
```diff
@@ -0,0 +1,278 @@
...
+  #searchItem(key) {
+    this.#clearKeysOnDelay();
+
+    const els = this.#items;
+    const activeIndex = els.findIndex(el => el.getAttribute('tabindex') === '0');
+    const repeatedKey = this.#keysSoFar === key || this.#keysSoFar.length === 0;
+
+    // don't accumulate keys if it's a repeated key
+    // unless we haven't started accumulating a key yet
+    if (!repeatedKey || this.#keysSoFar.length === 0) {
```
cjpillsbury, line 254: question(blocking): Does this mean it's invalid to search for, e.g. "oo"? Should it be (honest question)?

src/js/media-chrome-listbox.js
```diff
@@ -0,0 +1,278 @@
...
+  }
+
+  #clearKeysOnDelay() {
+    window.clearTimeout(this.#clearKeysTimeout);
+    this.#clearKeysTimeout = null;
+
+    this.#clearKeysTimeout = window.setTimeout(() => {
+      this.#keysSoFar = '';
+      this.#clearKeysTimeout = null;
+    }, 500);
```
cjpillsbury, line 272: idea(non-blocking): Should we make this an attr/prop with 500ms the default to help w/a11y for folks who need more time typing?

cjpillsbury (changes requested):
Will do a second pass on this with some smoke testing permutations, but overall 😎 (a couple of small asks).

src/js/media-chrome-listbox.js
```diff
@@ -0,0 +1,278 @@
...
+      if (!elToSelect) {
+        elToSelect = els[0];
+      }
+      
+      elToSelect.setAttribute('tabindex', 0);
+    });
+  }
+
+  get #items() {
+    return this.#assignedElements.filter(el => !el.hasAttribute('disabled'));
```
cjpillsbury, line 69: issue(blocking): We can end up with weird edge cases here if `disabled` is set dynamically, especially if the item was also `selected`. Should we consider adding a well-defined event for `disabled`? We could rely on mutation observer(s)? Thoughts

gkatsev: So, my thought is that `disabled` should win over "selected". If a user is adding `disabled` they probably should also unselect the item as well. We currently don't have a `selectedElement` getter (but we should add one for #282) but I think it would just return nothing if the selected element ends up disabled.

I'm not quite following why we'd need a mutation observer or a disabled event. Are you saying we should deselect `disabled` items for the user?

cjpillsbury: > Are you saying we should deselect disabled items for the user?

Yes, sorry that was my assumption and also wasn't made clear. If we aren't in a multiselect mode, we can get into weird states if we disable a selected and then re-enable (...undisable?) after having selected another item.

gkatsev: Interesting, so, with a `<select>` element, when an option gets disabled, it's still set as the selected item on the select. You just can't re-select it.

So, I think the fix here is that when we reset selected, we don't ignore disabled items. We only ignore selected items for movement.

gkatsev: Looks like I already thought of that and forgot to point it out, haha
https://github.com/muxinc/media-chrome/pull/365/files#diff-2b6b4ea5a2bc8a7b804b26adcccf0cc24948c120c635ca3f64e7929241cf4413R178

src/js/media-chrome-listbox.js
gkatsev: yes, with the current implementation, it isn't possible. Off the top of my head, not sure how to support both.

src/js/media-chrome-listbox.js
gkatsev: yeah, I can see the benefit of having it.

src/js/media-chrome-listbox.js
```diff
@@ -0,0 +1,278 @@
...
+<ul tabindex="0" role="listbox">
+  <slot></slot>
+</ul>
+`;
+
+class MediaChromeListbox extends window.HTMLElement {
+  #keysSoFar = '';
+  #clearKeysTimeout = null;
+  #slot;
+  #assignedElements;
```
cjpillsbury, line 25: suggestion(non-blocking): If this is likely to be a base class for the dynamic/data driven implementation(s), I suspect some of these likely should not be private (and ECMA doesn't have protected access modifiers 👎)

src/js/media-chrome-listbox.js
cjpillsbury: It looks like the reference implementation for listbox doesn't treat repeated keys as a special case:
https://codepen.io/pen?editors=0010
(referenced here: https://www.w3.org/WAI/ARIA/apg/example-index/listbox/listbox-scrollable.html)

src/js/media-chrome-listbox.js
```diff
@@ -0,0 +1,278 @@
...
+    const repeatedKey = this.#keysSoFar === key || this.#keysSoFar.length === 0;
+
+    // don't accumulate keys if it's a repeated key
+    // unless we haven't started accumulating a key yet
+    if (!repeatedKey || this.#keysSoFar.length === 0) {
+      this.#keysSoFar += key;
+    }
+
+    // if it's a repeat key, skip the current item
+    const after = els.slice(activeIndex + (repeatedKey ? 1 : 0)).filter(el => el.textContent.startsWith(this.#keysSoFar));
```
cjpillsbury, line 259: the reference impl handles this via the delay (relates to above convo).

src/js/media-chrome-listbox.js
gkatsev: possibly, though, might be useful to have actual getters for things. Like, we probably need a `selectedItem` and `selectedItems` getters and maybe an `options` getter too.

Originally, I was thinking this should stay in PR until we build #282 on top of this, but maybe we can merge this if it's "good enough" but not expose it from `index.js` until that time. That way, we don't have a long-lasting PR but we also don't give folks a half baked listbox unless they explicitly ask for it.

src/js/media-chrome-listbox.js
```diff
@@ -0,0 +1,278 @@
...
+    if (!item) return;
+
+    const selected = item.getAttribute('aria-selected') === 'true';
+
+    if (this.getAttribute('aria-multiselectable') !== 'true') {
+      this.#assignedElements.forEach(el => el.setAttribute('aria-selected', 'false'));
+    }
+
+    if (selected) {
+      item.setAttribute('aria-selected', 'false');
```
heff, line 182: This seems to assume that you can turn off a selection and have no item selected. Thinking of something like the  playback rate menu, you can't turn off playback rate. Something always has to be selected. Should we be making that an option at this level, or should whatever's using this handle that?

src/js/media-chrome-listbox.js
heff: Whitespace warrior Wes. :)

I don't know why, but anything sitting right on top of an if() feels wrong to me. Gary and I have clearly worked in the same codebases. :)

src/js/media-chrome-listitem.js
```diff
@@ -0,0 +1,106 @@
...
+      MediaUIAttributes.MEDIA_CONTROLLER
+    );
+    if (mediaControllerId) {
+      const mediaControllerEl = document.getElementById(mediaControllerId);
+      mediaControllerEl?.unassociateElement?.(this);
+    }
+  }
+
+  get keysUsed() {
+    return ['Enter', ' '];
```
heff, line 98: Does it use these? Not totally familiar with how keysUsed is used, but this feels handled by listbox.

heff (approved):
Good stuff 👍 None of my comments are blocking. I approve from what I see, assuming other threads are answered.

src/js/media-chrome-listbox.js
gkatsev: I think that would happen at a higher level, like the playback rate menu, rather than here. Because it's reasonable to be able to unselect things for a generic listbox.

src/js/media-chrome-listbox.js
gkatsev: > Gary and I have clearly worked in the same codebases. :)

Yeah, we might've 😆

src/js/media-chrome-listbox.js
gkatsev: Given that the w3c example implementation uses 500, I'd say we leave it as is until we need it.

src/js/media-chrome-listbox.js
gkatsev: Let me see if I can get both working. I personally find the repeated keys thing super useful. If it's not a lot of code and doesn't take too long to implement, I'll do it, otherwise I'll take it out.

gkatsev: I played around with how Windows Explorer behaves in this case, and I think I have a way forward here.
Basically, you'd still want to accumulate the keys, but if it's the same letter, just treat it as a single key unless there's a better match. Shouldn't be too bad to add in.

gkatsev: Pushed a change for this

src/js/media-chrome-listbox.js
cjpillsbury: Still have some concerns around the "duplicate key" approach here, but I don't think it needs to be a blocker since followups can be isolated if we decide to change this behavior.

cjpillsbury (approved):
🚢 🇮🇹

PR #368, opened 2022-11-18
Fix mc attr

### LINKED ISSUE #366: media-controller attribute broken for buttons
In the following reduced test case...
Expected: the media-play-button will work the same as a play button inside of a media-controller.
Actual: The first button click (play) works but the following (pause) does not

At load the `media-paused` attribute is appropriately getting added to `media-play-button`, so the first part of connecting the button to the controller is working. However when the video plays, the `media-paused` attribute is not removed as needed for the button to work as expected.

```html
      <media-controller id="controllerID">
        <video slot="media" src="https://stream.mux.com/DS00Spx1CV902MCtPj5WknGlR102V5HFkDe/high.mp4" crossorigin muted playsinline></video>
      </media-controller>

      <media-play-button media-controller="controllerID"></media-play-button>
```

heff:
Fixes #366 

The media-controller attribute on UI components stopped working at some point, and it appears to be because the attribute was being lumped in with all the state attributes. I think that was having a few weird side effects, but the result was the button's media-controller attribute was getting reset to nothing by the media controller, then causing the button to unregister itself from the media controller (which was only half working. The play request event was still working after being unregistered. So something maybe worth digging more into).

### DIFF
--- a/src/js/constants.js
+++ b/src/js/constants.js
@@ -50,6 +50,11 @@ export const MediaStateChangeEvents = {
   USER_INACTIVE: 'userinactivechange',
 };
 
+export const UIControllerAttributes = {
+  MEDIA_CHROME_ATTRIBUTES: 'media-chrome-attributes',
+  MEDIA_CONTROLLER: 'media-controller',
+};
+
 export const MediaUIAttributes = {
   MEDIA_AIRPLAY_UNAVAILABLE: 'media-airplay-unavailable',
   MEDIA_FULLSCREEN_UNAVAILABLE: 'media-fullscreen-unavailable',
@@ -75,8 +80,6 @@ export const MediaUIAttributes = {
   MEDIA_PREVIEW_TIME: 'media-preview-time',
   MEDIA_PREVIEW_IMAGE: 'media-preview-image',
   MEDIA_PREVIEW_COORDS: 'media-preview-coords',
-  MEDIA_CHROME_ATTRIBUTES: 'media-chrome-attributes',
-  MEDIA_CONTROLLER: 'media-controller',
   MEDIA_LOADING: 'media-loading',
   MEDIA_BUFFERED: 'media-buffered',
 };
--- a/src/js/constants.js
+++ b/src/js/constants.js
@@ -46,8 +46,6 @@ export const MediaStateChangeEvents = {
   MEDIA_SEEKABLE: 'mediaseekablechange',
   MEDIA_PREVIEW_IMAGE: 'mediapreviewimagechange',
   MEDIA_PREVIEW_COORDS: 'mediapreviewcoordschange',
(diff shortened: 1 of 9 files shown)


src/js/constants.js
```diff
@@ -46,12 +46,15 @@ export const MediaStateChangeEvents = {
...
   MEDIA_LOADING: 'medialoadingchange',
   USER_INACTIVE: 'userinactivechange',
 };
 
+export const UIControllerAttributes = {
+  MEDIA_CHROME_ATTRIBUTES: 'media-chrome-attributes',
+  MEDIA_CONTROLLER: 'media-controller',
+};
+
 export const MediaUIAttributes = {
```
heff, line 58: I was tempted to rename MediaUIAttributes to MediaStateAttributes to be clearer, but that would have touched a lot of files and it's only an internal thing.

src/js/constants.js
```diff
@@ -46,12 +46,15 @@ export const MediaStateChangeEvents = {
   MEDIA_SEEKABLE: 'mediaseekablechange',
   MEDIA_PREVIEW_IMAGE: 'mediapreviewimagechange',
   MEDIA_PREVIEW_COORDS: 'mediapreviewcoordschange',
-  // MEDIA_CHROME_ATTRIBUTES: 'media-chrome-attributes',
-  // MEDIA_CONTROLLER: 'media-controller',
```
dylanjha, line 50: does @cjpillsbury remember why these were commented out?

heff: If there was a good reason to keep them around it probably applied to the the other set of constants, and probably got carried over when creating multiple lists. These are attributes in a list events.

luwes (approved):
LGTM makes sense

src/js/constants.js
cjpillsbury: they were only there for parity with `MediaUIAttributes`. I found it easier to A/B between the two with them there, but they can be removed.

src/js/constants.js
cjpillsbury: I think that would be a good followup renaming though.

src/js/constants.js
```diff
@@ -46,12 +46,15 @@ export const MediaStateChangeEvents = {
...
   MEDIA_LOADING: 'medialoadingchange',
   USER_INACTIVE: 'userinactivechange',
 };
 
+export const UIControllerAttributes = {
```
cjpillsbury, line 53: idea(non-blocking): `MediaStateReceiverAttributes`?

cjpillsbury (approved):
One non-blocking suggestion/idea for the lookup object name, otherwise 😎

src/js/constants.js
heff: Oh yeah, that makes sense. 👍

PR #371, opened 2022-11-18
Media Stream Type state

heff:
Related to: https://github.com/video-dev/media-ui-extensions/issues/3

Adding a `media-stream-type` state/attribute for UI elements and themes to use when differentiating between live and on-demand use cases.

It relies on the the `media.duration` value, that you can rely on for Live vs. On-demand otherwise.
* Infinity = Live
* NaN = 🤷 
* Any other number = On-demand

I also added a `default-stream-type` attribute for `<media-controller>`, for when you want to preset the stream type UI before the media is loaded. It's somewhat modeled after defaultMuted, in that it gets overridden as soon as there's any real value for media-stream-type.


src/js/media-controller.js
```diff
@@ -43,7 +44,7 @@ const DEFAULT_TIME = 0;
  */
 class MediaController extends MediaContainer {
   static get observedAttributes() {
-    return super.observedAttributes.concat('nohotkeys', 'hotkeys');
+    return super.observedAttributes.concat('nohotkeys', 'hotkeys', 'default-stream-type');
```
heff, line 47: Question: Should this have dashes or not? I see hotkeys/nohotkeys didn't follow the dash format (more HTML-like). Do we want to move that direction in this area where it's not a media-state attribute?

src/js/media-controller.js
```diff
@@ -640,6 +644,10 @@ class MediaController extends MediaContainer {
   }
 
   propagateMediaState(stateName, state) {
+    if (arguments.length === 1) {
+      state = Delegates[stateName](this);
+    }
```
heff, line 649: Now that we have the Delegates concept I _think_ we could lean on those more in other areas, but I didn't want to do a big refactor in the middle of this one. I did however add this nicety to let propagateMediaState default to using the delegate, but interested if that could go wrong somehow.

src/js/media-controller.js
cjpillsbury: So right now the not quite perfect distinction we have is:

1. media state attributes: attribute names are hyphenated
2. Media Chrome component-specific attributes: unhyphenated

This one falls right in the middle, but I'd lean towards treating it like (1) since it directly relates to defaulting `media-stream-type`.

cjpillsbury: fwiw, I'm definitely on board with dropping hyphens on everything (relates to forward looking framework compat/ease when working with web components), though @dylanjha has some strong gut against readability/devex for "smashed attribute names". I just think that should be an "all at once" decision, if we make the change.

src/js/media-controller.js
```diff
@@ -947,7 +955,26 @@ const Delegates = {
...
-  }
+  },
+
+  [MediaUIAttributes.MEDIA_STREAM_TYPE](el) {
+    const media = el.media;
+    const duration = media.duration;
+
+    if (duration === Infinity) {
+      return StreamTypes.LIVE;
+    } else if (!isNaN(duration)) {
```
cjpillsbury, line 966: nitpick(blocking): we should use `Number.isNaN()` if we want to keep this strict:

```js
isNaN(undefined) // true
Number.isNaN(undefined) // false
isNaN(NaN) // true
Number.isNaN(NaN) // true
```

src/js/media-controller.js
```diff
@@ -947,7 +955,26 @@ const Delegates = {
...
+      return StreamTypes.ON_DEMAND;
+    } else {
+      const defaultType = el.getAttribute('default-stream-type');
+
+      if (Object.values(StreamTypes).includes(defaultType)) {
```
cjpillsbury, line 971: nitpick(non-blocking): consider pulling out `Object.values(StreamTypes)` since it shouldn't need to be re-evaluated every method invocation.

cjpillsbury (changes requested):
one minor ask and one non-blocking suggestion/nitpick.

src/js/media-controller.js
cjpillsbury: This is a pattern we tend to use on elements (see: https://github.com/muxinc/elements/search?q=Object.values)

src/js/media-controller.js
cjpillsbury: whoops per callout, for strictness this should actually be:
`typeof duration === 'number' && Number.isNaN(duration)` (though `Number.isNaN` vs. `isNaN` is superfluous here). Alternatively, consider changing the order of default vs. on demand checks.

cjpillsbury:
suggestion(non-blocking): Consider adding/updating an example to demonstrate using `media-stream-type` + CSS attribute selectors for UI updates

src/js/media-controller.js
gkatsev: an alternative "single check" is `Object.prototype.toString.call(duration) === '[object Number]'`, via https://github.com/ljharb/is/blob/master/index.js#L437

(still better to use `Number.isNaN` because `isNaN` could have false positives)

gkatsev: maybe a better check here is `Number.isFinite()`?

luwes (approved):
LGTM, could do docs later

docs/src/pages/en/media-controller.md
```diff
@@ -8,31 +8,17 @@ The `<media-controller/>` manages communication of state and state change reques
 
 ## Attributes
 
-- `autohide (seconds, default: 0)` - (`video` only) Use this to autohide all controls/chrome (except for the `media`) after `n` seconds of inactivity, unless the media is paused. To disable `autohide`, set the value to -1.
+### autohide
```
cjpillsbury, line 11: idea(non-blocking): As a followup, especially if we're pairing down the examples, I'm wondering if we should just make the `media-controller` documentation look more like our other Media State Receiver components (i.e. just tables) for consistency.

docs/src/pages/en/media-controller.md
```diff
@@ -42,92 +28,60 @@ Example (`autohide` disabled):
...
-  </media-control-bar>
+  ...
 </media-controller>
 ```
+
+### default-stream-type
+
+`default-stream-type` (values: `live`, `on-demand`)
+
+Media controller can't know the stream type of the media until the media is loaded. Setting `default-stream-type` can prevent UI changes happening between when the player is loaded and when the media is loaded. This may happen when a player is built to support both stream types, and then is used to play a stream type that is different from the player's default.
```
cjpillsbury, line 81: suggestion(non-blocking):
Since we're at least kind of introducing a new concept "stream type," consider replacing 
> Media controller can't know the stream type of the media until the media is loaded.
with
> Media controller can't know if the media will be live or on demand until after the media is loaded.
or similar

cjpillsbury: suggestion(non-blocking): Link to new stream-type docs page somewhere here "For more info on stream types, see"

examples/stream-type.html
```diff
@@ -0,0 +1,56 @@
...
+      media-controller[media-stream-type=on-demand] .on-demand {
+        display: block;
+      }
+
+    </style>
+  </head>
+  <body>
+    <main>
+
+      <media-controller default-stream-type>
```
cjpillsbury, line 44: suggestion(non-blocking): Since this isn't linked on the main `index.html` for the examples, I think this is fine, but we can/should provide some "real boy" examples of live and on demand (maybe after we do a fast follow/spike on supporting this in mux-video+mux-audio.

cjpillsbury (approved):
🏄‍♂️ a couple of non-blocking comments, mostly as potential followups.

examples/stream-type.html
heff: Oof, forgot to clean this one up. I was gonna just drop the example html file and rely on the docs.

docs/src/pages/en/media-controller.md
heff: I agree. Gonna save that for a separate pass.

PR #374, opened 2022-11-30
docs: add version, size badge

luwes:
SCR-20221130-eh0

it's actually a little noisy. useful info but maybe needs to be placed somewhere else?


heff:
> maybe needs to be placed somewhere else?

🤷 Maybe under the tag line? Feel free to merge as is.

What does "size" refer to specifically? Seeing as we have multiple versions you can use I wonder what the right answer to that is.

gkatsev:
> Seeing as we have multiple versions you can use I wonder what the right answer to that is.

it's set up as the min-gzipped size. The format is `minzip` in the url https://img.shields.io/bundlephobia/minzip/media-chrome?label=size

cjpillsbury (approved):
👍

PR #377, opened 2022-12-01
feat: Captions/Subtitles listbox

gkatsev:


src/js/media-captions-list.js
```diff
@@ -0,0 +1,97 @@
...
+    super.disconnectedCallback();
+  }
+
+  get mediaController() {
+    return this.#mc;
+  }
+
+  #setupList() {
+    const kinds = ['subtitles', 'captions'];
+    const tracks = Array.from(this.#mc.media.textTracks).filter(t => kinds.includes(t.kind));
```
cjpillsbury, line 56: This should be owned by the `MediaController` since text tracks will frequently be async and based on the `src`, can change over time for some use cases, etc. etc. It will also stay consistent with the current data flow architecture. That said, given the acrobatics required in my POC draft, I strongly suspect that we should have a single value for both captions _**and**_ subtitles. We can add a "kind" to the serialization used in the attribute, or experiment with using properties for this one (though that is a larger architectural introduction, so may slow down discussion).

gkatsev: yup, the next step is to use the captions and subtitles list attributes a la your PR. While I agree that maybe we should just have a single list or probably figure out how to do this via props, I don't think I'd want to do it as part of this PR.

gkatsev:
So, I think the three outstanding issues, two of which I'm not sure how to solve at the moment (and could potentially not be solvable):
- using just the keyboard, when you select an element in the captions list, the focus disappears.
- when using the screen reader, it doesn't announce that we are focused on an item that's inside a listbox (might need to put a role of listbox on the listbox element itself in additon to the shadow dom ul element)
- when using an SR, it doesn't announce how to select the currently selected item

src/js/media-captions-list.js
```diff
@@ -0,0 +1,133 @@
...
+  #subs = [];
+  #caps = [];
+
+  static get observedAttributes() {
+    return [
+      ...super.observedAttributes,
+      MediaUIAttributes.MEDIA_CAPTIONS_LIST,
+      MediaUIAttributes.MEDIA_CAPTIONS_SHOWING,
+      MediaUIAttributes.MEDIA_SUBTITLES_LIST,
+      MediaUIAttributes.MEDIA_SUBTITLES_SHOWING,
```
gkatsev, line 19: if the media-chrome captions and subtitles and selected list was modeled differently, it would be a lot easier to work with.

src/js/media-chrome-listbox.js
```diff
@@ -1,4 +1,4 @@
-import { MediaUIAttributes } from './constants.js';
```
gkatsev, line 1: these fixes can probably be separated into a separate PR against main as the listbox is currently broken due to MEDIA_CONTROLLER moving out from MediaUIAttributes.

gkatsev:
>  when using an SR, it doesn't announce how to select the currently selected item

I think this piece isn't really doable without swapping to native elements or using a checkbox or radio as the list item or something. This [combobox example](https://w3c.github.io/aria-practices/examples/combobox/combobox-select-only.html) behaves similarly, with VO only announcing it's a text element.

src/js/media-chrome-listbox.js
```diff
@@ -183,6 +194,9 @@ class MediaChromeListbox extends window.HTMLElement {
     } else {
       item.setAttribute('aria-selected', 'true');
     }
+
```
cjpillsbury, line 197: suggestion(non-blocking): we should check if the selection actually changed and only do the extra work (including dispatching the event) if the selection actually changed. Given the complexities of multiselectable support, this can be a followup if it feels like too heavy a lift rn.

cjpillsbury (changes requested):
Still reviewing but here's initial feedback!

src/js/media-chrome-listbox.js
gkatsev: you mean to verify whether the new item is equivalent to the previous item and not trigger otherwise?

I'm not sure if that would be an issue since if you select an already selected item, it deselects it currently.

src/js/media-captions-list.js
cjpillsbury: Upvote. @heff I believe we chatted about this. Tl;dr - If we had attributes and corresponding events that treated captions and subtitles as a "single list" (would need to add the type to our string-serialized representation), this would significantly reduce complexities.

src/js/media-chrome-listitem.js
```diff
@@ -34,6 +39,14 @@ class MediaChromeListitem extends window.HTMLElement {
     shadow.appendChild(listitemHTML);
   }
 
+  set value(value) {
```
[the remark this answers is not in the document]
gkatsev: This could be useful for either case. For example, you could have an image for each of the list items, but you'd want the value to be a number or something. This matches `<option>`'s `value` attr and propr.

src/js/media-captions-list.js
```diff
@@ -0,0 +1,137 @@
...
+    });
+  }
+
+  #render() {
+    this.#perTypeRender(this.#subs, 'subs');
+    this.#perTypeRender(this.#caps, 'cc');
+  }
+
+  #onChange(e) {
+    const [newType, selectedOption] = this.selectedOptions[0]?.value?.split('!') ?? [];
```
cjpillsbury, line 99: idea(non-blocking): A lot of UIs will have some sort of "default" item to select for lists. In the case of Captions/Subtitles Menus, that would be some kind of "None" option. In the case of (future) Rendition Selection, that would be some kind of "Automatic" option. We may want to anticipate that kind of capability/config use case (which may also mean some mechanism to disable "click etc. to deselect selected item"). Food for thought.

cjpillsbury, line 99: suggestion(non-blocking): given that we only use `value` for identifying data type here and it's "meaningless" to the base `media-chrome-list-item`, we should consider just creating an instance-level `Map` on `media-captions-list` where keys are the "options" (elements) and values are the `value`.

src/js/media-chrome-listitem.js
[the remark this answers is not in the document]
cjpillsbury: I'm sold.

src/js/media-chrome-listbox.js
```diff
@@ -195,20 +209,20 @@ class MediaChromeListbox extends window.HTMLElement {
...
+        if (nextOption?.hasAttribute('disabled')) {
           nextOption = nextOption.nextElementSibling;
         }
 
         break;
       case 'ArrowUp':
         nextOption = currentOption.previousElementSibling;
 
-        if (nextOption.hasAttribute('disabled')) {
+        if (nextOption?.hasAttribute('disabled')) {
```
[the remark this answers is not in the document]
gkatsev: oh, good call. Forgot to add it.

src/js/media-captions-list.js
gkatsev: Since we're keeping it on listitem, we don't need this

gkatsev: You mean like an `off` item? I think that would work but yeah, would require a bit of work. Potentially also need to not unset an option if clicked on a selected option in this case.

cjpillsbury: yup, we'd have to decouple some assumptions here. A "light lift" version would be to have a "baked in" version that can be turned on/off via an attr that we could identify as the "off" item in the logic here to do some "forking paths" on behavior.

cjpillsbury (approved):
oh word. 🚀

src/js/media-captions-list.js
dylanjha: Agree with consolidating. The only tradeoff I remember was that we were trying to be technically accurate with the terms "subtitles" / "captions" and not conflate two the terms.

Practically speaking, this did cause confusion when building the player Demuxed 2022 theme because I think:

- The test streams they were using had subtitles, so the theme worked in testing
- The real live stream on the day of the event had captions, and the theme didn't account for that

Also, I didn't fully confirm this but I think we had a bug where in Chrome it was being treated as captions and in Safari it was treated as subtitles, so that was hella confusing.

gkatsev: > I think we had a bug where in Chrome it was being treated as captions and in Safari it was treated as subtitles

That's because Safari treats [any characteristic on subtitles tracks](https://datatracker.ietf.org/doc/html/draft-pantos-hls-rfc8216bis#section-4-4-6-1:~:text=A%20SUBTITLES%20Rendition%20MAY%20include%20the%20following%20characteristics) in the manifest as a caption, while hls.js doesn't treat only treats both as a caption? Also, I think this may have been updated since.

> technically accurate with the terms "subtitles" / "captions" and not conflate two the terms

Basically, only the US treats them materially different. Most of the rest of the world treats them both as subtitles, but captions as "Subtitles for the Deaf and Hard of Hearing" (SDH). Most UIs show them in a single list and then specify that a particular track is specifically a captions track rather than just a regular subtitles track.

> this did cause confusion when building the player Demuxed 2022 theme

Another reason to just have a single list.

src/js/media-captions-list.js
```diff
@@ -0,0 +1,143 @@
+import MediaChromeListbox from './media-chrome-listbox.js';
+import { defineCustomElement } from './utils/defineCustomElement.js';
+import { Window as window, Document as document } from './utils/server-safe-globals.js';
+import { MediaUIAttributes, MediaStateReceiverAttributes, MediaUIEvents } from './constants.js';
+import { nouns } from './labels/labels.js';
+import { parseTextTracksStr, parseTextTrackStr, formatTextTrackObj } from './utils/captions.js';
+
+
+class MediaCaptionsList extends MediaChromeListbox {
```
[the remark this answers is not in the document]
gkatsev: I think I mostly left `box` off to make the name not be quite as long.

gkatsev:
Follow-ups issue https://github.com/muxinc/media-chrome/issues/387

src/js/media-captions-list.js
heff: Let's do it. 👍 (in a follow up)

PR #378, opened 2022-12-01
Fix directive aliases in inner template

luwes:
This fixes an issue where the inner template aliases were not transformed and broke the directives.

### DIFF
--- a/src/js/utils/template-processor.js
+++ b/src/js/utils/template-processor.js
@@ -63,7 +63,7 @@ export const processor = {
 
     for (const [expression, part] of parts) {
       if (part.directive && part instanceof InnerTemplatePart) {
-        Directives[part.directive](expression, part, state);
+        Directives[part.directive]?.(expression, part, state);
         continue;
       }
 
@@ -160,8 +160,7 @@ export function evaluateCondition(expr, state) {
       console.warn(`Warning: invalid expression \`${expr}\``);
       return false;
     }
-    const val = getParamValue(tokens[0].token, state);
-    return val;
+    return getParamValue(tokens[0].token, state);
   }
 
   const args = tokens.filter(({ type }) => type !== 'ws');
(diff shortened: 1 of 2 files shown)


src/js/utils/template-processor.js
```diff
@@ -23,33 +23,47 @@ const operators = {
...
-            if (evaluateCondition(part.expression, state)) {
-              part.replace(
-                new TemplateInstance(part.template, state, processor)
-              );
-            } else {
-              part.replace('');
-            }
-            break;
-        }
+      if (part.directive && part instanceof InnerTemplatePart) {
```
cjpillsbury, line 65: 😎 I like this.

cjpillsbury: question(non-blocking): Should this conditional order be switched?

src/js/utils/template-processor.js
```diff
@@ -23,33 +23,47 @@ const operators = {
...
+      part.replace(new TemplateInstance(part.template, state, processor));
+    } else {
+      part.replace('');
+    }
+  },
+};
+
+const DirectiveNames = Object.keys(Directives);
+
+export function transformDirectiveAliases(template) {
```
cjpillsbury, line 47: suggestion(non-blocking): I think `transform` functions tend to indicate (still potentially mutative) new objects (e.g. https://lodash.com/docs/4.17.15#transform or https://www.educative.io/answers/what-is-the-transform-function-in-cpp). Consider calling this `updateDirectiveAliases` or `modifyDirectAliases` or something like that? (I dunno, names are hard)

cjpillsbury (approved):
Only a few non-blocking comments. Otherwise, 👌

src/js/utils/template-processor.js
luwes: I imagined a property lookup is faster than an instanceof check but not sure.
yes, I can see why it makes a little more sense in an ordering sense to first check instanceof but perf :)

PR #382, opened 2022-12-02
Feature: Seek-to-live button with related events & states

heff:
Modeled after Mux Player's implementation.

### `<media-live-button>`
image

Indicate the video is currently live and click to seek to live. Includes slots for alt icons and test, and css variables for changing colors etc.

### mediaseektoliverequest event
Event that when handled will seek to the end of the seekable window

### media-time-is-live attribute
State for when the current time is near enough to the live edge to be considered "live"

### Todo

- [x] Docs/Examples
- [x] Finish MEDIA_TIME_IS_LIVE state
- [x] Optional auto seek to live on unpause


src/js/media-controller.js
```diff
@@ -981,6 +997,44 @@ const Delegates = {
...
+      return true;
+    }
+
+    if (media.seekable.length === 0) {
+      return false;
+    }
+
+    // Default to 10 seconds
+    // Assuming seekable range already accounts for appropriate buffer room
+    const liveTimeMargin = 10;
```
cjpillsbury, line 1018: idea(non-blocking): This is the sort of thing that could be more reliably computed by a(n extended) slotted media element based on spec-compliant/recommended behaviors on refetches, segment sizes, etc. etc. Would be a good followup to spike on.

src/js/media-live-button.js
```diff
@@ -0,0 +1,89 @@
...
+  <slot name="indicator">${indicatorSVG}</slot>
+  <slot name="text">Live</slot>
+`;
+
+class MediaLiveButton extends MediaChromeButton {
+  static get observedAttributes() {
+    return [
+      ...super.observedAttributes,
+      MEDIA_PAUSED,
+      MEDIA_TIME_IS_LIVE
```
cjpillsbury, line 58: suggestion(non-blocking): We should add `MEDIA_STREAM_TYPE` to the list of observed attributes to make styling cases easier.

src/js/media-controller.js
heff: Yeah, I agree. That'd be a good way to set your latency preferences on the media element specifically and have it impact the right things in the UI, without latency needing to be a bigger concept (at least yet).

Start a media ui extension issue?

src/js/media-live-button.js
heff: 👍 done

heff:
Notes from IRL conversation:
* Pin the progress bar to the right when in the live window
* Always seek to the live edge during unpause
* Turn off red indicator when unpaused, at least until we know the difference between seekable and non-seekable live streams

heff:
Office hours request: Option to turn on/off seek-to-live on unpause

heff:
- [x] Docs completed
- [x] Optional seek to live on unpause added. Attr is `autoseektolive="off"`

docs/src/pages/en/media-live-button.md
```diff
@@ -0,0 +1,115 @@
...
+| `spacer`  | `text` | A simple text space (`&nbsp;`) between the indicator and the text. |
+
+## CSS Vars
+
+See our [styling docs](./styling#Buttons)
+
+| Var    | Description |
+| ------ | ----------- |
+| `--media-live-indicator-icon-color` | The color of the indicator icon when the audio or video is live  |
+| `--media-live-indicator-off-icon-color` | The color of the indicator icon when the audio or video is *not* live  |
```
dylanjha, line 115: docs look great, but needs to be added to the side nav in `config.ts`:

https://github.com/muxinc/media-chrome/blob/main/docs/src/config.ts#L65

dylanjha (changes requested):

docs/src/pages/en/media-live-button.md
gkatsev, line 115: thought:
- `--media-live-indicator-icon-color` for the default/non-live color
- `--media-live-indicator-live-icon-color` for the live color

src/js/media-controller.js
```diff
@@ -93,7 +93,16 @@ class MediaController extends MediaContainer {
 
     // Capture request events from internal controls
     const mediaUIEventHandlers = {
-      MEDIA_PLAY_REQUEST: () => this.media.play(),
+      MEDIA_PLAY_REQUEST: (e, media) => {
+        const streamType = Delegates[MediaUIAttributes.MEDIA_STREAM_TYPE](this);
+        const autoSeekToLive = this.getAttribute('autoseektolive');
```
gkatsev, line 98: `autoseektolive=off` doesn't feel very htmly also increases inconsistency between attributes in media-chrome. That said, don't have any alternatives off the top of my head currently.

src/js/media-live-button.js
```diff
@@ -0,0 +1,80 @@
+import MediaChromeButton from './media-chrome-button.js';
+import { defineCustomElement } from './utils/defineCustomElement.js';
+import {
+  Window as window,
+  Document as document,
+} from './utils/server-safe-globals.js';
+import { MediaUIEvents, MediaUIAttributes } from './constants.js';
+import { verbs } from './labels/labels.js';
```
gkatsev, line 8: nit: unused
```suggestion
```

src/js/media-live-button.js
```diff
@@ -0,0 +1,80 @@
...
+  constructor(options = {}) {
+    super({ slotTemplate, ...options });
+    this.setAttribute('aria-label', 'Seek to Live');
+  }
+
+  attributeChangedCallback(attrName, oldValue, newValue) {
+    super.attributeChangedCallback(attrName, oldValue, newValue);
+  }
+
+  handleClick(e) {
```
gkatsev, line 62: nit: unused `e`
```suggestion
  handleClick() {
```

src/js/media-controller.js
luwes: yes for a boolean attribute we should try to not use `off`, `false` as value.
best to inverse it then.

`noseektolive`, `disableliveseek`, `noplaylive`

heff: Yeah naming this one was hard for me. Open to other ideas but I feel good enough about where it landed. It's also flexible incase we wanted to capture other options with it, like maybe one that tries to keep the playhead at the live edge even without unpausing.

> increases inconsistency between attributes in media-chrome

How so?

docs/src/pages/en/media-live-button.md
```diff
@@ -0,0 +1,115 @@
...
+  <span slot="text"></span>
+</media-live-button>
+```
+
+## Attributes
+
+| Name            | Type      | Default Value | Description |
+| --------------- | --------- | ------------- | ----------- |
+| `media-time-is-live` | `boolean` | `false`| Include when the media time is at or close to the most current time or _live edge_. Clicking the button will not seek to live when true. [_Set automatically by media-controller._] |
+| `media-paused` | `boolean` | `false`| Include when the media is paused. The button will unpause the video if paused when seeking to live. [_Set automatically by media-controller._] |
```
cjpillsbury, line 98:
```suggestion
_None_
```
Currently, we only document attributes intended to be set by an implementor for customization/configuration (and not attributes set by the Media Controller). See, e.g. [media-play-button.md](https://github.com/muxinc/media-chrome/blob/main/docs/src/pages/en/media-play-button.md). If we'd like to document both (ideally with some differentiation between those two with an eye towards devex), we should plan this and do it as a single effort.

cjpillsbury (changes requested):

docs/src/pages/en/media-live-button.md
heff: I see how you get there, and feel the awkwardness in it. Feels a little weird to have "live" in there twice, and I expect "live indicator color" to refer to the red not the "off" state.

I changed it to:
* `--media-live-indicator-color` - squarely focused on the color of live and re-useable
* `--media-live-button-icon-color` - because what it's really doing is overriding the default `--media-icon-color`

Make sense?

docs/src/pages/en/media-live-button.md
heff: Yeah, I saw that, and I disagree with only documenting the non-media-controller attributes, based on the goal of [element independence](https://github.com/muxinc/media-chrome/blob/main/ARCHITECTURE.md#element-independence). I also don't love requiring a big project to ship it all at once since the temporary inconsistency doesn't really hurt anyone and it's no small effort to document all the detail per attr per element. But I agree in that I think we should make it a goal of 1.0 to document those attrs everywhere so the inconsistency has a clear end date, and in that we can create some better visual separation between them.

docs/src/pages/en/media-live-button.md
heff: Added. Good now?

dylanjha (approved):

docs/src/config.ts
```diff
@@ -79,6 +79,7 @@ export const SIDEBAR: Sidebar = {
       { text: 'Time display', link: 'en/media-time-display' },
       { text: 'Time range', link: 'en/media-time-range' },
       { text: 'Volume range', link: 'en/media-volume-range' },
+      { text: 'Live Indicator/Button', link: 'en/media-live-button' },
```
gkatsev, line 82: Can this be moved above `Loading Inidcator` to keep this list in alphabetical order?

heff: Ah, missed that. Yes. Thanks.

src/js/media-controller.js
luwes: `autohide="2"`
`gestures-disabled` why not `gestures="off"` then?
`nohotkeys` => better remove this one and add an `hotkeys="off"` option
`hotkeys="noarrowleft noarrowright"` 
`default-stream-type`
`audio`

I agree that the uniformness is starting to worsen between these handful of attributes.

we should decide 
- dashes vs no dashes
- how should a disabled attribute be. if it's `feature="off"` fine with me but best to keep it consistent

docs/src/pages/en/media-live-button.md
cjpillsbury: I do worry that having this one place where we document the "internal" attributes will be confusing in the interim and increases the chance of mistakes when we **_do_** update the docs to include both. I'd still recommend that we don't document currently (there's already some implicit documentation of those attributes above in terms how how they update the visual state), but I won't treat this as a blocker.

src/js/media-controller.js
cjpillsbury: I'd agree with @gkatsev and @luwes here. For an attribute currently named `autoseektolive`, I'm having a hard time imagining expanding values beyond a `boolean` (though that may just be a failure of imagination). As such, we might as well stick with both our convention and the common convention for boolean attributes in HTML: if the attribute is present (typically with a value of `""`), that implies `true`. Otherwise (aka it's not present), that implies `false`.

gkatsev (approved):

src/js/media-controller.js
cjpillsbury: @heff Yeah, good call re: media ui extension.

docs/src/pages/en/media-live-button.md
cjpillsbury: This might be a good thing for our next media chrome weekly meeting agenda. A thought I'm having is:
Should this be two separate variables at all? Should we instead just assume folks can currently use attribute selectors and potentially eventually use [Custom States](https://developer.mozilla.org/en-US/docs/Web/API/CustomStateSet) (which we keep running into candidate use cases for)?

src/js/media-controller.js
heff: Yeah I see the reason here, just trying to get this thing shipped already. 😆 Happy to give up on a multi-value approach if no one sees opportunity for expansion.

`<form>` at least has `novalidate`, so there's some precedence for no*, and I can get behind that. I think my favorite might be `noautoseektolive` (or dashed). `noseektolive` would be less verbose, but that kind of sounds like we're disabling the button from seeking to live too (which is why I added `auto` in the first place).

I can't think of any good option to reverse the relationship so it's not a negative. `manualseektolive`, `maintainliveplayheadposition`, `makeuserclicklivebuttontoseektolive`. :)

src/js/media-live-button.js
```diff
@@ -0,0 +1,78 @@
...
+    color: var(--media-live-indicator-color, rgb(255, 0, 0));
+  }
+
+  </style>
+
+  <slot name="indicator">${indicatorSVG}</slot>
+  <!-- 
+    A new line between spacer and text creates inconsistent spacing
+    between slotted items and default slots.
+  -->
```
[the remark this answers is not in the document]
heff: This is a good idea. Just tried this and had the same result of mismatched spacing. There's something weird happening between slotted elements and default slots that I can't quite figure out. Going to ship as is but feel free to try your hand at it if you're interested enough. If I'm not just missing something dumb I'm sure we'll run into it again.

docs/src/pages/en/media-live-button.md
heff: You can't use attribute selectors to target the indicator color specifically of the built-in indicator. You'd have to override the default slot to use attribute selectors. Or we'd have to expose it as a part. I could be fine with the latter, but I'm not sure how that works if the part is also a slot, and then implications of that for skins/themes.

CustomStateSet is the dream. Do we have a sense for how long we're going to be waiting for those?

PR #383, opened 2022-12-03
fix: support more expression syntax for size logic

luwes:
Discussion in https://github.com/muxinc/media-chrome/discussions/370

The goal is to extend the expression syntax enough to evaluate more complex conditions like 

```html
<template if="streamtype == 'live' && breakpoint == 'md'"></template>
```


luwes:
my S1 is that it's a bit limiting but it could work if we don't need much complex conditions and conditional chaining.

my mind went to Nunjuck's layout blocks initially but I can get used to the `block` naming.
https://mozilla.github.io/nunjucks/

the new extra attributes don't communicate much to what they do though. would it need more specific naming?

```html
<template block="streamType" equals="live">
  <template block="containerWidth" lessthan="490">...</template>
  <template block="containerWidth" greaterthan="590">...</template>
</template>
```

I'm a bit less fond of it because it's more verbose.
container queries and media queries are now all accepting the range syntax
https://caniuse.com/css-media-range-syntax

what about a middle ground? 

option 1
still have conditional chaining but make it clear it's not JS 

```html
<template if="type == 'on-demand' and width < 420"></template>
<template if="type == 'on-demand' or width > 590"></template>
```

or even (can be anything)

```html
<template match="(type: on-demand) and (width < 420)"></template>
<template match="(type: on-demand) or (width > 590)"></template>
```

option 2
no conditional chaining but keep condition operators

```html
<template if="streamType == live">
  <template if="containerWidth < 490">...</template>
  <template if="containerWidth > 590">...</template>
</template>
```

luwes:
yes, those last examples would work like that. the `if` and `for` or `foreach` are separate directives.
this is the base of that https://github.com/luwes/template-extensions#innertemplatepart

> Two of the benefits I see with block are the simplicity of "exists/does not exist" and templating a list.

Is there a way to write the "does not exist" condition w/ blocks? I didn't see that one. Or would the user need to do that themselves? Provide 2 params, one for the if condition and one for the else condition.

luwes:
I think I'm on board with trying the `block` directive.
it will be hard to deprecate the JS like expressions later on.

So syntax will be:

`<template block="streamType">...</template>`
`<template block="streamType" value="live">...</template>`
`<template block="listVariable"><li class={{class}} data-value={{value}}>{{label}}</li></template>`

Maybe wait with max / min until there is a real need? 
The breakpoints PR should cover the values to test equality against.

luwes:
Closing this PR in favor of keeping expression syntax as possible.
For a logical AND we will use nested templates

PR #389, opened 2022-12-08
docs: README and docs update

dylanjha:
- [x] slimmed down README to remove a lot of the duplication that's already in docs
- [x] added 'Audio player' Guide with a simple audio example
- [x] added missing attribute docs for `audio` on Media controller
- [x] added "In the wild" to Learn More section in nav to show places where Media Chrome is used in the wild
- [x] added "Basics" > "Media slot" doc with details about the media slot, media element and how that works
- [x] added "Media Clip Selector" under components


docs/src/pages/en/in-the-wild.md
```diff
@@ -0,0 +1,34 @@
+---
+title: Media Chrome in the wild
+description: Links to other resources for all things Media Chrome and the ideas behind Media Chrome
+layout: ../../layouts/MainLayout.astro
+---
+
+<h2>Using Media Chrome</h2>
```
gkatsev, line 7: long term, I think something like https://babeljs.io/en/users.html would be cool.

docs/src/config.ts
```diff
@@ -79,6 +81,10 @@ export const SIDEBAR: Sidebar = {
       { text: 'Time display', link: 'en/media-time-display' },
       { text: 'Time range', link: 'en/media-time-range' },
       { text: 'Volume range', link: 'en/media-volume-range' },
+      { text: 'Clip selector (extra)', link: 'en/media-clip-selector' },
+    ],
+    "Learn More": [
+      { text: 'In the wild', link: 'en/in-the-wild' },
```
davekiss, line 87: I wonder if there's a better phrase to use here for navigational purposes. Spitballing...

- Who's using media-chrome?
- Public sightings
- Production demos
- Showcase
- Talks and examples

docs/src/pages/en/audio-player.md
```diff
@@ -0,0 +1,28 @@
+---
+title: Audio player
+description: Build an audio player with Media Chrome
+layout: ../../layouts/MainLayout.astro
+---
+
+Specify the `audio` attribute on `<media-controller>` in order to use the default layout. When using the `audio` attribute the positioning slots like `"top-chrome"`, `"centered-chrome"`, etc. that are specified in the [Position controls](/en/position-controls) guide do not exist. There is only one default slot that all your markup can go into.
```
davekiss, line 7: The second sentence here feels a bit passive. A stab at reworking it:

"Using the `audio` attribute will remove the standard [positioning slots](/en/position-controls) from your player (eg: `"top-chrome"`, `"centered-chrome"`)."

docs/src/pages/en/get-started.md
```diff
@@ -124,26 +124,4 @@ Use HTML to add or remove any of the controls. Then you can use CSS to style the
...
-
-Be sure to include the `slot="media"` attribute in the player's tag.
-
-```html
-<media-controller>
-  <youtube-video slot="media" src="https://www.youtube.com/watch?v=rubNgGj3pYo">
-  </youtube-video>
-</media-controller>
-```
+Media Chrome will work with any HTML element that exposes the same API as HTML media elements (`<video>` and `<audio>`). See the guide for the [media slot](/en/media-slot) to learn more about how media elements work with the media slot
```
davekiss, line 127: Add period

docs/src/pages/en/in-the-wild.md
```diff
@@ -0,0 +1,34 @@
...
+
+Player by Caffeine.tv on [brobible.com](https://brobible.com/sports/article/mike-tyson-made-30-million-biting-evander-holyfield/)
+
+
+
+<h2>Talks, blog posts and other resources</h2>
+
+Here are some resources that relate to the evolving ideas, development, & history of Media Chrome.
+
+[Designing a player for the context and UX you want](https://tmi.mux.com/sessions/designing-a-player-for-the-context-and-ux-you-want) -- Christian Pillsbury, TMI, 2022
```
davekiss, line 21: Do we want to link to the author's bio somewhere? Always nice to be able to learn who the person is behind the work.

docs/src/pages/en/media-clip-selector.md
```diff
@@ -0,0 +1,42 @@
...
+description: Media Clip Selector
+layout: ../../layouts/MainLayout.astro
+source: https://github.com/muxinc/media-chrome/tree/main/src/js/extras/media-clip-selector/index.js
+---
+
+A draggle UI on the time range for creating media clips.
+
+<img src="/assets/clip-selector.png" alt="Media Clip Selector" />
+
+**Note:** `<media-clip-selector>` is an extra, which means it is not imported with the default bundle, it needs to be imported as-needed from the dist directory, for example:
```
davekiss, line 12: **Note:** `<media-clip-selector>` is an extra, which means it is not imported with the default bundle. To use it, you will need to import it from the `dist` directory manually. For example:

docs/src/pages/en/media-clip-selector.md
```diff
@@ -0,0 +1,42 @@
...
+```js
+const mediaClipSelector = document.querySelector('media-clip-selector');
+
+mediaClipSelector.addEventListener('update', (evt) => {
+  const { startTime, endTime } = evt.detail;
+  console.log('clip selection updated:', startTime, endTime);
+});
+```
+
+See [example](https://media-chrome.mux.dev/examples/control-elements/media-clip-selector.html)
```
davekiss, line 29: Here's a [live example](https://media-chrome.mux.dev/examples/control-elements/media-clip-selector.html) of the media clip selector in action.

docs/src/pages/en/media-clip-selector.md
```diff
@@ -0,0 +1,42 @@
...
+
+_None_
+
+## Slots
+
+_None_
+
+## Events
+
+- `update` fires when the clip selector range has been updated. See usage example.
```
davekiss, line 41: Link to usage example

docs/src/pages/en/media-slot.md
```diff
@@ -0,0 +1,43 @@
...
+</media-controller>
+```
+
+In this example, Media controller will:
+
+* Listen for events emitted from the `<video>` element
+* Understand the current state of the `<video>` element by using the known props
+* Call methods on the `<video>` element like `play()`, `pause()`, etc.
+
+The superpower of Media Chrome is that the media slot does not have to be a `<video>` element. It can be any HTML element that [implements the HTML5 Video Element API](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video).
```
davekiss, line 27: Can we provide examples of other HTML elements that implement this API?

davekiss (changes requested):

docs/src/pages/en/media-slot.md
dylanjha: that's in the next section under `Media Elements`, could that be more clear?

docs/src/config.ts
dylanjha: I like "Showcase" 👍🏼 I think when we have more to show we can break these up into sub-pages but for now since we only have a few I think having a single "Showcase" page that has links to all of this stuff I think makes sense?

docs/src/pages/en/in-the-wild.md
dylanjha: That's a good idea. I'm going to skip that for now, but I think this showcase page is something we will iterate on.

Mostly right now I want to have it so that we have a place to put things as they come up, and accept PRs from outside folks adding their things to it. We should continue to improve it!

README.md
```diff
@@ -52,192 +52,9 @@ Visit the [official documentation for Media Chrome](https://www.media-chrome.org
 
 #### <a href="https://media-chrome.mux.dev/examples/" target="_blank">See all of the repo examples in action.</a>
 
-## Audio Example
```
heff, line 55: Let's keep the simple audio example here too, please. This will be a pretty common landing point for users and they're pretty different audiences. I think we could do better upholding the audio side of media chrome in general.

README.md
```diff
@@ -5,7 +5,7 @@ Your media player's dancing suit. :man_dancing:
 Fully customizable media player controls using web components (native custom elements).
 
 - [x] Compatible with any javascript framework (React, Angular, Svelte, etc.)
-- [x] Compatible with the `<video>` and `<audio>` elements and [a lot of players](#compatible-players) (YouTube, HLS.js, and more)
+- [x] Compatible with the `<video>` and `<audio>` elements and [a lot of players](https://www.media-chrome.org/en/media-slot) (YouTube, HLS.js, and more)
```
heff, line 8: Nit: Don't need `www`

heff (approved):
👍

docs/src/pages/en/media-slot.md
davekiss: Ah I didn't make the connection initially. Since a new h2 follows this paragraph, it felt like a new thought. Maybe it's clear after all, or we could add a "Read on for some examples" or "Here are a few examples:" something to that effect?

luwes:
This is great! 

Could the media-clip-selector also be a live example like the other controls? the screenshot doesn't look too good
https://media-chrome-docs-git-fork-dylanjha-dj-docs-updates-mux.vercel.app/en/media-clip-selector

dylanjha:
Thanks @luwes -- I split out the clip selector guide into another branch because it wasn't working quite right on the docs page. The thumbnail container was positioned incorrectly. Might be a bug with the clip selector itself, will have to dig into that one.

Good call on the other improvements though, better to actually render the real web component instead of doing screenshots!

@heff I put the audio example back

docs/src/pages/en/showcase.md
```diff
@@ -0,0 +1,36 @@
...
+layout: ../../layouts/MainLayout.astro
+---
+
+## Who's using Media Chrome
+
+Here are some projects and web applications that currently use Media Chrome:
+
+[Mux Player](https://mux.com/player). Mux's in-house video player uses Media Chrome under the hood. In addition to the linked product page, Mux Player is also [open source under Mux Elements](https://github.com/muxinc/elements/tree/main/packages/mux-player).
+
+<img src="/public/mux-player-media-chrome.png" width="500" alt="Mux Player built with Media Chrome" />
```
davekiss, line 13: This image is missing in the deploy preview

https://media-chrome-docs-git-fork-dylanjha-dj-docs-updates-mux.vercel.app/en/showcase

docs/src/pages/en/showcase.md
```diff
@@ -0,0 +1,36 @@
...
+
+Player by Caffeine.tv on [brobible.com](https://brobible.com/sports/article/mike-tyson-made-30-million-biting-evander-holyfield/)
+
+<img src="/public/caffeine-tv-media-chrome.png" width="500" alt="Caffeine TV player built with Media Chrome" />
```
davekiss, line 17: This one seems missing too

davekiss: CleanShot 2022-12-13 at 12 43 29@2x

davekiss:
This seems to 404 as well, maybe we need a rebuild?

https://media-chrome-docs-git-fork-dylanjha-dj-docs-updates-mux.vercel.app/en/media-clip-selector

davekiss (approved):

PR #390, opened 2022-12-13
fix: remove Demuxed theme 2022 from MC bundle

luwes:
was accidentally included in the main Media Chrome bundle


gkatsev (approved):

dylanjha (approved):
I'm surprised we don't have a lint rule to catch unused variables

luwes:
looks like we don't lint in CI, it would have caught that. will add that here

Issue #387, opened 2022-12-07
Listbox followups


gkatsev:
- [x] a selectmenu-like menu - https://github.com/muxinc/media-chrome/pull/393
- [x] default styles for the listbox (from https://github.com/muxinc/media-chrome/pull/377#discussion_r1041606019) - https://github.com/muxinc/media-chrome/pull/399
- [x] removing a track should update the captions listbox appropriately - https://github.com/muxinc/media-chrome/pull/399
- [x] Off item for the captions listbox. Could be an option for the listbox whether to create an automatic off item or could be a captions listbox specific feature. Currently, the way to turn off an item on a listbox is to interact on the selected item. This probably is unwanted if there is an explicit Off item. (from https://github.com/muxinc/media-chrome/pull/377#discussion_r1041424407 https://github.com/muxinc/media-chrome/pull/377#pullrequestreview-1207545629) - https://github.com/muxinc/media-chrome/pull/399
- [x] have a hover style on listitems - https://github.com/muxinc/media-chrome/pull/426
- [x] keep the captions menu button within the controller bounds - https://github.com/muxinc/media-chrome/pull/426
- [ ] The captions button should toggle no captions and the previously selected track. If no previous selection was present, do the current algorithm selection of choosing the first item. See https://github.com/muxinc/media-chrome/pull/442/commits/5dbc248b108a4caeb06d59870fbec94be98a3f3c for inspiration, but should be contained to the captions button only.
- [x] rename `media-chrome-listitem` to `media-chrome-option`
- [x] make `media-chrome-listitem` more similar to `<option>` element in terms of props/attrs
  - things like `selected` attribute instead of require `aria-selected="true"`, and then we'd want to sprout the corresponding aria attr dependingly.
- [ ] make `media-chrome-listbox` more similar to `<select>`/`<selectmenu>` element in terms of props/attrs
  - things like `multiple` prop instead of requiring `aria-multiselectable="true"`
- [x] when a listbox is used in a selectmenu, clicking the option itself shouldn't deselect it
- [ ] provide a programmatic interface to the listbox so that the selected items can be changed programmatically.
- [x] consolidate MEDIA_CAPTIONS_LIST and MEDIA_SUBTITLES_LIST
- [ ] Consider choosing the captions track that matches the media-controller's lang/locale if available.
- [ ] (maybe) open on hover for captions menu
- [x] animate opening/closing of the menu
- [x] customize things like selected item by prepending a checkmark
[...]

PR #393, opened 2022-12-14
media-chrome-menu-button

gkatsev:
Adds a media-chrome-menu-button that mostly does what we expect. Not sure if menu-button is necessarily the best name, but it works. Happy to change it, though.

Usage:
```html
<media-chrome-menu-button aria-label="captions menu button">
  <svg slot="button-content" aria-hidden="true" viewBox="0 0 26 24"></svg>
  <media-captions-listbox slot="listbox" media-controller="mc"></media-captions-listbox>
</media-chrome-menu-button>
```

The next simple follow-up is a captions-menu-button that basically wraps up this above usage in a prepackaged component.
There's still some more follow up to do as per #387. With a new Off menu item, and to make it more accessible, we may want to update listbox to be able to be a role of menu instead of listbox with a listitem a role of menuitem over options.
Those things would be the immediate follow-ups, but I think we can still get this PR in with the caveat that things could change as we continue building this out.

### DIFF
--- a/examples/control-elements/media-chrome-menu-button.html
+++ b/examples/control-elements/media-chrome-menu-button.html
@@ -7,6 +7,7 @@
   <script type="module" src="../../../dist/media-chrome-listbox.js"></script>
   <script type="module" src="../../../dist/media-chrome-listitem.js"></script>
   <script type="module" src="../../../dist/media-captions-listbox.js"></script>
+  <script type="module" src="../../../dist/media-chrome-menu-button.js"></script>
   <style>
     media-captions-listbox {
       display: block;
@@ -31,7 +32,24 @@
 <body>
   <main>
     <h1>Media Chrome Standard Video Usage Example</h1>
-    <media-captions-listbox title="Captions & Subtitles List" media-controller="mc" ></media-captions-listbox>
+    <media-chrome-menu-button>
+      <media-chrome-listitem>lorem</media-chrome-listitem>
+      <media-chrome-listitem>ipsum</media-chrome-listitem>
+      <media-chrome-listitem>dolor</media-chrome-listitem>
+      <media-chrome-listitem>sit</media-chrome-listitem>
+    </media-chrome-menu-button>
+
+    <br>
+
+    <media-chrome-menu-button aria-label="captions menu button">
+      <svg slot="button-content" aria-hidden="true" viewBox="0 0 26 24">
+        <path d="M22.83 5.68a2.58 2.58 0 0 0-2.3-2.5c-3.62-.24-11.44-.24-15.06 0a2.58 2.58 0 0 0-2.3 2.5c-.23 4.21-.23 8.43 0 12.64a2.58 2.58 0 0 0 2.3 2.5c3.62.24 11.44.24 15.06 0a2.58 2.58 0 0 0 2.3-2.5c.23-4.21.23-8.43 0-12.64Zm-11.39 9.45a3.07 3.07 0 0 1-1.91.57 3.06 3.06 0 0 1-2.34-1 3.75 3.75 0 0 1-.92-2.67 3.92 3.92 0 0 1 .92-2.77 3.18 3.18 0 0 1 2.43-1 2.94 2.94 0 0 1 2.13.78c.364.359.62.813.74 1.31l-1.43.35a1.49 1.49 0 0 0-1.51-1.17 1.61 1.61 0 0 0-1.29.58 2.79 2.79 0 0 0-.5 1.89 3 3 0 0 0 .49 1.93 1.61 1.61 0 0 0 1.27.58 1.48 1.48 0 0 0 1-.37 2.1 2.1 0 0 0 .59-1.14l1.4.44a3.23 3.23 0 0 1-1.07 1.69Zm7.22 0a3.07 3.07 0 0 1-1.91.57 3.06 3.06 0 0 1-2.34-1 3.75 3.75 0 0 1-.92-2.67 3.88 3.88 0 0 1 .93-2.77 3.14 3.14 0 0 1 2.42-1 3 3 0 0 1 2.16.82 2.8 2.8 0 0 1 .73 1.31l-1.43.35a1.49 1.49 0 0 0-1.51-1.21 1.61 1.61 0 0 0-1.29.58A2.79 2.79 0 0 0 15 12a3 3 0 0 0 .49 1.93 1.61 1.61 0 0 0 1.27.58 1.44 1.44 0 0 0 1-.37 2.1 2.1 0 0 0 .6-1.15l1.4.44a3.17 3.17 0 0 1-1.1 1.7Z"/>
+      </svg>
+      <media-captions-listbox slot="listbox" media-controller="mc"></media-captions-listbox>
+    </media-chrome-menu-button>
+
+    <br>
+
     <media-controller id="mc">
       <video
         slot="media"
--- a/src/js/media-chrome-menu-button.js
+++ b/src/js/media-chrome-menu-button.js
@@ -1,0 +1,156 @@
+import './media-chrome-button.js';
+import './media-chrome-listbox.js';
+import { defineCustomElement } from './utils/defineCustomElement.js';
+import { Window as window, Document as document } from './utils/server-safe-globals.js';
+import { MediaStateReceiverAttributes } from './constants.js';
+
+const template = document.createElement('template');
+template.innerHTML = `
+  <style>
+  :host {
+    display: block;
+  }
+
+  media-chrome-button:not(:focus-visible) {
+    outline: none;
+  }
+  </style>
+
+  <media-chrome-button aria-haspopup="listbox">
+    <slot name="button-content"></slot>
+  </media-chrome-button>
+  <slot name="listbox" hidden>
+    <media-chrome-listbox id="listbox">
+      <slot></slot>
+    </media-chrome-listbox>
+  </slot>
+`;
+
+class MediaChromeMenuButton extends window.HTMLElement {
+  #handleClick;
+  #handleChange;
+  #button;
+  #listbox;
+  #listboxSlot;
+  #expanded = false;
+
+  static get observedAttributes() {
+    return [
+      'disabled', MediaStateReceiverAttributes.MEDIA_CONTROLLER,
+    ];
+  }
+
+  constructor() {
+    super();
+
+    const shadow = this.attachShadow({ mode: 'open' });
+
+    const buttonHTML = template.content.cloneNode(true);
+    this.nativeEl = buttonHTML;
+
+    shadow.appendChild(buttonHTML);
+
+    this.#handleClick = this.#handleClick_.bind(this);
+    this.#handleChange = this.#handleChange_.bind(this);
+
+    this.#button = this.shadowRoot.querySelector('media-chrome-button');
+    this.#listbox = this.shadowRoot.querySelector('media-chrome-listbox');
+
+    this.#listboxSlot = this.shadowRoot.querySelector('slot[name=listbox]');
+    this.#listboxSlot.addEventListener('slotchange', () => {
+      this.#listbox.removeEventListener('change', this.#handleChange);
+      // update listbox reference if necessary
+      this.#listbox = this.#listboxSlot.assignedElements()[0] || this.#listbox;
+      this.#listbox.addEventListener('change', this.#handleChange);
+    });
+  }
+
+  #handleClick_() {
+    this.#toggle();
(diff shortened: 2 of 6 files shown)


src/js/media-chrome-menu-button.js
```diff
@@ -0,0 +1,147 @@
...
+  .hidden {
+    display: none;
+  }
+
+  media-chrome-button:not(:focus-visible) {
+    outline: none;
+  }
+  </style>
+
+  <media-chrome-button aria-haspopup="listbox" aria-controls="listbox">
```
cjpillsbury, line 23: question(non-blocking): since `aria-controls="listbox"` relies on `id="listbox"` on the default slotted listbox, won't this break for any externally-slotted listbox? Haven't tested yet in your `media-chrome-menu-button` example, but I believe the captions use case should be sufficient to test this?

src/js/media-chrome-menu-button.js
```diff
@@ -0,0 +1,147 @@
...
+
+      if (assignedListbox) {
+        this.#listbox = this.#listboxSlot.assignedElements()[0];
+      }
+    });
+  }
+
+  #handleClick_() {
+    this.#listboxSlot.classList.toggle('hidden');
+    this.#expanded = !this.#expanded;
```
cjpillsbury, line 73: suggestion(non-blocking): Consider moving `this.#expanded = !this.#expanded` into `#toggleExpanded()`

cjpillsbury (approved):
One minor potentially blocking callout (depending on your answer) and a few suggestions/questions. Otherwise, this is great progress 🎉. Pre-approving.

src/js/media-chrome-menu-button.js
```diff
@@ -0,0 +1,147 @@
...
+    this.#button = this.shadowRoot.querySelector('media-chrome-button');
+    this.#listbox = this.shadowRoot.querySelector('media-chrome-listbox');
+
+    this.#listboxSlot = this.shadowRoot.querySelector('slot[name=listbox]');
+    this.#listboxSlot.addEventListener('slotchange', () => {
+      const assignedListbox = this.#listboxSlot.assignedElements()[0];
+
+      if (assignedListbox) {
+        this.#listbox = this.#listboxSlot.assignedElements()[0];
+      }
```
luwes, line 67:
```suggestion
this.#listbox = this.#listboxSlot.assignedElements()[0] || this.#listbox;
```

src/js/media-chrome-menu-button.js
```diff
@@ -0,0 +1,147 @@
...
+      if (oldValue) {
+        const mediaControllerEl = document.getElementById(oldValue);
+        mediaControllerEl?.unassociateElement?.(this);
+        this.#listbox.removeAttribute(MediaStateReceiverAttributes.MEDIA_CONTROLLER);
+      }
+      if (newValue) {
+        const mediaControllerEl = document.getElementById(newValue);
+        mediaControllerEl?.associateElement?.(this);
+        this.#listbox.setAttribute(MediaStateReceiverAttributes.MEDIA_CONTROLLER, newValue);
+      }
```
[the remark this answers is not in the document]
cjpillsbury: > is the unassociating needed when there is a newValue? seems unneeded

@luwes : Yeah. We don't have examples/use cases of this, but it's for cases where there are 2+ `<media-controller/>` instances on the same page and the media state receiver element has some mechanism for toggling between them. A "Quad View" with an "active video" could be implemented this way, for example.

examples/control-elements/media-chrome-menu-button.html
```diff
@@ -0,0 +1,89 @@
+<!DOCTYPE html>
```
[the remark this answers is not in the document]
gkatsev: nope, oversight.

src/js/media-chrome-menu-button.js
gkatsev: yeah, that makes sense. Not exactly sure how to handle this, since we probably don't want to add an ID to a slotted component if it doesn't already have one?

Unfortunately, the new upcoming additions to aria that help in this case won't be available for a while (also, I'm not sure if there's a settled solution quite yet or that there's multiple solutions).

src/js/media-chrome-menu-button.js
```diff
@@ -0,0 +1,147 @@
...
+    const shadow = this.attachShadow({ mode: 'open' });
+
+    const buttonHTML = template.content.cloneNode(true);
+    this.nativeEl = buttonHTML;
+
+    shadow.appendChild(buttonHTML);
+
+    this.#handleClick = this.#handleClick_.bind(this);
+
+    this.#button = this.shadowRoot.querySelector('media-chrome-button');
```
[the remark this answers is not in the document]
gkatsev: Does it matter? I think it's probably best to not query the DOM each time.

src/js/media-chrome-menu-button.js
gkatsev: https://www.w3.org/WAI/ARIA/apg/patterns/menubutton/ says that aria-controls is optional.
So, I think for the MVP there are two options:
- not use it at all
- use it but when a listbox is slotted, remove aria-controls from the button.

gkatsev: Based on this article, https://heydonworks.com/article/aria-controls-is-poop/ I say we ignore that attribute altogether.

src/js/media-chrome-menu-button.js
```diff
@@ -0,0 +1,147 @@
...
+const template = document.createElement('template');
+template.innerHTML = `
+  <style>
+  :host {
+    display: block;
+  }
+
+  .hidden {
+    display: none;
+  }
```
gkatsev, line 16: this should be replaced with the hidden attribute instead.

src/js/media-chrome-menu-button.js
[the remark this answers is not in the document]
cjpillsbury: Mostly just makes refactors/code management easier over change. Definitely only a suggestion. Will resolve.

src/js/media-chrome-menu-button.js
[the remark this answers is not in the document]
luwes: oh ok cool, good to know

examples/control-elements/media-chrome-menu-button.html
```diff
@@ -0,0 +1,92 @@
...
+<head>
+  <meta name="viewport" content="width=device-width">
+  <title>Media Chrome Captions Usage Examples</title>
+  <script type="module" src="../../dist/index.js"></script>
+  <script type="module" src="../../../dist/media-chrome-listbox.js"></script>
+  <script type="module" src="../../../dist/media-chrome-listitem.js"></script>
+  <script type="module" src="../../../dist/media-captions-listbox.js"></script>
+  <script type="module" src="../../../dist/media-chrome-menu-button.js"></script>
+  <style>
+    media-captions-listbox {
```
heff, line 12: Really not loving the growing examples that require CSS to work. I can see myself using this component directly, so I don't think the same reasoning applies here as in the other conversation.

examples/control-elements/media-chrome-menu-button.html
```diff
@@ -0,0 +1,92 @@
...
+    .examples {
+      margin-top: 20px;
+    }
+  </style>
+</head>
+<body>
+  <main>
+    <h1>Media Chrome Standard Video Usage Example</h1>
+    <media-chrome-menu-button>
+      <media-chrome-listitem>lorem</media-chrome-listitem>
```
heff, line 36: You can either do list items in the default slot or swap the listbox? That's pretty cool.

examples/control-elements/media-chrome-menu-button.html
```diff
@@ -0,0 +1,92 @@
...
+      <svg slot="button-content" aria-hidden="true" viewBox="0 0 26 24">
+        <path d="M22.83 5.68a2.58 2.58 0 0 0-2.3-2.5c-3.62-.24-11.44-.24-15.06 0a2.58 2.58 0 0 0-2.3 2.5c-.23 4.21-.23 8.43 0 12.64a2.58 2.58 0 0 0 2.3 2.5c3.62.24 11.44.24 15.06 0a2.58 2.58 0 0 0 2.3-2.5c.23-4.21.23-8.43 0-12.64Zm-11.39 9.45a3.07 3.07 0 0 1-1.91.57 3.06 3.06 0 0 1-2.34-1 3.75 3.75 0 0 1-.92-2.67 3.92 3.92 0 0 1 .92-2.77 3.18 3.18 0 0 1 2.43-1 2.94 2.94 0 0 1 2.13.78c.364.359.62.813.74 1.31l-1.43.35a1.49 1.49 0 0 0-1.51-1.17 1.61 1.61 0 0 0-1.29.58 2.79 2.79 0 0 0-.5 1.89 3 3 0 0 0 .49 1.93 1.61 1.61 0 0 0 1.27.58 1.48 1.48 0 0 0 1-.37 2.1 2.1 0 0 0 .59-1.14l1.4.44a3.23 3.23 0 0 1-1.07 1.69Zm7.22 0a3.07 3.07 0 0 1-1.91.57 3.06 3.06 0 0 1-2.34-1 3.75 3.75 0 0 1-.92-2.67 3.88 3.88 0 0 1 .93-2.77 3.14 3.14 0 0 1 2.42-1 3 3 0 0 1 2.16.82 2.8 2.8 0 0 1 .73 1.31l-1.43.35a1.49 1.49 0 0 0-1.51-1.21 1.61 1.61 0 0 0-1.29.58A2.79 2.79 0 0 0 15 12a3 3 0 0 0 .49 1.93 1.61 1.61 0 0 0 1.27.58 1.44 1.44 0 0 0 1-.37 2.1 2.1 0 0 0 .6-1.15l1.4.44a3.17 3.17 0 0 1-1.1 1.7Z"/>
+      </svg>
+      <media-captions-listbox slot="listbox" media-controller="mc"></media-captions-listbox>
```
heff, line 48: Does the slot _have_ to be a listbox? Or should we call this slot `menu` (since it's the menu button) and allow that slot to be more flexible?

heff (commented):
A few comments, but in general I'm excited for this one. It's gonna unlock a lot.

Not a blocker, but a related topic that came up was handling responsiveness + menus, i.e. when the menu might break out of the player box. Do we have any thoughts on solving that yet?

examples/control-elements/media-chrome-menu-button.html
gkatsev: Default styles is a follow-up that I'll be tackling next. See https://github.com/muxinc/media-chrome/issues/387 before we call this whole thing complete.

examples/control-elements/media-chrome-menu-button.html
gkatsev: Called it `listbox` to match [selectmenu's slot](https://open-ui.org/prototypes/selectmenu). Ultimately, anything with a matching API would work.

gkatsev:
> Not a blocker, but a related topic that came up was handling responsiveness + menus, i.e. when the menu might break out of the player box. Do we have any thoughts on solving that yet?

Definitely something to consider once it's in the player. Theoretically, container queries could help since it can give us a size based on both the block and inline sizes of the container. Plus, there's the breakpoint stuff that's coming in themes.

examples/control-elements/media-chrome-menu-button.html
gkatsev: Yup, inspired by selectmenu where you can either use the component directly or slot your own if you wanted/needed more functionality.

examples/control-elements/media-chrome-menu-button.html
gkatsev: This component in particular definitely showed me that we do need default styles.

heff:
> Definitely something to consider once it's in the player. Theoretically, container queries could help since it can give us a size based on both the block and inline sizes of the container. Plus, there's the breakpoint stuff that's coming in themes.

The other approach, which may have implications for this button or another form of it, is triggering a menu that's not part of the button. 
* Youtube positions the settings menu right-bottom of the screen (above the control bar), which makes it look connected to the settings button when it's actually not
* Apple TV interfaces open a menu that that's positioned at the top center of the screen.

Both are creative ways of getting around the complications of responsiveness with menus connected to buttons. And come with tradeoffs. Just noting it for exploration, whatever the next step here is.

gkatsev:
the menu itself is a sibling of the button, so, should be easier to style.
Also, I realized that Container Queries' units might be a better here than CQs directly.

gkatsev:
But I guess depending on the needs, this component wouldn't be used directly and separate button and listbox components would be used.

PR #396, opened 2022-12-15
fix: mangled globals for IIFE bundle

### LINKED ISSUE #252: MediaController Safari conflict + minify naming bug
Found out legacy MediaController actually is still alive and kicking in Safari 😮 
Never really knew what it did, pretty cool if it works like intended. Syncing time on multiple media elements.

Open link in Safari
https://codesandbox.io/s/media-controller-safari-9xjr02

This would have resulted in a naming conflict with the Media Chrome MediaController due to 

```js
window[element.name] = element;
```

but there's also a little bug at least with the minified bundles where the class naming gets mangled so these all are added to the window with single letters like `window.A`, `window.d`, etc 

I think it's best we wouldn't override `window.MediaController`, maybe add `Element` in the end after fixing the mangling naming bug.

luwes:
fixes #252

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "build:types": "tsc",
     "build:esm": "esbuild src/js/*.js src/js/*/*.js src/js/*/*/*.js --target=es2019 --format=esm --outdir=dist",
     "build:cjs": "esbuild src/js/*.js src/js/*/*.js src/js/*/*/*.js --target=es2019 --format=cjs --outdir=dist/cjs",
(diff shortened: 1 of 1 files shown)


gkatsev:
This seems reasonable to me, but I think this does bring up whether we should be making the elements accessible this way on `window`.
Potentially, we should have all the elements be available from the main import or if they want the element, users can always get it from the `customElements` registry? Most probably would be using it directly in the HTML and never need these references.

PR #398, opened 2022-12-16
feat: container-size attr w/ range naming

luwes:
test url: https://media-chrome-git-fork-luwes-resize-observer-mux.vercel.app/examples/responsive-attribute.html

- removed the media-container-size attribute
- added attributes for the breakpoints: `breakpoint-sm`, `breakpoint-md`, `breakpoint-lg` and `breakpoint-xl`
- added attribute `breakpoints` for customizing breakpoints. e.g. `breakpoints="sm:300 md:400 lg:500"`

### DIFF
--- a/examples/responsive-attribute.html
+++ b/examples/responsive-attribute.html
@@ -11,7 +11,7 @@
 
       <p>This is an example of making Media Chrome responsive using the <code>media-container-size</code> attribute.</p>
 
-      <media-controller containerbreakpoints="s:0 sm:384 m:576 ml:960 l:1280 xl:1600">
+      <media-controller containerbreakpoints="xs:0 sm:384 md:576 lg:768 xl:960">
         <video
           slot="media"
           src="https://stream.mux.com/DS00Spx1CV902MCtPj5WknGlR102V5HFkDe/high.mp4"
@@ -87,12 +87,11 @@ video {
  * It only shows the center controls and hides the bottom control bar.
  *
  */
-
-media-controller[media-container-size=s] .centered-controls-overlay {
+[media-container-size="xs"] .centered-controls-overlay {
   display: flex;
 }
 
-media-controller[media-container-size=s] media-control-bar {
+[media-container-size="xs"] media-control-bar {
   display: none;
 }
 
@@ -104,15 +103,15 @@ media-controller[media-container-size=s] media-control-bar {
  * room for the other buttons in the control bar.
  *
  */
-media-controller[media-container-size$=m] .centered-controls-overlay {
+[media-container-size="xs sm"] .centered-controls-overlay {
   display: flex;
 }
 
-media-controller[media-container-size$=m] media-control-bar {
+[media-container-size="xs sm"] media-control-bar {
   display: flex;
 }
 
-media-controller[media-container-size$=m] media-control-bar :is(
+[media-container-size="xs sm"] media-control-bar :is(
   media-play-button,
   media-seek-backward-button,
   media-seek-forward-button
@@ -122,18 +121,18 @@ media-controller[media-container-size$=m] media-control-bar :is(
 }
 
 /*
- * The default and large container size,
+ * The default and larger container size,
  * which will be selected when the container is greater than 576px wide.
  *
  * Since we have a default to hide the .centered-controls-overlay,
  * this query isn't strictly necessary but is included for the sake of completeness.
  *
  */
-media-controller[media-container-size*=l] .centered-controls-overlay {
+[media-container-size="xs sm md"] .centered-controls-overlay {
   display: none;
 }
 
-media-controller[media-container-size*=l] media-control-bar {
+[media-container-size="xs sm md"] media-control-bar {
   display: flex;
 }
 
@@ -194,6 +193,7 @@ style {
   white-space: pre;
   font-family: monospace;
   margin-left: 1em;
+  overflow-x: auto;
 }
       </style>
       <div class="examples">
--- a/examples/responsive-attribute.html
+++ b/examples/responsive-attribute.html
@@ -82,7 +82,7 @@ video {
 
 /*
  * Starting with the smallest container, this query will be selected
- * if the inline-size of the container is smaller than 396px.
+ * if the inline-size of the container is smaller than 384px.
  *
  * It only shows the center controls and hides the bottom control bar.
  *
@@ -97,7 +97,7 @@ media-controller[media-container-size=s] media-control-bar {
 }
 
 /*
- * The middle container size will be selected if the width is between 396px and 576px.
+ * The middle container size will be selected if the width is between 384px and 576px.
  *
  * Here, move the player toggle and the seek buttons from the bottom
  * control bar to the center controls section to give more breathing
--- a/examples/responsive-attribute.html
+++ b/examples/responsive-attribute.html
@@ -11,7 +11,7 @@
 
       <p>This is an example of making Media Chrome responsive using the <code>media-container-size</code> attribute.</p>
 
-      <media-controller containerbreakpoints="xs:0 sm:384 md:576 lg:768 xl:960">
+      <media-controller breakpoints="sm:384 md:576 lg:768 xl:960">
         <video
           slot="media"
           src="https://stream.mux.com/DS00Spx1CV902MCtPj5WknGlR102V5HFkDe/high.mp4"
@@ -87,11 +87,11 @@ video {
  * It only shows the center controls and hides the bottom control bar.
  *
  */
-[media-container-size="xs"] .centered-controls-overlay {
+.centered-controls-overlay {
   display: flex;
 }
 
-[media-container-size="xs"] media-control-bar {
+media-control-bar {
   display: none;
 }
 
@@ -103,21 +103,20 @@ video {
  * room for the other buttons in the control bar.
  *
  */
-[media-container-size="xs sm"] .centered-controls-overlay {
+[breakpoint-sm] .centered-controls-overlay {
   display: flex;
 }
 
-[media-container-size="xs sm"] media-control-bar {
+[breakpoint-sm] media-control-bar {
   display: flex;
 }
 
-[media-container-size="xs sm"] media-control-bar :is(
+[breakpoint-sm] media-control-bar :is(
   media-play-button,
   media-seek-backward-button,
   media-seek-forward-button
 ) {
-    display: none;
-  }
+  display: none;
 }
 
 /*
@@ -128,11 +127,11 @@ video {
  * this query isn't strictly necessary but is included for the sake of completeness.
  *
  */
-[media-container-size="xs sm md"] .centered-controls-overlay {
+[breakpoint-md] .centered-controls-overlay {
   display: none;
 }
 
-[media-container-size="xs sm md"] media-control-bar {
+[breakpoint-md] media-control-bar {
   display: flex;
 }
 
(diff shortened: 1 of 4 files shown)


src/js/media-container.js
```diff
@@ -145,6 +145,94 @@ template.innerHTML = `
...
+              container.mediaSetCallback(media)
+            );
+          }
+        });
+      }
+    }
+  }
+};
+
+const defaultBreakpoints = 's:0 sm:384 m:576 ml:960 l:1280 xl:1600 fhd:1920 qhd:2560 uhd:3840';
```
gkatsev, line 201: Proposal: `media-container-breakpoints` as the attribute to configure this

gkatsev, line 201: Though: it is a bit weird that these represent the minimum for the size. Though, I can see how that simplifies some things, like having the uhd match the 4k-uhd dimensions as the minimum and that it doesn't need to specify Infinity or something.
I think it's a bit easier to read these and reason about them if they are maximums, but then it leaves how to represent the last item, though, I guess could use Infinity directly.
```
s:384 sm:576 m:960 ml:1280 l:1600 xl:1920 fhd:2560 qhd:3840 uhd:Infinity
```
That said, I'm not sure the difference is big enough compared to the downsides.

luwes: one other solution which would also solve the multiple CSS selector issue is to be able to define ranges for each label but it's a little more verbose.

```
s:0:384 sm:384:576 smaller:0:479 bigger:480:1920
```

luwes: `container-breakpoints` or `containerbreakpoints` ?

gkatsev: It's probably fine as is, I'm not sure if the alternatives are any better. Just pointing out, that to me, it's a bit easier to know that the config is the maximum as it feels a bit different from setting the CSS directly or usage via bootstrap/tailwind.

gkatsev: we really need to come up with a style guide for media-chrome that lays out the options explicitly, so, we don't need to think about it each time.

src/js/constants.js
```diff
@@ -83,6 +83,7 @@ export const MediaUIAttributes = {
   MEDIA_LOADING: 'media-loading',
   MEDIA_BUFFERED: 'media-buffered',
   MEDIA_STREAM_TYPE: 'media-stream-type',
+  MEDIA_CONTAINER_SIZE: 'media-container-size'
```
cjpillsbury, line 86: question(blocking): Is this a `MediaUIAttribute`? It isn't modeling media state and it isn't designed to be shared with Media State Receivers (aka other Media Chrome Components). I think don't want to muddy that distinction here.

cjpillsbury (dismissed):
Aside from my one callout, I think this looks 😍 per our goals/prior discussions. Now that it's actually implemented, I do have some inchoate thoughts/concerns around mobile first + responsive design and unintentionally loosing UI CSS at different breakpoints, but I don't think that should hold us up here.

src/js/constants.js
luwes: good point, will change it

luwes: though, it is state derived from media-controller which is influenced by the media.

there's also `MediaStateReceiverAttributes`

```js
export const MediaStateReceiverAttributes = {
  MEDIA_CHROME_ATTRIBUTES: 'media-chrome-attributes',
  MEDIA_CONTROLLER: 'media-controller',
};
```

is it better this to be `containersize`? then it almost feels like one of the attributes the user controls

cjpillsbury: ah no that's a good point... I think the biggest concern is adding it to the `MediaUIAttribute` object, which is exclusively for state propagation stuff. I think we can keep the name but it probably should be in a different/new object (or just exported as its own const? defer to you on where it should be moved).

heff:
> I was looking for a naming pattern that overlapped in characters with sizes next to each other

This is really clever. I'm trying to understand more deeply how a user might use these. When you're doing responsive design you're starting with one design and tweaking the details as the page gets bigger/smaller. So the first question is do we think people will start with the small (mobile) design and go up from there, or start with the biggest version and go down from there. The general guidance for designers is start with the mobile size and widen the browser window until the page looks ugly and put the next breakpoint at that size. I'm not positive the same goes for video players.

I do think though that there's need for a css selector that is "true for this size and everything bigger than it", especially if we have detailed breakpoints like QHD, UHD. I don't expect most players will make layout changes above 1080p, or even 720p. So I think I question the value of selecting a narrow range like `*=m` does, versus the current breakpoint and everything above it (or everything below it, whichever is best for video players). I think that still leads to needing many repeated selectors.

I can see an option here where the container gets an attribute for each breakpoint, that survives as long as the width is greater than the breakpoint. For example if the breakpoint is "medium", "small" would still exist. I think bootstrap grids work like that. 
```
<media-container breakpoint-md breakpoint-sm>
``` 
Or go the other way, and if the breakpoint is "medium" everything bigger than it is also true. Either way you can capture larger ranges of width with a single selector. Thoughts on that?

dylanjha:
This is very neat @luwes. I think you found a clever naming convention that allows ranges by using the right CSS attr selectors.

Even though it's clever and neat, I am finding it a little confusing to follow the different syntaxes. I know it's just leveraging CSS [attribute selectors](https://www.w3schools.com/css/css_attribute_selectors.asp), but those selectors are kind of fringe (maybe not something developers commonly use?).

If it's going to be really common to have to use a mix of `=`, `$=`, and `*=` syntaxes in order to achieve the range naming behavior, then I tend to think we should just bake-in the range naming behavior more directly without making developers have the mental overhead of remembering "oh yeah `*=` matches all, and `$=` matches end, so which breakpoints would that apply to again?".

I agree with what @heff said:

> I do think though that there's need for a css selector that is "true for this size and everything bigger than it", especially if we have detailed breakpoints like QHD, UHD. I don't expect most players will make layout changes above 1080p, or even 720p.

> I can see an option here where the container gets an attribute for each breakpoint, that survives as long as the width is greater than the breakpoint. For example if the breakpoint is "medium", "small" would still exist. I think bootstrap grids work like that.

```html
<media-container breakpoint-md breakpoint-sm>
```
@heff is your idea to do _both_ ?

```html
<media-container breakpoint-sm breakpoint-md breakpoint="md">
```
I can't decide if that's good because it gives ultimate flexibility, or bad because it'll just be too confusing

luwes:
that makes sense to me, I tried to find a solution for having both that's why I ended up here.
in the last meeting about this we decided we would go for having the single breakpoint as attribute.

I'm fine with inclusive breakpoints.
how would it be in the template blocks then?

```html
<template block="if containerSize == 'sm md xl'">
```

or only the last one?

```html
<template block="if containerSize == 'xl'">
```

dylanjha:
I would vote for the first option in templates `<template block="if containerSize == 'sm md xl'">`, so this would apply to `sm`, `md` and `xl`. I think it's more clear

heff:
> @heff is your idea to do both ?

I wasn't thinking that, but open to it. That would allow you to target one width specifically when needed. But I'm not clear how necessary that is if I'm starting with mobile as the default then tweaking as it gets bigger. i.e. target "medium, but not large or small". Or maybe it's just "target mobile and not have it be the default".

If we do that I might choose a different attribute name to avoid confusion.

```
<media-container breakpoint-sm breakpoint-md containerSize="md">
```

So you can use "breakpoints" for traditional responsive design, and "size" for targeted changes.

> how would it be in the template blocks then?

With templates we're less limited than CSS, so we can do more things like:

```
<template block="if containerSize == sm">
<template block="if containerSize > sm">
<template block="if containerSize <= lg">
```

> containerSize

I think I like "size" for "width size" either way, because width will be the dominant dimension, but acknowledging that we're likely to have height-based responsiveness too.

luwes:
okay cool

what would the naming be of attributes that are set on the <media-controller> element? 
should these also be set on the `fullscreen-element` that @cjpillsbury is implementing?

should they be prefixed with `media-`?

`<media-controller media-breakpoint-sm media-breakpoint-md media-container-size="md">` ?

--- 
 
and the attribute to customize will be `container-breakpoints`? 

`<media-controller container-breakpoints="xs:0 sm:384 md:576 lg:768 xl:960">`

heff:
I see this as a unique feature of `<media-container>`, specifically. If you don't use media-container/controller, you don't get it. So for me that means...

> should these also be set on the fullscreen-element

No

> should they be prefixed with media-

No. Just `breakpoint`. 

> and the attribute to customize will be container-breakpoints?

Just `breakpoints`. Unless there's a reason I'm missing why it needs to be more verbose.

cjpillsbury:
> should these also be set on the `fullscreen-element` that @cjpillsbury is implementing?

I would hold off on this. I think we're going to eventually want a new "container-style" Media Chrome component that can facilitate these kinds of things, but, right now, `fullscreen-element` shouldn't be assumed as synonymous with "the root media chrome container for this media controller"

cjpillsbury:
> Just `breakpoints`. Unless there's a reason I'm missing why it needs to be more verbose.

@heff I suspect @luwes was being forward-looking, relating to conversations here, in my PR, and elsewhere about eventually wanting/needing a "container that isn't the media controller." That said, I think `breakpoints` will still make sense if/when we build this.

heff (approved):
Clean implementation. 👍

examples/responsive-attribute.html
```diff
@@ -0,0 +1,203 @@
...
+[breakpoint-sm] media-control-bar {
+  display: flex;
+}
+
+[breakpoint-sm] media-control-bar :is(
+  media-play-button,
+  media-seek-backward-button,
+  media-seek-forward-button
+) {
+  display: none;
```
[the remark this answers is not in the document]
luwes: it's intentional, it's modeled after @gkatsev 's responsive example 
https://github.com/muxinc/media-chrome/blob/5da8d58f22be5d74f126adc3aeab9cdecf1f9493/examples/responsive.html#L143-L160

src/js/media-container.js
```diff
@@ -141,6 +141,42 @@ template.innerHTML = `
 
 const MEDIA_UI_ATTRIBUTE_NAMES = Object.values(MediaUIAttributes);
 
+const defaultBreakpoints = 'sm:384 md:576 lg:768 xl:960';
```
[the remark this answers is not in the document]
luwes: it's been a while, I first used them in this doc 
https://github.com/muxinc/media-chrome/blob/e1ec5f36059cb74b101942f9c9a8a6ade38be1e2/docs/src/pages/en/responsive-controls.md#using-resize-observer

I think it was partly inspired by Bootstrap's lower breakpoints. 
960px is from the 960 grid I think, I used this as a breakpoint in other player's as well.

pretty sure I also looked at 
https://en.wikipedia.org/wiki/List_of_common_resolutions#/media/File:Vector_Video_Standards.svg

let me know if you like to research this a little bit.
just checked the Mux player they're quite a bit off at the moment.

```js
const SMALL_BREAKPOINT = 700;
const XSMALL_BREAKPOINT = 300;
const MediaChromeSizes = {
  LG: 'large',
  SM: 'small',
  XS: 'extra-small',
};```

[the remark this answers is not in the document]
luwes: here's it visualized

examples/responsive-attribute.html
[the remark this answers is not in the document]
heff: I might be looking at the wrong demo, but Gary's example has a play button for the large version.
image

[Wes's version doesn't.](https://media-chrome-git-fork-luwes-resize-observer-mux.vercel.app/examples/responsive-attribute.html)
image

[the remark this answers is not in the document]
luwes: my bad, I overlooked this. will fix this

PR #399, opened 2022-12-16
Listbox followups

gkatsev:
Implementing some of https://github.com/muxinc/media-chrome/issues/387

### DIFF
--- a/src/js/media-captions-menu-button.js
+++ b/src/js/media-captions-menu-button.js
@@ -26,7 +26,6 @@ template.innerHTML = `
   media-captions-listbox {
     position: absolute;
     bottom: 44px;
-    width: var(--media-captions-listbox-width, 150px);
     max-height: 300px;
     overflow: auto;
   }
--- a/src/js/media-captions-menu-button.js
+++ b/src/js/media-captions-menu-button.js
@@ -4,12 +4,12 @@ import { MediaUIAttributes, MediaStateReceiverAttributes } from './constants.js'
 import { window, document, } from './utils/server-safe-globals.js';
 
 const ccEnabledIcon = `
-<svg class="enabled" slot="button-content" aria-hidden="true" viewBox="0 0 26 24">
+<svg aria-hidden="true" viewBox="0 0 26 24">
   <path d="M22.83 5.68a2.58 2.58 0 0 0-2.3-2.5c-3.62-.24-11.44-.24-15.06 0a2.58 2.58 0 0 0-2.3 2.5c-.23 4.21-.23 8.43 0 12.64a2.58 2.58 0 0 0 2.3 2.5c3.62.24 11.44.24 15.06 0a2.58 2.58 0 0 0 2.3-2.5c.23-4.21.23-8.43 0-12.64Zm-11.39 9.45a3.07 3.07 0 0 1-1.91.57 3.06 3.06 0 0 1-2.34-1 3.75 3.75 0 0 1-.92-2.67 3.92 3.92 0 0 1 .92-2.77 3.18 3.18 0 0 1 2.43-1 2.94 2.94 0 0 1 2.13.78c.364.359.62.813.74 1.31l-1.43.35a1.49 1.49 0 0 0-1.51-1.17 1.61 1.61 0 0 0-1.29.58 2.79 2.79 0 0 0-.5 1.89 3 3 0 0 0 .49 1.93 1.61 1.61 0 0 0 1.27.58 1.48 1.48 0 0 0 1-.37 2.1 2.1 0 0 0 .59-1.14l1.4.44a3.23 3.23 0 0 1-1.07 1.69Zm7.22 0a3.07 3.07 0 0 1-1.91.57 3.06 3.06 0 0 1-2.34-1 3.75 3.75 0 0 1-.92-2.67 3.88 3.88 0 0 1 .93-2.77 3.14 3.14 0 0 1 2.42-1 3 3 0 0 1 2.16.82 2.8 2.8 0 0 1 .73 1.31l-1.43.35a1.49 1.49 0 0 0-1.51-1.21 1.61 1.61 0 0 0-1.29.58A2.79 2.79 0 0 0 15 12a3 3 0 0 0 .49 1.93 1.61 1.61 0 0 0 1.27.58 1.44 1.44 0 0 0 1-.37 2.1 2.1 0 0 0 .6-1.15l1.4.44a3.17 3.17 0 0 1-1.1 1.7Z"/>
 </svg>`;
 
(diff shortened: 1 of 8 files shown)


src/js/media-captions-menu-button.js
```diff
@@ -0,0 +1,161 @@
...
+   * if no captions selected, disabled icon should be shown
+   */
+  media-chrome-menu-button.enabled .enabled {
+  media-chrome-menu-button:not(.enabled) .disabled {
+    display: initial;
+  }
+</style>
+
+<media-chrome-menu-button aria-label="captions menu button">
+  <svg class="enabled" slot="button-content" aria-hidden="true" viewBox="0 0 26 24">
```
cjpillsbury, line 46: nitpick(non-blocking): Let's break these svgs out into their own strings to conform with our tacit convention in other buttons.

cjpillsbury (approved):
callout on the commented out css vars and one (non-blocking) nitpick. Otherwise 👌

cjpillsbury:
suggestion(non-blocking): Copy over the list of todos from the prior PR and indicate which were completed in the previous PR, which were completed in this PR, and which are still outstanding.

src/js/media-captions-menu-button.js
```diff
@@ -0,0 +1,184 @@
...
+  media-chrome-menu-button.enabled .enabled {
+  media-chrome-menu-button:not(.enabled) .disabled {
+    display: initial;
+  }
+</style>
+
+<media-chrome-menu-button aria-label="captions menu button">
+  ${ccEnabledIcon}
+  ${ccDisabledIcon}
+  <media-captions-listbox slot="listbox" media-controller="mc"></media-captions-listbox>
```
luwes, line 54: should this have a `media-controller` attr?

gkatsev: Good catch, I thought I addressed it already but I forgot.

It should get the captions-menu-button's `media-controller` if it's given, but should work otherwise. Definitely needs to get updated before this PR can be merged.

src/js/media-captions-menu-button.js
```diff
@@ -0,0 +1,184 @@
...
+    this.#updateMenuPosition();
+  }
+
+  #updateMenuPosition() {
+    const svgs = this.shadowRoot.querySelectorAll('svg');
+    const listbox = this.#menuButton.querySelector('media-captions-listbox');
+    const xOn = svgs[0].getBoundingClientRect().x;
+    const xOff = svgs[1].getBoundingClientRect().x;
+    const leftOffset = xOn > 0 ? xOn : xOff > 0 ? xOff : 0;
+    const xBody = /** @type {Document} */(document).body.getBoundingClientRect().x;
```
luwes, line 130: (blocking): doesn't this need to be the bounding box of the container or wrapper?

src/js/media-chrome-listbox.js
```diff
@@ -7,6 +7,28 @@ template.innerHTML = `
...
   :host ul {
     list-style: none;
+    display: flex;
+    flex-direction: column;
+    gap: 0.5em;
+    margin: 0;
+    padding: 0.5em;
+    /* background-color: var(--media-control-background, rgba(20,20,30, 0.7)); */
+    background-color: black;
+    color: #ffffff;
```
luwes, line 17: is this color overridable if a user sets the CSS via `media-captions-menu-button`?

src/js/media-chrome-listbox.js
```diff
@@ -7,6 +7,28 @@ template.innerHTML = `
 <style>
   :host ul {
     list-style: none;
+    display: flex;
+    flex-direction: column;
+    gap: 0.5em;
+    margin: 0;
+    padding: 0.5em;
+    /* background-color: var(--media-control-background, rgba(20,20,30, 0.7)); */
+    background-color: black;
```
luwes, line 16: don't forget to revert this to use the CSS var

src/js/media-captions-menu-button.js
gkatsev: you're right. Maybe `this.parentElement.getBoundingClientRect().x` or something like that should be done instead (with the appropriate null checks).

luwes: we did a similar thing for the time range preview thumbs. parent element is probably sufficient as first pass.
https://github.com/muxinc/media-chrome/blob/5c7acfbdd0998f6b18e8dc2c01006abfa8ce7b58/src/js/media-time-range.js#L353-L357

gkatsev: Ah, nice, I'll take a look there.

src/js/media-chrome-listbox.js
gkatsev: When I was first trying those other colors, there were issues around stacking transparencies and generally not looking great. However, definitely worth revisiting this.

gkatsev: Went back to this one, but tweaked the color of the selected item a bit so that contrast ratio between it and unselected items is higher.

src/js/media-chrome-listbox.js
gkatsev: I added a CSS variable. Otherwise, the default color outside media-controller won't work with the default background color.

src/js/media-captions-menu-button.js
```diff
@@ -0,0 +1,210 @@
...
+    vertical-align: middle;
+    max-width: 100%;
+    max-height: 100%;
+    min-width: 100%;
+  }
+</style>
+
+<media-chrome-menu-button aria-label="captions menu button">
+  <slot slot="button-content" class="enabled" name="on">${ccEnabledIcon}</slot>
+  <slot slot="button-content" class="disabled" name="off" hidden>${ccDisabledIcon}</slot>
```
heff, line 54: The magic of `<slot slot="">` 😄

src/js/media-chrome-listbox.js
```diff
@@ -7,6 +7,26 @@ template.innerHTML = `
 <style>
   :host ul {
     list-style: none;
+    display: flex;
```
heff, line 10: Thanks for adding default styles!
Any thoughts on making this inline-flex so it just fits the content?

gkatsev: that probably makes sense. I'll take a look and tweak styles as appropriate.

src/js/media-captions-menu-button.js
gkatsev: It's slots all the way down.

src/js/media-chrome-listbox.js
gkatsev: updated

cjpillsbury (approved):
Re-approved! 🚀

PR #400, opened 2022-12-16
fix: template syntax & add streamType to template props

luwes:
- adds `>`, `>=`, `<`, `<=` operators
- add the `media-stream-type` attribute to the available theme props as `streamType`

For a logical AND we decided to use nested `if` templates instead of supporting more complex expressions like in #383

### DIFF
--- a/src/js/media-theme-element.js
+++ b/src/js/media-theme-element.js
@@ -6,6 +6,13 @@ import { camelCase } from './utils/utils.js';
 // Export Template parts for players.
 export * from './utils/template-parts.js';
 
+const observedMediaAttributes = {
+  'media-stream-type': 'streamType'
+};
+
+/**
+ * @extends {HTMLElement}
+ */
 export class MediaThemeElement extends window.HTMLElement {
   static template;
   static observedAttributes = ['template'];
@@ -21,8 +28,12 @@ export class MediaThemeElement extends window.HTMLElement {
     this.renderRoot = this.attachShadow({ mode: 'open' });
 
     const observer = new MutationObserver(() => this.render());
-    // @ts-ignore
     observer.observe(this, { attributes: true });
+    observer.observe(this.renderRoot, {
+      attributeFilter: Object.keys(observedMediaAttributes),
+      attributeOldValue: true,
+      subtree: true,
+    })
 
     this.createRenderer();
   }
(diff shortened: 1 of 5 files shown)


src/js/media-theme-element.js
```diff
@@ -49,12 +61,18 @@ export class MediaThemeElement extends window.HTMLElement {
...
-    for (let attr of this.attributes) {
+    for (let attr of observedAttributes) {
+      const name = observedMediaAttributes[attr.name] ?? camelCase(attr.name);
       if (attr.value != null) {
-        props[camelCase(attr.name)] = attr.value === '' ? true : attr.value;
+        props[name] = attr.value === '' ? true : attr.value;
       } else {
-        props[camelCase(attr.name)] = false;
+        props[name] = false;
       }
```
luwes, line 76: ^^ changes of this PR

the code below here is from the other PR https://github.com/muxinc/media-chrome/pull/383

src/js/utils/template-processor.js
```diff
@@ -140,7 +144,7 @@ export function evaluateCondition(expr, state) {
     boolean: /true|false/,
     number: /-?\d+\.?\d*/,
     string: /(["'])((?:\\.|[^\\])*?)\1/,
-    operator: /[!=]=/,
+    operator: /[!=><]=|[><]/,
```
cjpillsbury, line 147: suggestion(non-blocking): Consider deriving this regex from the set of valid `conditions` for maintainability. E.g.:

```suggestion
    operator: new RegExp(Object.keys(conditions).join('|')),
```

test/unit/template-processor.spec.js
```diff
@@ -22,3 +29,57 @@ describe('template processor', () => {
...
+    assert.equal(instance.children[0].innerHTML, 'hello world!');
+  });
+
+  it('InnerTemplatePart: simple value if condition', async () => {
+    const template = document.createElement('template');
+    template.innerHTML = `<div>hello<template if="x == 'hello'"> world</template>!</div>`;
+    const instance = new TemplateInstance(template, { x: false }, processor);
+    assert.equal(instance.childNodes[0].innerHTML, 'hello!');
+
+    instance.update({ x: 'hell' });
```
cjpillsbury, line 62: 😂

PR #360, opened 2022-10-29
feat: winamp theme

wietsehage:
This is a draft PR for the MediaAmp (Winamp Classic) theme we are working on. This theme will be part of a 'how to make a Media Chrome theme' blog article.

### Todo

- [x] Find a way to have different ``--media-range-thumb`` styles for the time and volume range sliders.
- [x] Move all inline css into a global stylesheet.
- [x] Add VU meter.
- [x] Reusing the ‘visualizer display’ from the original Winamp to add a nice canvas for the video. (see: MilkDrop)
- [ ] Upgrade theme to be a portable, single file theme JS (see Demuxed 2022 theme).
- [ ] Find a way to hide the time indicator on hover.

PR #401, opened 2022-12-19
Winamp theme checks

luwes:
Original here https://github.com/muxinc/media-chrome/pull/360


examples/themes/winamp-theme.html
```diff
@@ -0,0 +1,426 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta name="viewport" content="width=device-width" />
+    <title>Media Chrome Winamp Theme Example</title>
+    <script type="module" src="../../dist/index.js"></script>
```
gkatsev, line 6: should this page duplicate media-theme-winamp.js? Or include it and then use the media-theme-winamp element?

gkatsev (approved):
Preemptive approval.

examples/themes/winamp-theme.html
luwes: ah yes good call, I'll change that so it uses the web comp

PR #414, opened 2023-01-05
test: add more tests for controller flow & add test coverage

luwes:


test/unit/media-container.spec.js
```diff
@@ -0,0 +1,51 @@
+import { spy } from 'sinon';
+import { fixture, assert, aTimeout } from '@open-wc/testing';
+import '../../src/js/index.js';
+
+describe('<media-container>', () => {
```
cjpillsbury, line 5: thought(non-blocking): We should probably either test the `<media-container>` directly or move these tests to the `<media-controller>` spec, since it is built on the assumption that `<media-controller/>` isn't overriding any of these behaviors. I'd be fine with either, and I don't think this should be treated as a blocker if you'd like to discuss further.

test/unit/media-controller.spec.js
```diff
@@ -0,0 +1,160 @@
+import { fixture, assert, aTimeout, waitUntil } from '@open-wc/testing';
+import { constants } from '../../src/js/index.js';
+
+const { MediaUIEvents } = constants;
+
+describe('<media-controller>', () => {
+
+  it('associates itself to observe for state receivers', async () => {
```
cjpillsbury, line 8: praise: great test and description.

test/unit/media-controller.spec.js
```diff
@@ -0,0 +1,160 @@
...
+
+    // Also includes the media-gesture-receiver by default
+    assert.equal(mediaController.mediaStateReceivers.length, 3);
+    assert(mediaController.mediaStateReceivers.indexOf(mediaController) >= 0, 'registers itself');
+    const div = mediaController.querySelector('div');
+    assert(mediaController.mediaStateReceivers.indexOf(div) >= 0, 'registers div');
+  });
+});
+
+describe('receiving state / dispatching (bubbling) events', () => {
```
cjpillsbury, line 70: suggestion(non-blocking): You can nest `describe()`s. Since these are "more specific `<media-controller>` tests. It might be useful to have them nested in the generic `media-controller` describe, above.

test/unit/media-controller.spec.js
```diff
@@ -0,0 +1,160 @@
...
+  let mediaController;
+  let video;
+  let div;
+
+  beforeEach(async () => {
+    mediaController = await fixture(`
+      <media-controller>
+        <video
+          slot="media"
+          src="https://stream.mux.com/A3VXy02VoUinw01pwyomEO3bHnG4P32xzV7u1j1FSzjNg/low.mp4"
```
cjpillsbury, line 80: praise: smart use of `low.mp4` here. Good attention to detail.

test/unit/media-controller.spec.js
```diff
@@ -0,0 +1,160 @@
...
+    assert(true, 'media-current-time is 2');
+  });
+
+  it('can change volume', async () => {
+    div.dispatchEvent(new CustomEvent(MediaUIEvents.MEDIA_VOLUME_REQUEST, {
+      detail: 0.73,
+      bubbles: true
+    }));
+
+    await waitUntil(() => mediaController.getAttribute('media-volume') == 0.73);
```
cjpillsbury, line 156: praise: 😎 good test structure.

test/unit/media-play-button.spec.js
```diff
@@ -4,7 +4,7 @@ import { MediaUIAttributes, MediaUIEvents } from '../../src/js/constants';
 import { verbs } from '../../src/js/labels/labels';
 import '../../src/js/media-play-button';
 
-describe('<media-play-button/>', () => {
+describe('<media-play-button>', () => {
```
cjpillsbury, line 7: I see you have a preference for our `<component-name>` 😄. Happy to conform unless other folks lean another way.

test/unit/template-parts.spec.js
```diff
@@ -0,0 +1,579 @@
...
+  const originalHTML = `Hello {{x}}!`;
+  template.innerHTML = originalHTML;
+  const instance = new TemplateInstance(template, { x: 'world' });
+  is(template.innerHTML, originalHTML);
+  const root = document.createElement('div');
+  root.appendChild(instance);
+  is(root.innerHTML, `Hello world!`);
+});
+
+test('create: can render nested text nodes', () => {
```
cjpillsbury, line 32: suggestion(non-blocking): When I read this, it seems to be about supporting rendering HTML + templated values. Consider rewording as something like
> create: can render HTML with templated text

test/unit/template-parts.spec.js
```diff
@@ -0,0 +1,579 @@
...
+    y: 'bar',
+    z: 'baz',
+  });
+  is(template.innerHTML, originalHTML);
+  const root = document.createElement('div');
+  root.appendChild(instance);
+  is(root.innerHTML, `<div class="my-foo-state bar">baz</div>`);
+});
+
+test('create: never writes mustache syntax into an instantiated template even if no state given', () => {
```
cjpillsbury, line 95: praise: good test to include. ❤️

test/unit/template-parts.spec.js
```diff
@@ -0,0 +1,579 @@
...
+  const instance = new TemplateInstance(template, {
+    x: 'foo',
+    y: 'bar',
+    z: 'baz',
+  });
+  is(template.innerHTML, originalHTML);
+  const root = document.createElement('div');
+  root.appendChild(instance);
+  is(root.innerHTML, `<div class="my-foo-state bar">baz</div>`);
+  instance.update({ x: 'bing', y: 'bong', z: 'quux' });
```
cjpillsbury, line 119: 😂 solid renames.

test/unit/template-parts.spec.js
```diff
@@ -0,0 +1,579 @@
...
+  is(root.innerHTML, `<div class="my-bing-state bong">quux</div>`);
+});
+
+test('update: performs noop when update() is called with partial args', () => {
```
cjpillsbury, line 123: praise: another really good test to include for a spec, especially since it's not "intuitive" that it would behave this way.

test/unit/template-parts.spec.js
```diff
@@ -0,0 +1,579 @@
+/* Adapted from https://github.com/dy/template-parts - ISC - Dmitry Iv. */
+
+import { assert } from '@open-wc/testing';
+import { TemplateInstance, AttrPart, AttrPartList } from '../../src/js/utils/template-parts.js';
+
+const test = it;
+const is = assert.deepEqual;
+const any = (a, b) => assert.include(b, a);
+
+test('create: applies data to templated text nodes', () => {
```
cjpillsbury, line 10: nitpick(blocking): Unless we expect to regularly update this via a copy-paste from the original `template-parts` repo, I'd recommend refactoring this to conform to our other test modules (aka wrap in a `describe` and use our standard assertions.

test/unit/template-parts.spec.js
```diff
@@ -0,0 +1,579 @@
...
+  const originalHTML = `<div class="my-{{y}}-state"></div>`;
+  template.innerHTML = originalHTML;
+  const instance = new TemplateInstance(template, { y: 'foo' });
+  is(template.innerHTML, originalHTML);
+  const root = document.createElement('div');
+  root.appendChild(instance);
+  is(root.innerHTML, `<div class="my-foo-state"></div>`);
+});
+
+test('create: can render into many values', () => {
```
cjpillsbury, line 65: suggestion(non-blocking): Add tests for "partial args" and "no args", but for the constructor, similar to `update` tests. Aka, using this test as an example, how do these behave:

- `new TemplateInstance(template, { x: 'foo' })`
- `new TemplateInstance(template)`

cjpillsbury (approved):
This is awesome. Mostly praise, with a few non-blocking nits/suggestions.

test/unit/media-controller.spec.js
luwes: yes, I intentionally didn't nest here. I find it easier to read

test/unit/media-play-button.spec.js
luwes: yes we mentioned this before I think, @gkatsev as well. 
custom elements don't have a self closing feature so it could confuse devs. for React components this is fine

test/unit/media-controller.spec.js
cjpillsbury: Easier to read the code or easier to read the results/output? I can see it being easier for the former, but probably not the latter? I'll add it to our eng sync discussion this week so we can just settle on one and stick to it.

test/unit/media-play-button.spec.js
cjpillsbury: works for me. I don't have strong opinions here either way.

PR #415, opened 2023-01-05
Feat/configurable fullscreen element

cjpillsbury:
This PR adds the ability to target something other than the `<media-controller/>` instance for fullscreen. It also adds an example that takes advantage of this.

TODO:
- [x] Update Docs

### DIFF
--- a/examples/standalone-controls.html
+++ b/examples/standalone-controls.html
@@ -26,7 +26,7 @@
       <h1>Standalone Controls Example</h1>
       <p>
         This example shows how individual controls can be used without the
-        <code>media-container</code> element.
+        <code>media-controller</code> element.
       </p>
 
       <media-controller id="controller">
--- a/examples/wrapper.html
+++ b/examples/wrapper.html
@@ -1,0 +1,116 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta name="viewport" content="width=device-width" />
+    <title>Media Chrome Standalone Controls Example</title>
+    <script type="module" src="../dist/index.js"></script>
+    <style>
+      /** add styles to prevent CLS (Cumulative Layout Shift) */
+      media-controller:not([audio]) {
+        aspect-ratio: 16 / 9; /* set container aspect ratio if preload=none */
+      }
+
+      #wrapper {
+        display: flex;
+        flex-direction: column;
+        width: 50vw;
+        align-items: stretch;
+      }
+
+      #wrapper > media-control-bar {
+        flex-grow: 0;
+        flex-shrink: 1;
+      }
+
+      #wrapper > .spacer {
+        background-color: var(--media-background-color, #000);
+        flex-grow: 1;
+      }
+
+      #wrapper > .fullscreen-overlay {
+        display: var(--fullscreen-overaly-display, none);
+        /* Can use fixed since we're fullscreen */
+        position: fixed;
+        top: 0;
+        left: 0;
+        right: 0;
+        bottom: 0;
+        pointer-events: none;
+        color: #ffffff;
+      }
+
+      #wrapper:is(:fullscreen, :-webkit-full-screen, :-moz-full-screen, :-ms-fullscreen) {
+        --fullscreen-overaly-display: 'initial';
+      }
+
+      video {
+        width: 100%; /* prevents video to expand beyond its container */
+      }
+
+      .examples {
+        margin-top: 20px;
+      }
+    </style>
+  </head>
+  <body>
+    <main>
+      <h1>Using a Wrapper Element with Media Chrome Example</h1>
+      <p>
+        This example shows how you can use a wrapper element for your Player UI
+        or "chrome". It takes advantage of:
+      </p>
+      <ul>
+        <li>
+          Using "Standalone controls" via the
+          <code>media-controller</code> attribute
+        </li>
+        <li>
+          Targetting the wrapper instead of the <code>media-controller</code> via the
+          <code>fullscreen-element</code> attribute
+        </li>
+        <li>
+          Using CSS selectors and CSS variables to show an overlay only when in fullscreen.
+        </li>
+      </ul>
+
+      <span id="wrapper">
+        <media-controller fullscreen-element="wrapper" id="controller">
+          <video
+            playsinline
+            slot="media"
+            src="https://stream.mux.com/DS00Spx1CV902MCtPj5WknGlR102V5HFkDe/high.mp4"
+          >
+            <track
+              label="English"
+              kind="captions"
+              srclang="en"
+              default
+              src="./vtt/en-cc.vtt"
+            />
+          </video>
+        </media-controller>
+        <div class="spacer"></div>
+        <media-control-bar media-controller="controller">
+          <media-play-button></media-play-button>
+          <media-mute-button></media-mute-button>
+          <media-seek-backward-button></media-seek-backward-button>
+          <media-seek-forward-button></media-seek-forward-button>
+          <media-volume-range></media-volume-range>
+          <media-time-range></media-time-range>
+          <media-time-display remaining show-duration></media-time-display>
+          <media-captions-button></media-captions-button>
+          <media-playback-rate-button></media-playback-rate-button>
+          <media-pip-button></media-pip-button>
+          <media-fullscreen-button></media-fullscreen-button>
+        </media-control-bar>
+        <span class="fullscreen-overlay">
+          <span>This is an overlay</span>
+        </span>
+      </span>
+
+      <div class="examples">
+        <a href="./">View more examples</a>
+      </div>
+    </main>
+  </body>
+</html>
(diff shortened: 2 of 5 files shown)


examples/standalone-controls.html
```diff
@@ -26,7 +26,7 @@
       <h1>Standalone Controls Example</h1>
       <p>
         This example shows how individual controls can be used without the
-        <code>media-container</code> element.
+        <code>media-controller</code> element.
```
cjpillsbury, line 29: This change is unrelated. Just noticed that this still referenced `media-container`.

src/js/media-controller.js
```diff
@@ -551,6 +552,17 @@ class MediaController extends MediaContainer {
...
+  get fullscreenElement() {
+    if (this.hasAttribute('fullscreen-element')) {
+      return document.getElementById(this.getAttribute('fullscreen-element'));
+    }
+    return this.#fullscreenElement ?? this;
+  }
+
+  set fullscreenElement(element) {
+    this.#fullscreenElement = element;
+  }
```
gkatsev, line 564: I would assume that if this property is set, it should unset the attribute/getter should check the private var before checking the attribute.

dylanjha (changes requested):
Missing `docs/` update for the new `fullscreen-element` attribute?

src/js/media-controller.js
```diff
@@ -551,6 +552,17 @@ class MediaController extends MediaContainer {
     this.enableHotkeys();
   }
 
+  get fullscreenElement() {
+    if (this.hasAttribute('fullscreen-element')) {
+      return document.getElementById(this.getAttribute('fullscreen-element'));
```
gkatsev, line 557: I think this is probably good enough for the majority usecase. If someone has nested WCs, they can use the setter.

cjpillsbury:
> Missing `docs/` update for the new `fullscreen-element` attribute?

lol added a todo to the description already (good timing).

src/js/media-controller.js
cjpillsbury: Yeah I held off on that for a couple of reasons:

1. I was trying to avoid @luwes's common "attribute sprouting" concerns here.
2. It wasn't obvious to me that if someone sets and then unsets the value that they'd want the `fullscreen-element` attribute to "disappear into the aether"

Originally I just had an `attributeChangedCallback` case that would explicitly set the `fullscreenElement`, but I ended up removing that to allow for the separation and precendence of the different values that I implemented here.

That said, I don't feel strongly about this impl, and it's easy enough to change. If we want to treat them as "synonymous", I think we would need to re-add my `attributeChangedCallback` logic and do what you're proposing.

examples/wrapper.html
```diff
@@ -0,0 +1,116 @@
...
+        top: 0;
+        left: 0;
+        right: 0;
+        bottom: 0;
+        pointer-events: none;
+        color: #ffffff;
+      }
+
+      #wrapper:is(:fullscreen, :-webkit-full-screen, :-moz-full-screen, :-ms-fullscreen) {
+        --fullscreen-overaly-display: 'initial';
```
luwes, line 43:
```suggestion
        --fullscreen-overlay-display: 'initial';
```

examples/wrapper.html
```diff
@@ -0,0 +1,116 @@
...
+        flex-shrink: 1;
+      }
+
+      #wrapper > .spacer {
+        background-color: var(--media-background-color, #000);
+        flex-grow: 1;
+      }
+
+      #wrapper > .fullscreen-overlay {
+        display: var(--fullscreen-overaly-display, none);
```
luwes, line 31:
```suggestion
        display: var(--fullscreen-overlay-display, none);
```

src/js/media-controller.js
```diff
@@ -48,6 +48,7 @@ class MediaController extends MediaContainer {
   }
 
   #hotKeys = new AttributeTokenList(this, 'hotkeys');
+  #fullscreenElement;
```
luwes, line 51: I have a little concern about the naming here. will fullscreen be the only feature that needs a settable element?

luwes: yes, came back to me. the resize observer will also have to be set to this element right?

examples/wrapper.html
cjpillsbury: whoops that should also not be a string (I make this mistake all the time with CSS, esp. CSS vars. @luwes Wanna update your suggestion to include fixing not one, but two of my silly mistakes?

luwes: updated

src/js/media-controller.js
cjpillsbury: @luwes Yeah I was struggling with a name here, too (names are hard). I definitely don't feel confident that there isn't a future world where there isn't either:
1. a single prop/attr that treats many things as the "media container substitute" (we've discussed some of these in the past, e.g. gestures and keypresses)
2. multiple props that are for setting elements for different purposes on a `media-controller`

Because of that, I didn't want to be overly presumptuous about the name right now (since, at least for case 1, that would be a breaking change). That said, I also don't like the specific name, so if you have an alternative proposal that doesn't assume too much of what it _**could**_ be used for in the future, I'm down.

src/js/media-controller.js
cjpillsbury: Yeah, at the very least, it feels like something that could be a followup.

src/js/media-controller.js
dylanjha: `#fullscreenElement` seemed okay to me since this is a narrow use case, but if we're looking for something that would also apply to the resize observer `#containerElement`?

Then again, there could be use cases where you want the fullscreenElement to be one thing, and some other element to be another thing and not be forced to make them the same thing. For that reason, I'd go with `#fullscreenElement`

luwes: true, it's hard to say at the moment. if the rumors are true we'll have at least pipElement soon too.

it's not it will be a breaking change I think. 

we could add containerElement later as well and make it clear that it's the element of the full application in any presentation mode.

src/js/media-controller.js
```diff
@@ -592,6 +604,13 @@ class MediaController extends MediaContainer {
         this.#hotKeys.value = newValue;
     } else if (attrName === 'default-stream-type') {
       this.propagateMediaState(MediaUIAttributes.MEDIA_STREAM_TYPE);
+    } else if (attrName === 'fullscreen-element') {
+      const el = newValue 
+        ? document.getElementById('fullscreen-element')
```
luwes, line 609:
```suggestion
        ? this.getRootNode()?.getElementById(newValue) 
```

- could be in a shadow DOM so best to get it from the shadow root where media controller is in 
- fixes the element id string

examples/wrapper.html
```diff
@@ -0,0 +1,116 @@
+<!DOCTYPE html>
```
heff, line 1: I think we're trying to push more things to docs rather than examples, right? Would this example fit under a guide? Can it be demoed in docs easily enough?

examples/wrapper.html
```diff
@@ -0,0 +1,116 @@
...
+<html lang="en">
+  <head>
+    <meta name="viewport" content="width=device-width" />
+    <title>Media Chrome Standalone Controls Example</title>
```
heff, line 5: New title needed

src/js/media-controller.js
```diff
@@ -592,6 +604,14 @@ class MediaController extends MediaContainer {
         this.#hotKeys.value = newValue;
     } else if (attrName === 'default-stream-type') {
       this.propagateMediaState(MediaUIAttributes.MEDIA_STREAM_TYPE);
+    } else if (attrName === 'fullscreen-element') {
```
heff, line 607: We've really gotta nail down this dash/no-dash thing 😆 (not a blocker)

src/js/media-controller.js
```diff
@@ -592,6 +604,14 @@ class MediaController extends MediaContainer {
...
+      const el = newValue 
+        // @ts-ignore
+        ? this.getRootNode()?.getElementById(newValue) 
+        : undefined;
```
heff, line 611: Non-blocker: I think `null` tends to be more common for "intentionally empty". We end up with a potential mix of undefined and null negative values as is (e.g. `getElementById(notAnElement) == null`).

src/js/media-controller.js
```diff
@@ -592,6 +604,14 @@ class MediaController extends MediaContainer {
...
+      // NOTE: Setting the internal private prop here to not
+      // clear the attribute that was just set (CJP).
+      this.#fullscreenElement = el;
```
heff, line 614: Probably overkill, but I'm wondering if we should exit fullscreen on any existing real element when changing this value.

heff (approved):
Love it. 👍

dylanjha (approved):

src/js/media-controller.js
cjpillsbury: I think it should be fine and I kinda like that this doesn't have any direct visible side effects. I know I'd be annoyed if I had to worry about my programmatic workflow unavoidably having a user experience impact.

PR #419, opened 2023-01-06
doc: Manual readme version bump to 0.17

heff:
Also added more specific versions for a few projects where needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Visit the [official documentation for Media Chrome](https://media-chrome.org/en/
 ## Video Example
 
 ```html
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "media-chrome",
-  "version": "0.16.3",
+  "version": "0.17.0",
```
heff, line 3: Shouldn't this have happened as part of the conventional CD?

cjpillsbury (approved):
🚢 🇮🇹

luwes (changes requested):
use tag commits

luwes (approved):

PR #425, opened 2023-01-10
feat: add audio, targetLiveWindow and breakpointX template params

luwes:
this adds the breakpoint attributes to the template params that this should work:

```html
<template if="breakpointSm">
  <template if="breakpointMd == null">....
```

### DIFF
--- a/src/js/media-theme-element.js
+++ b/src/js/media-theme-element.js
@@ -27,11 +27,30 @@ export class MediaThemeElement extends window.HTMLElement {
     super();
     this.renderRoot = this.attachShadow({ mode: 'open' });
 
-    const observer = new MutationObserver(() => this.render());
+    const observer = new MutationObserver((mutationList) => {
+      if (mutationList.some((mutation) => {
+        if (mutation.type !== 'attributes') return false;
+
+        // Render on each attribute change of the `<media-theme>` element.
+        if (this.hasAttribute(mutation.attributeName)) return true;
+
+        // Render if this attribute is directly observed.
+        if (observedMediaAttributes[mutation.attributeName]) return true;
+
+        // Render if `breakpoint-x` attributes change.
+        if (mutation.attributeName.startsWith('breakpoint-')) return true;
+
+      })) {
+        this.render();
+      }
+    });
+
+    // Observe the `<media-theme>` element for attribute changes.
     observer.observe(this, { attributes: true });
+
+    // Observe the subtree of the render root, by default the elements in the shadow dom.
     observer.observe(this.renderRoot, {
-      attributeFilter: Object.keys(observedMediaAttributes),
-      attributeOldValue: true,
+      attributes: true,
       subtree: true,
     })
 
@@ -63,7 +82,9 @@ export class MediaThemeElement extends window.HTMLElement {
     const observedAttributes = [
       ...Array.from(this.attributes),
       ...Array.from(this.mediaController?.attributes ?? [])
-        .filter(({ name }) => observedMediaAttributes[name])
+        .filter(({ name }) => {
+          return observedMediaAttributes[name] || name.startsWith('breakpoint-')
+        })
     ];
 
     const props = {};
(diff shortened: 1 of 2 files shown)


src/js/media-container.js
```diff
@@ -278,6 +277,25 @@ class MediaContainer extends window.HTMLElement {
...
+  }
+
+  set breakpoints(val) {
+    if (val === this.breakpoints) return;
+
+    this.setAttribute('breakpoints', val);
+  }
+
+  get breakpoints() {
+    return this.getAttribute('breakpoints') ?? 'sm:384 md:576 lg:768 xl:960';
```
heff, line 296: I'm not sure if we need this property, but if we have it I'd expect something more helpful than the unsplit string. Probably an object. The setter could still receive either a string or an object.

heff (commented):
Looks good but a suggestion on the breakpoints value.

src/js/media-container.js
luwes: yes, maybe I was a bit quick to add this property.
generally you'd like to reflect props / attrs but this will add unneeded complexity. I will remove it.

heff: I'm on the same page and I don't have anything against adding it, it's just up to you if it's worth the extra work right now.

src/js/media-theme-element.js
```diff
@@ -27,11 +29,30 @@ export class MediaThemeElement extends window.HTMLElement {
...
-    const observer = new MutationObserver(() => this.render());
+    const observer = new MutationObserver((mutationList) => {
+      if (mutationList.some((mutation) => {
+        if (mutation.type !== 'attributes') return false;
+
+        // Render on each attribute change of the `<media-theme>` element.
+        if (this.hasAttribute(mutation.attributeName)) return true;
+
+        // Render if this attribute is directly observed.
+        if (observedMediaAttributes[mutation.attributeName]) return true;
```
cjpillsbury, line 40: question(blocking): Should we filter down to only monitor based on specific [`target`s](https://developer.mozilla.org/en-US/docs/Web/API/MutationRecord/target)?

luwes: good point! I'll add a check for `media-controller` only then.

src/js/media-theme-element.js
```diff
@@ -27,11 +29,34 @@ export class MediaThemeElement extends window.HTMLElement {
...
-    const observer = new MutationObserver(() => this.render());
+    const observer = new MutationObserver((mutationList) => {
+      if (mutationList.some((mutation) => {
+        const target = /** @type HTMLElement */ (mutation.target);
+
+        // Render on each attribute change of the `<media-theme(-x)>` element.
+        if (this.hasAttribute(mutation.attributeName)) return true;
+
+        // Only check `<media-controller>`'s attributes below.
+        if (target.localName !== 'media-controller') return false;
```
cjpillsbury, line 40: suggestion(non-blocking): As long as we're not worried about the module ref, consider `target instanceof MediaController` instead

cjpillsbury (approved):
😎

src/js/media-theme-element.js
```diff
@@ -7,7 +7,9 @@ import { camelCase } from './utils/utils.js';
 export * from './utils/template-parts.js';
 
 const observedMediaAttributes = {
+  'media-target-live-window': 'targetLiveWindow',
```
heff, line 10: Is this just a placeholder for a future media state?

src/js/media-theme-element.js
```diff
@@ -7,7 +7,9 @@ import { camelCase } from './utils/utils.js';
...
   'media-stream-type': 'streamType',
+  'audio': 'audio',
```
heff, line 12: This attr might deserve a rename/rethink. With media container it sets the _layout_ to audio, no matter what the media. i.e. you may want the audio layout even though you're playing a video.

src/js/media-theme-element.js
luwes: yes, I can set it manually in Mux player for the time being

src/js/media-theme-element.js
luwes: good call, in Mux player this is ok but yes can see we'd want the source type in some situations

PR #426, opened 2023-01-11
Listbox followups 2

gkatsev:
Builds on top of https://github.com/muxinc/media-chrome/pull/399. Currently, only 2 commits are new: c468584031fd95f0c3e676b99fd45f6d8fdcc5f0 for hover styles and 47698dcdebc1b2bcec2ba62774eb98b14c148312 the current WIP for keeping the menu within the bounds of the player if it's too close to the right edge.


luwes:
(blocking for full captions menu feature):
- font-family needs to be a sans-serif font
- on mouse hover the cursor should be a pointer or something other than text selection

nit (non-blocking):
- I think it would be more satisfying to see my selected menu item for 100ms or so before the menu closes
- can I add a before icon to the active menu item? like a check mark

src/js/media-captions-menu-button.js
```diff
@@ -149,9 +154,19 @@ class MediaCaptionsMenuButton extends window.HTMLElement {
...
-      this.#listbox.style.marginLeft = `calc(${leftOffset}px - ${parentOffset}px - 10px)`;
+      // Get the element that enforces the bounds for the time range boxes.
+      const bounds =
+        (this.getAttribute('bounds')
+          ? closestComposedNode(this, `#${this.getAttribute('bounds')}`)
+          : this.parentElement) ?? this;
+      const parentOffset = bounds.getBoundingClientRect().x;
+
+      if (this.#listbox.offsetWidth + leftOffset > bounds.offsetWidth) {
+        this.#listbox.style.translate = (bounds.offsetWidth - this.#listbox.offsetWidth) + 'px'
```
luwes, line 166:
```suggestion
        this.#listbox.style.translate = (bounds.offsetWidth - this.#listbox.offsetWidth) + 'px';
```

luwes (approved):
some reminder comments for the DoD but LGTM

gkatsev:
Thanks for the review! Those make sense for blockers. Also, Christian asked that I change how toggling the previously selected caption works, so, this PR is also blocked on that.

Animating showing and hiding is something that I was thinking about, but haven't gotten around to yet.

As for modifying the menu, that probably isn't possible right now. I could add parts to things. Alternatively, users could use the menu-button itself.

gkatsev:
Decided to drop the captions-button toggle as well as adding the captions-menu-button to the build from this PR to unblock it. This is particularly due to the issues around being able to disable the button. But it'll be the next things I'm working on.

src/js/media-chrome-listitem.js
```diff
@@ -10,6 +10,7 @@ template.innerHTML = `
     line-height: 1em;
     padding: 0.5em;
     margin: 0em;
+    cursor: pointer;
```
cjpillsbury, line 13: thought(non-blocking): I'm always torn by this one, since I like the cursor to let me know something is clickable but I at least sometimes want to copy/paste some text that happens to be clickable.

cjpillsbury (approved):
👍

PR #427, opened 2023-01-11
Spike: Support smushedcase attrs

heff:
Exploring an option to also support the smushedcase (lowercase, no spaces) version of dashed attribute names (`media-paused`), for the sake of React and other frameworks that plan to support custom element attributes with a simple translation of property names to attribute names that won't support dashes in the name.

My first impression is that while this isn't super difficult, it's a decent amount of overhead for anyone trying to contribute a new component to the ecosystem.

My vote is we don't try to support both at the same time. 

To build on that, I think we ignore future React needs in the attribute naming conversation. That doesn't mean we won't choose smushed case, but just not because of React.
* React envs everywhere won't support web components fully for a while, and how the support looks could evolve
* There are other reasons why we want React-specific versions of media chrome components, including just feeling first-class in the react ecosystem, which is less friendly to web components compared to other frameworks
* If you do happen to use the web component version in React, it will still work after client-side hydration, whether you use the dashed version of the attribute (media-paused) or the property name (mediaPaused). But if you're using the web component version you should use it as documented.

The message being: "If you're using React, use the React components. If you're using the web component, use the HTML attributes as documented."

Thoughts?

### DIFF
--- a/src/js/media-live-button.js
+++ b/src/js/media-live-button.js
@@ -38,13 +38,37 @@ slotTemplate.innerHTML = `
   <slot name="spacer">&nbsp;</slot><slot name="text">LIVE</slot>
 `;
 
+function smushify(str){
+  return str.replaceAll('-', '').toLowerCase();
+}
+
+function addSmushifiedVersions(observedAttrs) {
+  // Need a mapping of smushed to dashed
+  // or need to manually handle each smushed attr
+  // in attributeChangedCallback
+  let smushedAttrs = {};
+
+  observedAttrs.forEach((item)=>{
+    const newItem = smushify(item);
+
+    if (!observedAttrs.includes(newItem)) { 
+      observedAttrs.push(newItem);
+      smushedAttrs[newItem] = item;
+    }
+  });
+
+  return [ observedAttrs, smushedAttrs ];
+}
+
+let [ observedAttrs, smushedAttrs ] = addSmushifiedVersions([
+  ...MediaChromeButton.observedAttributes,
+  MEDIA_PAUSED,
+  MEDIA_TIME_IS_LIVE
+]);
+
 class MediaLiveButton extends MediaChromeButton {
   static get observedAttributes() {
-    return [
-      ...super.observedAttributes,
-      MEDIA_PAUSED,
-      MEDIA_TIME_IS_LIVE
-    ];
+    return observedAttrs;
   }
 
   constructor(options = {}) {
@@ -53,7 +77,13 @@ class MediaLiveButton extends MediaChromeButton {
   }
 
   attributeChangedCallback(attrName, oldValue, newValue) {
-    super.attributeChangedCallback(attrName, oldValue, newValue);
+    if (smushedAttrs[attrName]) {
+      this.setAttribute(smushedAttrs[attrName], newValue);
+    } else {
+      // Assuming super classes may also convert smushed versions
+      // we don't want to pass through the smushed version
+      super.attributeChangedCallback(attrName, oldValue, newValue);
+    }
   }
 
   handleClick() {


cjpillsbury:
@heff & @muxinc/open-elements folks (and anyone else interested) - While I think the there are legitimate concerns/considerations for added complexity and code maintenance with simultaneously supporting "smushedcase" and "kebab-case" for our well, defined attributes, I want to make sure we aren't being overly "hand wavey" or "straw man" with the React side of the fence.

> * React envs everywhere won't support web components fully for a while, and how the support looks could evolve

For the assumptions on the "native Web" (W3C standards & specifications, browsers, etc.), we've had at least a "soft rule" of targeting evergreen/forward-looking standards & specs, including "un-finalized but unlikely to change drastically" standards, APIs, etc. So the question is, how analogous is this to React 19 (and beyond) + how it will support custom elements attributes + properties "out of box."

First, the planned solution is already in react@experimental. At this point, I think it's fairly unlikely that _**some**_ solution won't land in React 19.

Second, since landing in react@experimental, there have been 0 comments on issues or proposed changes in https://github.com/facebook/react/issues/11347. With that, plus the other motivations that led to the current planned implementation (e.g. rough consistency with Preact's solution alongside lessons learned with pain points for reasons to diverge, consistency with native HTML element behavior in ReactDOM), it also seems unlikely that this is going to change significantly before landing in React 19.

Finally, there may be a concern that, unlike the Web, React v19 is much less likely to be _**forward compatible**_ with future versions of React. However, particularly with "core features and functionality" plus JSX syntax, React has maintained forward compatibility since v1.0.

In other words, aside from the timing of React 19 being released and things like NextJS targetting React 19, I think this concern should be fairly low.

> * There are other reasons why we want React-specific versions of media chrome components, including just feeling first-class in the react ecosystem, which is less friendly to web components compared to other frameworks

I'd like to hear more on this concern. Is there any reason to think that writing JSX + JS/React code with our Media Chrome custom elements will feel less "idiomatically React" than JSX + JS/React code for, say, a `<video>` element? I'm asking this because React is currently the only framework where we have decided to maintain framework-specific permutations of our components, and that was only because of its poor support. Given our core team size and our general vision for building on top of web components, I think we should be aiming to drop the need to maintain "React-flavored components" if at all possible.

> * If you do happen to use the web component version in React, it will still work after client-side hydration, whether you use the dashed version of the attribute (media-paused) or the property name (mediaPaused). But if you're using the web component version you should use it as documented.

Not to focus too much on the specific example, but it's unlikely that we need to focus on `media-paused`, since the core use case assumes that attribute will be propagated via the `<media-controller>`, and so it won't be relevant to React considerations for that case, at least not for the foreseeable future. Instead, let's consider these two more plausible cases and imagine writing them in "idiomatic React", also assuming a single SSR and a subsequent initial client side render but _**not**_ supporting "smushedcase":

* **Example 1**: `playbackrates`

Idiomatic React JSX:

```jsx
<media-playback-rate playbackRates={[0.5, 0.75, 1, 1.5, 2]}/>
```

Render:

```html
<media-playback-rate></media-playback-rate>
```

In this case, at least for complex object values for props, we'd want to add _**properties**_ to our web components (I think we likely want to add property-equivalents for all well-defined attributes). If we do nothing, _**all**_ complex values will need to be converted to their string equivalents by the developer and it will not be "idiomatically React".

* **Example 2**: `default-stream-type`

Idiomatic React JSX:

```jsx
<media-controller defaultStreamType={StreamTypes.LIVE}>...</media-controller>
```

Render:

```html
<media-controller defaultstreamtype="live" ...other attrs >...</media-controller>
```

Since this is a "simple value" for a prop, the server-side render will set the attribute, and the client-side render will attempt to set the (again, hopefully existing) property. If we add the property but assume that it's going to "sprout" to an official kebab-case attribute, this will instead be:

```html
<media-controller defaultstreamtype="live" default-stream-type="live" ...other attrs, potentially duplicated>...</media-controller>
```

It's when we put the two together that it gets hairy, because we are effectively either coding _**ourselves**_ into a corner where we will have to maintain "react-flavored components", or forcing devs using web components to write non-idiomatic React.

> The message being: "If you're using React, use the React components. If you're using the web component, use the HTML attributes as documented."

tl;dr - as stated above, we should be trying to make it as viable as possible to deprecate and remove the React components ASAP.

dylanjha:
I think the point I'm getting stuck on is:

Instead of telling someone to write this:

```jsx
<media-controller defaultStreamType={StreamTypes.LIVE}>...</media-controller>
```

They have 2 other options:

```jsx
<MediaController defaultStreamType={StreamTypes.LIVE} />
```

OR

```jsx
<media-controller default-stream-type={StreamTypes.LIVE}>...</media-controller>
```

Those 2 other options (including the overhead of maintaining the React wrapper) feels better than the overhead of supporting both smushed + kebab

cjpillsbury:
> Those 2 other options (including the overhead of maintaining the React wrapper) feels better than the overhead of supporting both smushed + kebab

Any reason you think this? We can arguably deprioritize supporting both cases for any `MediaUIAttributes` (for the reasons discussed above) and can abstract away the other cases with some utility functions. If anything, this feels mostly like an "initial effort" hurdle and not a "maintenance hurdle", and simultaneously gives us a path forward for no longer needing any framework-specific components. Moreover, folks implementing their own components need not add this support, but we would also make it easier for them to ensure **_their_** components work with "idiomatic React"+JSX, which we have no answer for with our current react solution.

Also, per "above the fold" callout, this was more about the reasons listed about React. There may be _**other**_ reasons, including complexity, to not support _**both**_ smushedcase and kebab-case. This was specifically calling out the React-specific reasons _**why**_. I don't think those were especially good reasons, and your response doesn't touch on that. If those are _**not**_ good reasons, but we still aren't comfortable supporting both, that means the only reason to go whole hog with (pork?) kebab-case is legibility _**and**_ that effectively hog-ties (see what I did, there?) us to having to maintain React-flavored components (on both Media Chrome and Open Elements, since we're assuming where one goes, we all go).

> 
> > tl;dr - as stated above, we should be trying to make it as viable as possible to deprecate and remove the React components ASAP.
> 
> I'm not feeling the need to deprecate the React components ASAP

That may have been lost in translation, by ASAP I meant "As soon as _**possible**_". Since one of the primary reasons we chose Web Components as our core component architecture was to build on top of something that should work with any Web application framework, it feels like a reasonable goal to make that a reality, even for React, which means that should be at least informing our decisions on implementations here.

heff:
Good points all around... This decision sucks. (Also I'm on cough meds so forgive any lack of tact.)

I'm caving and leaning towards our worst of both worlds option of `media-smushedcase`. i.e. media states are prefixed with `media-`, everything else is smushed.

It at least makes media attrs slightly more readable. And while stand-alone components are the right design (yes, I still believe that Christian, and I look forward to bringing you the first example use), media states are a special case of attrs that deserve special documentation anyway (kind of like aria, which follows the same pattern)

I'm not excited about fighting any kind of ongoing battle here. Devs are just gonna copy/paste from the docs anyway, right? Do they really need to read the attr? (half serious)  If it was simply aesthetic that'd be one thing, but the more it feels like we're signing ourselves up for more work in one way or another, even if it's just more docs, the less readability feels like the priority.

Otherwise I'm on the same page that we should have matching props for attrs.

luwes:
Great comments, I still choose for consistency but I vote only smushedcase everything considering.

I would like Media Chrome to be as close to native as possible, so native attribute style can only make things easier in the long run. Removing the React wrapper is part of our vision 👍 

- Yes - smushedcase
- No - supporting both
- Yes - `media-` prefixed attribute states, hopefully can be replaced by custom state pseudo classes eventually

---

maybe another discussion but these should be smushed as well then IMO, they are not state attributes
https://github.com/muxinc/media-chrome/blob/51717890146d262a6e6843ad21ca0a29762521d8/src/js/constants.js#L54-L57

luwes:
can be closed for now?

PR #440, opened 2023-01-18
fix: Media Live Button style changes during live window

AdamJaggard:
- Makes the live indicator remain visually active while we are within the live window, as opposed to going grey immediately after pausing
- Shows disabled cursor for the button while within the live window, mimics old behaviour

@cjpillsbury can you double check these are as we discussed

### DIFF
--- a/src/js/media-live-button.js
+++ b/src/js/media-live-button.js
@@ -22,10 +22,14 @@ slotTemplate.innerHTML = `
     color: var(--media-live-button-icon-color, rgb(140, 140, 140));
   }
 
-  :host(:not([${MEDIA_PAUSED}])[${MEDIA_TIME_IS_LIVE}]) slot[name=indicator] > *,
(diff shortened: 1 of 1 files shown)


cjpillsbury (approved):
LGTM unless I'm missing something!

gkatsev (approved):

heff:
Disabled cursor 👍 

Gray on pause was intentional. @gkatsev called out before that this was how video.js worked by default, and it make sense for the other default workflow where we auto seek to live on unpause. We can add an option to make that behave differently when we introduce DVR.

Unless there's a good argument otherwise we should revert that part of this PR. And maybe it deserves a note in the CSS to make it more clear it's intentional as the default behavior.

heff:
Got the additional details on this one in Slack (reminder not to skimp on the details for why we're doing something in the PR).

When the video paused the red light would turn gray, but clicking the button was still disabled until the playhead was out of the live window. So the two behaviors were off in an unexpected way. 👍 We'll fix the intended behavior in a followup.

PR #442, opened 2023-01-19
Listbox followups 3

gkatsev:
- include captions menu button in the build
- propagate disabled attribute down to the button
- allow supplying a background color for the listbox that's separate from the control background via `--media-listbox-background`
- allow supplying an outline for hover state listitems via `--media-listbox-hover-outline`.

### DIFF
--- a/src/js/media-captions-listbox.js
+++ b/src/js/media-captions-listbox.js
@@ -2,6 +2,7 @@ import MediaChromeListbox from './media-chrome-listbox.js';
 import { window, document } from './utils/server-safe-globals.js';
 import { MediaUIAttributes, MediaUIEvents } from './constants.js';
 import { parseTextTracksStr, parseTextTrackStr, formatTextTrackObj } from './utils/captions.js';
+import { toggleSubsCaps } from './utils/captions.js';
 
 const captionsIndicatorInlineStyle = `
   fill: var(--media-icon-color, #eee);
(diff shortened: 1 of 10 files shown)


src/js/media-captions-menu-button.js
```diff
@@ -123,8 +132,45 @@ class MediaCaptionsMenuButton extends window.HTMLElement {
...
       }
     }
-  }
 
+    if (
+      this.hasAttribute('default-showing') && // we want to show captions by default
+    this.#enabledSlot.hidden // and we aren't currently showing them
+    ) {
+      // Make sure we're only checking against the relevant attributes based on whether or not we are using subtitles fallback
+      const subtitlesIncluded = !this.hasAttribute('no-subtitles-fallback');
```
luwes, line 141:
```suggestion
      const subtitlesIncluded = !this.hasAttribute('nosubtitlesfallback');
```

src/js/media-captions-menu-button.js
```diff
@@ -123,8 +132,45 @@ class MediaCaptionsMenuButton extends window.HTMLElement {
         this.disable();
       }
     }
-  }
 
+    if (
+      this.hasAttribute('default-showing') && // we want to show captions by default
+    this.#enabledSlot.hidden // and we aren't currently showing them
```
luwes, line 138:
```suggestion
      this.#enabledSlot.hidden // and we aren't currently showing them
```

src/js/media-captions-listbox.js
```diff
@@ -91,15 +94,15 @@ class MediaCaptionsListbox extends MediaChromeListbox {
       this.#render();
 
     } else if (attrName === MediaUIAttributes.MEDIA_SUBTITLES_SHOWING && oldValue !== newValue) {
-      const selectedTrack = parseTextTrackStr(newValue ?? '');
+      const selectedTrack = parseTextTracksStr(newValue ?? '')[0];
```
luwes, line 97: shame TS didn't catch this. I would think it should have with the props `.language` and `.label` later on

gkatsev: TS wouldn't catch this, since it's a behavioral thing.

src/js/media-captions-menu-button.js
gkatsev: Since this was a port from media-captions-button, I was hoping to leave as is to minimize discrepancy (and means you can swap easily swap between the two)

cjpillsbury: Yeah def appreciate the attention here, but for this case, I wouldn't treat this like a "new attribute" since it's already named/used elsewhere.

src/js/media-captions-menu-button.js
```diff
@@ -180,16 +226,22 @@ class MediaCaptionsMenuButton extends window.HTMLElement {
 
   enable() {
     this.addEventListener('click', this.#handleClick);
+    this.#menuButton.removeAttribute('disabled');
   }
   disable() {
     this.removeEventListener('click', this.#handleClick);
+    this.#menuButton.setAttribute('disabled', '');
```
cjpillsbury, line 233: suggestion(non-blocking): might want to hide/collapse the list on disabled if it's open.

cjpillsbury (approved):
🚢 🇮🇹

src/js/media-captions-menu-button.js
gkatsev: Going to bump this to subsequent listbox followups.

PR #448, opened 2023-01-23
fix: theme logic

luwes:
- [x] removes `audio` from template vars, this is set manually on the `<media-theme>` element so this was a mistake
- [x] fix the overwrite priority so media theme attributes set the by the consumer override the vars that are automatically injected.
- [ ] add inner template caching


src/js/utils/template-parts.js
```diff
@@ -378,3 +377,41 @@ export class InnerTemplatePart extends ChildNodePart {
...
+      // insert
+      else parent.insertBefore(bi, cur);
+    }
+
+    // remove tail
+    while (cur != end)
+      (next = cur.nextSibling), parent.removeChild(cur), (cur = next);
+  }
+
+  return b;
```
luwes, line 416: I put in this minimal node list differ because something was wrong with my remove/add logic and it left an element from a previous permutation in the DOM.

src/js/utils/template-parts.js
```diff
@@ -341,15 +341,14 @@ export class ChildNodePart extends Part {
...
+    if (!normalisedNodes.length) normalisedNodes.push(new Text());
 
-    this.#nodes[0].before(...normalisedNodes);
-
-    for (const oldNode of this.#nodes) {
-      if (!normalisedNodes.includes(oldNode)) oldNode.remove();
-    }
-
-    this.#nodes = normalisedNodes;
+    this.#nodes = swapdom(
```
cjpillsbury, line 346: praise: 😎

src/js/utils/template-parts.js
cjpillsbury: nice!

src/js/media-theme-element.js
```diff
@@ -86,11 +85,11 @@ export class MediaThemeElement extends window.HTMLElement {
 
   get props() {
     const observedAttributes = [
-      ...Array.from(this.attributes),
       ...Array.from(this.mediaController?.attributes ?? [])
         .filter(({ name }) => {
           return observedMediaAttributes[name] || name.startsWith('breakpoint-')
-        })
+        }),
+      ...Array.from(this.attributes),
```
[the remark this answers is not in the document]
luwes: yes indeed, I needed this behavior in Mux player so when `stream-type` is set manually it takes effect instantly and overrides the one coming from Media Chrome

cjpillsbury (approved):
🚢 🇮🇹

src/js/media-theme-element.js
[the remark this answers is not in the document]
cjpillsbury: Cool that sounds like the right contract/precedence imo.

PR #454, opened 2023-01-27
Etch-a-sketch theme

heff:
Pretty simple implementation.

+Had to fix the media-controller attribute in chrome-button so it can work inside a shadow dom. Should be applied (in a followup) to other components.

Goals for a more advanced one (later PRs):
* Make the knobs actually work
  * Left knob seeks
  * Right knob increases volume
* Add shine to the corners to make it look more real-ish
* On mobile, shake the device to start over (@dylanjha 's idea 😆)

image

### DIFF
--- a/src/js/media-chrome-button.js
+++ b/src/js/media-chrome-button.js
@@ -158,7 +158,7 @@ class MediaChromeButton extends window.HTMLElement {
       MediaStateReceiverAttributes.MEDIA_CONTROLLER
     );
     if (mediaControllerId) {
-      const mediaControllerEl = document.getElementById(mediaControllerId);
+      const mediaControllerEl = this.getRootNode().getElementById(mediaControllerId);
       mediaControllerEl?.associateElement?.(this);
     }
   }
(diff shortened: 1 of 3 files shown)


src/js/media-chrome-button.js
```diff
@@ -157,9 +157,16 @@ class MediaChromeButton extends window.HTMLElement {
     const mediaControllerId = this.getAttribute(
       MediaStateReceiverAttributes.MEDIA_CONTROLLER
     );
+
     if (mediaControllerId) {
-      const mediaControllerEl = document.getElementById(mediaControllerId);
-      mediaControllerEl?.associateElement?.(this);
+      const rootNode = this.getRootNode();
```
heff, line 162: Trying to remember if there's any specific downsides to getRootNode. The TS parser at least doesn't like it when trying to use properties on it, so it needs to be ignored.

luwes: we'll need to replace document.getElementById with this.getRootNode().getElementById for themes to work nicely with a decoupled media controller

luwes:
sorry @heff 😿. let's move it to https://github.com/muxinc/player.style

PR #457, opened 2023-01-27
fix: remove old MediaTheme element

luwes:
This removes the old MediaTheme element.

It's safe to remove now I believe, we renamed the HTML element `media-theme-base` and no one has reported anything.

Importing the new media theme element will be kept like the below snippet 

```html
    <script>
      import 'media-chrome';
      import 'media-chrome/dist/media-theme-element.js';

      // could also be an alias without /dist with pkg.exports
      // import 'media-chrome/media-theme-element';
    </script>
```


cjpillsbury (approved):
🩹 💪

PR #458, opened 2023-01-27
Add simple `media-viewer-counts` web component

jaredsmith:
This adds a simple `media-viewer-counts` web component, that makes it easy to 
display the currrent number of viewers of a particular video or live-stream.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,8 @@
     "start": "yarn dev",
     "test": "web-test-runner --coverage --config test/web-test-runner.config.js",
     "publish-release": "./scripts/publish.sh",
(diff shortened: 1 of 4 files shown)


luwes:
This is awesome! 

I might be wrong but I thought this would be more of a Mux specific web component so it would better fit as an element in https://github.com/muxinc/elements 

Maybe it could work if it was more generic where you could pass in different platform API urls.

cjpillsbury:
Closing this PR. Current efforts for an Open Elements mux-viewer-count MLP can be found here: https://github.com/jaredsmith/elements/tree/mux-viewer-count 🎉

PR #459, opened 2023-01-27
fix: decoupled controller in media-theme

luwes:
This change fixes an issue with using media controls that are not nested in `media-controller` but connect via the `id` and `media-controller` attribute in the context of a shadow DOM.

The bug is that `document.getElementById` doesn't search in shadow DOM.

The fix implemented here is to use `this.getRootNode().getElementById`. This only works when connected to the DOM.

- [x] add solution to other controller coupled elements when we agree on this method

### DIFF
--- a/src/js/media-time-display.js
+++ b/src/js/media-time-display.js
@@ -76,6 +76,7 @@ class MediaTimeDisplay extends MediaTextDisplay {
 
   disconnectedCallback() {
     this.disable();
+    super.disconnectedCallback();
   }
 
   attributeChangedCallback(attrName, oldValue, newValue) {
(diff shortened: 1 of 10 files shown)


gkatsev:
Using `getRootNode` sounds good to me. Though, I wonder if we want a helper method that's like more like `getAncestorById` in case there's multiple nested shadow doms?

luwes:
it crossed my mind as well, I'm leaning to leaving it out for now because ID's are unique per shadow realm and it seems to me how native HTML would work

src/js/media-time-display.js
```diff
@@ -76,6 +76,7 @@ class MediaTimeDisplay extends MediaTextDisplay {
 
   disconnectedCallback() {
     this.disable();
+    super.disconnectedCallback();
```
luwes, line 79: fixed an existing bug here with the help of a test, un-associating a state receiver

heff (approved):

gkatsev (approved):

PR #463, opened 2023-02-01
chore: rename theme files for shorter imports

luwes:
would make the imports a bit nicer knowing that themes are not bundled in the main bundle by default

```
import 'media-chrome/dist/themes/media-theme-minimal'
<mux-player theme="minimal"></mux-player>
```

vs

```
import 'media-chrome/dist/themes/minimal'
<mux-player theme="minimal"></mux-player>
```


gkatsev (approved):
Seems fine to me.

Additionally, I think it would be worth seeing if we can build the themes folder into the root dir, or use `package.json`'s `exports` so that you could do `import 'media-chrome/themes/netflix'` without the `dist`. But that can be a separate effort.

PR #468, opened 2023-02-02
fix: chrome range track colors

luwes:
this adds a new color `--media-range-track-color` which is the color that takes up the negative space of the track.
by default this is `transparent`

for time range it will be a linear color with stops for 
- `--media-range-bar-color`
- `--media-time-buffered-color`
- `--media-range-track-color`

does this still make sense naming wise?

use the rich diff for easy checking
https://github.com/muxinc/media-chrome/pull/468/files?short_path=989a9d6#diff-989a9d625d6cedfb292ec608fcb44bd86560ee8d5b3c5c142234ab32ace96406


dylanjha:
We might want to set [CodeCov to informational](https://docs.codecov.com/docs/commit-status#informational) -- for patch & project I think? To avoid Codecov blocking merges.

I *think* this would be the right config?

```yaml
coverage:
  status:
    project:
      default:
        informational: true
    patch:
      default:
        informational: true
```

gkatsev (approved):
I think the names do make sense. It might be a tad confusing with the usage of both `bar` and `track`, but I'm not sure if there's a much better option that doesn't also make it a lot more verbose. Like maybe `--media-range-progress-color`? But it isn't a lot better.

PR #469, opened 2023-02-03
feat: add Micro theme

luwes:
- [x] complete design specs
- [x] expanding volume range
- [x] control bar responsiveness
- [x] control bar grid placement
- [x] control bar direction
- [x] live & DVR

test url:
https://media-chrome-git-fork-luwes-micro-theme-mux.vercel.app/examples/themes/micro-theme.html


src/js/media-chrome-range.js
```diff
@@ -151,27 +159,6 @@ template.innerHTML = `
...
-     * set input to focus-visible, unless host-context is available (in chrome)
-     * in which case we can have the focus ring be on the host itself
-     */''}
-    :host-context([media-keyboard-control]):host(:focus),
-    :host-context([media-keyboard-control]):host(:focus-within) {
-      box-shadow: inset 0 0 0 2px rgba(27, 127, 204, 0.9);
-    }
-    :host-context([media-keyboard-control]) input[type=range]:focus-visible {
-      box-shadow: none;
-    }
```
luwes, line 164: I can be wrong but I believe this couldn't work if this control was not in the media-controller element.
more high level `:host-context` is probably not a good solution for loose coupled components which is one the goals of media-chrome.

I know the solution requires some extra JS but since `media-keyboard-control` is gotten from event listeners too seems this is a good enough solution listening for focusin / focusout events and checking .matches(':focus-visible')

src/js/media-live-button.js
```diff
@@ -35,10 +35,10 @@ slotTemplate.innerHTML = `
   </style>
 
   <slot name="indicator">${indicatorSVG}</slot>
-  <!--
```
gkatsev, line 38: I believe this is an HTML comment on purpose and would break things if it's removed. Though, it would be very nice if it wasn't needed anymore, which could also be the case. In which case, it can be removed altogether.

src/js/media-chrome-range.js
```diff
@@ -39,8 +38,8 @@ template.innerHTML = `
     :host {
       --thumb-height: var(--media-range-thumb-height, 10px);
       --track-height: var(--media-range-track-height, 4px);
-      --media-range-padding: var(--media-control-padding, 10px);
-      --_media-range-padding: var(--media-range-padding, 10px);
+      --_focus-box-shadow: var(--media-focus-box-shadow, inset 0 0 0 2px rgba(27, 127, 204, 0.9));
+      --_media-range-padding: var(--media-range-padding, var(--media-control-padding, 10px));
```
gkatsev, line 42: 👍🏻

src/js/media-chrome-range.js
gkatsev: I wonder if instead we want to move some of the media-container stuff like this into a base media-chrome element that everything extends (or a mixin or something).

However, yeah, your code does seem like it should work.

src/js/themes/media-theme-micro.js
```diff
@@ -0,0 +1,538 @@
...
+
+import { window, document } from '../utils/server-safe-globals.js';
+import { MediaThemeElement } from '../media-theme-element.js';
+
+const html = (raw, ...keys) => String.raw({ raw }, ...keys);
+
+const template = document.createElement('template');
+template.innerHTML = html`
+<style>
+  media-captions-button:not(:is([media-captions-list], [media-subtitles-list])) {
```
gkatsev, line 18: suggestion(non-blocking):
```suggestion
  media-captions-button:not(:is(media-captions-button[media-captions-list], media-captions-button[media-subtitles-list])) {
```

Might be better to repeat the `media-captions-button` selector. Otherwise, the `:is` will try to look through *all* elements.

src/js/themes/media-theme-micro.js
```diff
@@ -0,0 +1,538 @@
...
+</template>
+
+<template partial="MuteButton">
+  <media-mute-button part="mute button" disabled="{{disabled}}" aria-disabled="{{disabled}}">
+    <svg aria-hidden="true" viewBox="0 0 24 24" slot="high">
+      <path
+        d="m11.14 4.86-4 4a.49.49 0 0 1-.35.14H3.25a.25.25 0 0 0-.25.25v5.5a.25.25 0 0 0 .25.25h3.54a.49.49 0 0 1 .36.15l4 4a.5.5 0 0 0 .85-.36V5.21a.5.5 0 0 0-.86-.35Zm2.74-1.56v1.52A7.52 7.52 0 0 1 19.47 12a7.52 7.52 0 0 1-5.59 7.18v1.52A9 9 0 0 0 21 12a9 9 0 0 0-7.12-8.7Zm3.56 8.7a5.49 5.49 0 0 0-3.56-5.1v1.66a3.93 3.93 0 0 1 0 6.88v1.66a5.49 5.49 0 0 0 3.56-5.1Z"
+      />
+    </svg>
+    <svg aria-hidden="true" viewBox="0 0 24 24" slot="low">
```
gkatsev, line 323: nit: can low and medium swap, so they're ordered: high, medium, low, off?

gkatsev (commented):
Looks great!

src/js/media-live-button.js
luwes: it's about these 2 elements `<slot name="spacer">&nbsp;</slot><slot name="text">LIVE</slot>`
they shouldn't have a line break between them

removing the comment in runtime is fine I think

src/js/media-time-range.js
```diff
@@ -51,15 +51,27 @@ template.innerHTML = `
     }
 
     [part~="preview-box"] {
-      transition: visibility .25s, opacity .25s;
```
luwes, line 54: these changes below add the possibility to change the time range preview time/thumb transition, specifically for the start delay which Matt called out. it's a better experience if it has a small delay on mouse hover instead of showing up right away

cjpillsbury: suggestion(non-blocking): I think this is fine for now, but I think these at best have a non-trivial learning curve and at worst folks may break things when trying to use them (especially `--media-preview-transition-property`). We may want to do a followup to figure out a simpler way of modeling this stuff.

src/js/utils/template-processor.js
```diff
@@ -149,7 +149,11 @@ export const processor = {
         }
       } else {
         if (part instanceof AttrPart) {
-          part.booleanValue = false;
+          try {
+            part.booleanValue = false;
+          } catch {
+            // attribute can't be turned off, consists of multiple parts.
```
cjpillsbury, line 155: question(blocking): I'm assuming this was added/discovered as an edge case during implementation. If yes, can you describe the case where it happened? Should this just be a "silent catch" or should there be some visibility somehow if/when this case occurs?

src/js/themes/media-theme-micro.js
```diff
@@ -0,0 +1,538 @@
...
+    width: 5px;
+  }
+
+  media-live-button::after {
+    content: '';
+    width: 7px;
+  }
+
+  media-play-button {
+    display: var(--controls, var(--play-button, inline-flex));
```
cjpillsbury, line 227: suggestion(non-blocking): We may want to consider adding `--bottom-controls`, `--bottom-play-button`, etc. for predictability when folks swap out themes (not a high priority)

src/js/themes/media-theme-micro.js
```diff
@@ -0,0 +1,538 @@
...
+          {{>PipButton}}
+          {{>FullscreenButton}}
+        </div>
+      </media-control-bar>
+      {{>TimeRange}}
+
+    </template>
+
+  </template>
+
```
cjpillsbury, line 526: suggestion(non-blocking): We should strongly consider adding a slot for the mux player error dialog for feature parity.

cjpillsbury (changes requested):
Just one question blocker and a few non-blocking callouts. Overall, this looks 😍 and my smoke testing was 👌. Great job, @luwes

src/js/utils/template-processor.js
luwes: the issue was that the booleanValue throws when setting an attribute consisting of multiple attrparts

https://github.com/muxinc/media-chrome/blob/289611899aeb07c460a6c7261cf128f62fad7ffd/src/js/utils/template-parts.js#L288-L289

I changed it to use the regular `.value` which doesn't need this try / catch. 
I think this is better, silently failing might be unexpected.

cjpillsbury (approved):
Looks 👌

PR #471, opened 2023-02-07
Media chrome menu button improvements

gkatsev:


gkatsev:
So, I think this PR is probably good to go. @luwes managed to figure out the icon sizing issue, as well as simplify some stuff by having positioning be relative to the chrome-select-menu element itself.

The big changes here:
- rename media-chrome-menu-button to media-chrome-select-menu
- make media-captions-button behave specially when it's a child of media-chrome-select-menu
- make media-captions-menu-button to extend media-chrome-select-menu
  - this creates a media-captions-button and media-captions-listbox that are in the *lightdom* as opposed to the shadowom.
- sets `white-space:nowrap` on listitems of captions-listbox, with a css var to control it

A couple of questions that probably aren't blocked:
- Should media-captions-menu-button also be renamed?
- If you use the default listbox and button that are provided with media-chrome-select-menu, they are also hooked up via aria-haspopup attribute. Are we expecting users to set these up manually, or should media-chrome-select-menu set it up for them when they are slotted?

src/js/media-captions-button.js
```diff
@@ -112,6 +114,9 @@ class MediaCaptionsButton extends MediaChromeButton {
   }
 
   handleClick() {
+    // do nothing if parent is an instance of MediaChromeSelectMenu
+    if (this.parentElement instanceof MediaChromeSelectMenu) return;
```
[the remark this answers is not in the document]
gkatsev: It's needed because otherwise the button will do its default behavior, which we don't want in this case. Happy to do the check in some other way, such as getting the select-menu from the custom elements registry or something.

src/js/experimental/media-chrome-select-menu.js
```diff
@@ -78,14 +118,50 @@ class MediaChromeMenuButton extends window.HTMLElement {
...
+    this.#listbox.style.transform = `translateX(${position}px)`;
+  }
+
   #toggleExpanded() {
-    this.#button.setAttribute('aria-expanded', this.#expanded);
     this.#expanded = !this.#expanded;
+    this.#button.setAttribute('aria-expanded', this.#expanded);
   }
 
   enable() {
```
luwes, line 167: the lines below in enable / disable mutating `this.#button.handleClick` should also be removed I think.
we can't make assumptions this is a media-chrome button.
can make this work with preventing the click event on the button with adding an event listener w/ use capture and stopping propagation

src/js/media-captions-button.js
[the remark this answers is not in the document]
luwes: yes, we need to stop propagation of the click event but from outside this button, not inside

src/js/experimental/media-chrome-select-menu.js
gkatsev: Yeah, you're right, we definitely shouldn't assume that it's a media-chrome-button.

src/js/media-captions-button.js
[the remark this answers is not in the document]
gkatsev: Unfortunately, it won't work. The click listener here is registered earlier than the one in media-chrome-select-menu, so, even if we call `preventDefault` or `stopPropagation` or even `stopImmediatePropagation`, this method had already run.

src/js/experimental/media-chrome-select-menu.js
gkatsev: switched to an event handler.

src/js/media-captions-button.js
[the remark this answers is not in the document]
gkatsev: I updated this to use `window.customElements.get('media-chrome-select-menu')` rather than depend on it directly so that the other files won't get included if someone is importing just media-captions-button.

luwes (approved):
LGTM 👍

src/js/experimental/media-captions-selectmenu.js
```diff
@@ -0,0 +1,30 @@
+import MediaChromeSelectmenu from './media-chrome-selectmenu.js';
+import '../media-captions-button.js';
+import './media-captions-listbox.js';
+import { window, document, } from '../utils/server-safe-globals.js';
+
+class MediaCaptionsSelectmenu extends MediaChromeSelectmenu {
```
cjpillsbury, line 6: praise: love how tiny/simple this new impl is!

src/js/experimental/media-chrome-selectmenu.js
```diff
@@ -1,34 +1,47 @@
...
+  </slot>
   <slot name="listbox" hidden>
     <media-chrome-listbox id="listbox" part="listbox">
       <slot></slot>
     </media-chrome-listbox>
   </slot>
 `;
 
-class MediaChromeMenuButton extends window.HTMLElement {
+class MediaChromeSelectmenu extends window.HTMLElement {
```
cjpillsbury, line 38: thought(non-blocking): because nothing can be predictable 😞 https://chromium.googlesource.com/chromium/src/+/eaf93b7df0ff4c73760e56d40a54fc740841cd16/third_party/blink/renderer/core/html/forms/html_select_menu_element.h#23 casing is `HtmlSelectMenuElement`... just putting it on your radar. Defer to you on preferred casing for us.

src/js/experimental/media-chrome-selectmenu.js
```diff
@@ -78,26 +121,62 @@ class MediaChromeMenuButton extends window.HTMLElement {
...
     this.#expanded = !this.#expanded;
+    this.#button.setAttribute('aria-expanded', this.#expanded);
   }
 
   enable() {
     this.#button.removeAttribute('disabled');
-    this.#button.handleClick = this.#handleClick;
-    this.#toggleExpanded()
+    this.#button.addEventListener('click', this.#handleClick);
+    this.#toggleExpanded();
```
cjpillsbury, line 173: praise: nice catch 🦅 👀

src/js/experimental/media-chrome-selectmenu.js
```diff
@@ -55,6 +68,36 @@ class MediaChromeMenuButton extends window.HTMLElement {
...
+      if (this.#button.hasAttribute('disabled')) {
+        this.#enabledState = false;
+      }
+
+      // reconnect new button
+      if (this.#enabledState) {
+        this.enable();
+        this.#button.setAttribute('aria-haspopup', 'listbox');
+        // if it's a media-chrome-button, ask it to not handle the click event
+        this.#button.preventClick = true;
```
cjpillsbury, line 95: praise: 👌

cjpillsbury (approved):
One non-blocking callout and some praise. Great work 🏄‍♂️

src/js/experimental/media-chrome-selectmenu.js
gkatsev: why do they do this???

gkatsev: decided to follow suit.

PR #473, opened 2023-02-13
Feat/css vars spike

cjpillsbury:
Draft PR for discussion on CSS variables usage. None of these changes are set in stone. Rather, this is to facilitate discussion, including different tradeoffs between approaches. This example focuses on colors.

This is a non-exhaustive spike example of making the CSS variables more consistent/predictable along a few dimensions:
1. consistent/predictable existence/presence
2. consistent/predictable names
3. consistent/predictable application to css properties and components
4. consistent/predictable default values

I've also added comments to this PR to further facilitate discussion.

### DIFF
--- a/src/js/experimental/media-captions-listbox.js
+++ b/src/js/experimental/media-captions-listbox.js
@@ -6,7 +6,7 @@ import { parseTextTracksStr, formatTextTrackObj } from '../utils/captions.js';
 import { toggleSubsCaps } from '../utils/captions.js';
 
 const captionsIndicatorInlineStyle = `
-  fill: var(--media-icon-color, #eee);
+  fill: var(--media-control-icon-color, var(--media-primary-color, rgb(238 238 238)));
   height: var(--media-captions-indicator-height, 1em);
   vertical-align: var(--media-captions-indicator-vertical-align, bottom);
   margin-inline-start: 1ch;
--- a/src/js/experimental/media-chrome-listbox.js
+++ b/src/js/experimental/media-chrome-listbox.js
@@ -5,6 +5,11 @@ const template = document.createElement('template');
 
 template.innerHTML = `
 <style>
+  :host {
+    
+    
+  }
+
   :host ul {
     list-style: none;
     display: inline-flex;
@@ -13,7 +18,7 @@ template.innerHTML = `
     margin: 0;
     padding: 0.5em;
     background-color: var(--media-listbox-background, var(--media-control-background, rgba(10,10,15, .8)));
-    color: var(--media-text-color, white);
+    color: var(--media-control-font-color, var(--media-primary-color, rgb(238 238 238)));
     font-family: Arial, sans-serif;
   }
 
--- a/src/js/experimental/media-chrome-listbox.js
+++ b/src/js/experimental/media-chrome-listbox.js
@@ -23,11 +23,11 @@ template.innerHTML = `
   }
 
   ::slotted(media-chrome-listitem[aria-selected="true"]) {
-    background-color: var(--media-listbox-selected-background, rgba(122,122,184, .8));
+    background: var(--media-listbox-selected-background, rgb(122 122 184 / 0.7));
   }
 
   ::slotted(media-chrome-listitem:hover) {
-    background-color: var(--media-listbox-hover-background, rgba(82,82,122, .8));
+    background: var(--media-listbox-hover-background, rgb(82 82 122 / 0.8));
     outline: var(--media-listbox-hover-outline, none);
   }
 </style>
--- a/src/js/experimental/media-chrome-listbox.js
+++ b/src/js/experimental/media-chrome-listbox.js
@@ -5,11 +5,6 @@ const template = document.createElement('template');
 
 template.innerHTML = `
 <style>
-  :host {
-    
-    
-  }
-
   :host ul {
     list-style: none;
     display: inline-flex;
@@ -17,7 +12,7 @@ template.innerHTML = `
     gap: 0.5em;
     margin: 0;
     padding: 0.5em;
-    background-color: var(--media-listbox-background, var(--media-control-background, rgba(10,10,15, .8)));
+    background: var(--media-control-background, var(--media-secondary-color, rgb(20 20 30 / 0.7)));
     color: var(--media-control-font-color, var(--media-primary-color, rgb(238 238 238)));
     font-family: Arial, sans-serif;
   }
--- a/src/js/media-chrome-button.js
+++ b/src/js/media-chrome-button.js
@@ -6,9 +6,6 @@ const template = document.createElement('template');
 template.innerHTML = `
 <style>
   :host {
-    
-    
-
     display: inline-flex;
     align-items: center;
     justify-content: center;
--- a/src/js/media-chrome-button.js
+++ b/src/js/media-chrome-button.js
@@ -6,12 +6,15 @@ const template = document.createElement('template');
 template.innerHTML = `
 <style>
   :host {
+    
+    
+
     display: inline-flex;
     align-items: center;
     justify-content: center;
     vertical-align: middle;
     box-sizing: border-box;
-    background: var(--media-control-background, rgba(20,20,30, 0.7));
+    background: var(--media-control-background, var(--media-secondary-color, rgb(20 20 30 / 0.7)));
 
     padding: var(--media-control-padding, 10px);
 
@@ -19,7 +22,7 @@ template.innerHTML = `
     font-size: 14px;
     line-height: var(--media-text-content-height, var(--media-control-height, 24px));
     font-weight: bold;
-    color: #fff;
+    color: var(--media-control-font-color, var(--media-primary-color, rgb(238 238 238)));
 
     transition: background 0.15s linear;
 
@@ -53,7 +56,7 @@ template.innerHTML = `
     height: var(--media-button-icon-height, var(--media-control-height, 24px));
     transform: var(--media-button-icon-transform);
     transition: var(--media-button-icon-transition);
-    fill: var(--media-icon-color, #eee);
+    fill: var(--media-control-icon-color, var(--media-primary-color, rgb(238 238 238)));
     vertical-align: middle;
     max-width: 100%;
     max-height: 100%;
--- a/src/js/media-chrome-range.js
+++ b/src/js/media-chrome-range.js
@@ -11,7 +11,7 @@ const thumbStyles = `
   width: var(--media-range-thumb-width, 10px);
   border: var(--media-range-thumb-border, none);
   border-radius: var(--media-range-thumb-border-radius, 10px);
-  background: var(--media-range-thumb-background, #fff);
+  background: var(--media-range-thumb-background, var(--media-primary-color, rgb(238 238 238)));
   box-shadow: var(--media-range-thumb-box-shadow, 1px 1px 1px transparent);
   cursor: pointer;
   transition: var(--media-range-thumb-transition, none);
@@ -37,6 +37,9 @@ const trackStyles = `
 template.innerHTML = `
   <style>
     :host {
+      
+      
+
       --thumb-height: var(--media-range-thumb-height, 10px);
       --track-height: var(--media-range-track-height, 4px);
       --media-range-padding: var(--media-control-padding, 10px);
@@ -46,7 +49,7 @@ template.innerHTML = `
       box-sizing: border-box;
       display: inline-block;
       position: relative;
-      background: var(--media-control-background, rgba(20,20,30, 0.7));
+      background: var(--media-control-background, var(--media-secondary-color, rgb(20 20 30 / 0.7)));
       transition: background 0.15s linear;
       width: 100px;
       height: calc(var(--media-control-height, 24px) + 2 * var(--_media-range-padding));
@@ -315,7 +318,7 @@ class MediaChromeRange extends window.HTMLElement {
     }
 
     let colorArray = [
-      ['var(--media-range-bar-color, #fff)', rangePercent + thumbPercent],
+      ['var(--media-range-bar-color, var(--media-primary-color, rgb(238 238 238)))', rangePercent + thumbPercent],
       ['transparent', 100],
     ];
 
--- a/src/js/media-chrome-range.js
+++ b/src/js/media-chrome-range.js
@@ -37,9 +37,6 @@ const trackStyles = `
 template.innerHTML = `
   <style>
     :host {
-      
-      
-
       --thumb-height: var(--media-range-thumb-height, 10px);
       --track-height: var(--media-range-track-height, 4px);
       --media-range-padding: var(--media-control-padding, 10px);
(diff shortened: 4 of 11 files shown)


src/js/experimental/media-captions-listbox.js
```diff
@@ -6,7 +6,7 @@ import { parseTextTracksStr, formatTextTrackObj } from '../utils/captions.js';
 import { toggleSubsCaps } from '../utils/captions.js';
 
 const captionsIndicatorInlineStyle = `
-  fill: var(--media-icon-color, #eee);
+  fill: var(--media-control-icon-color, var(--media-primary-color, rgb(238 238 238)));
```
cjpillsbury, line 9: - Renaming `--media-icon-color` to `--media-control-icon-color` for consistency.
- Adding `--media-primary-color` for scoping and relationships of different "generic" vs. "specific" CSS vars.

src/js/experimental/media-chrome-listbox.js
```diff
@@ -12,8 +12,8 @@ template.innerHTML = `
     gap: 0.5em;
     margin: 0;
     padding: 0.5em;
-    background-color: var(--media-listbox-background, var(--media-control-background, rgba(10,10,15, .8)));
-    color: var(--media-text-color, white);
+    background: var(--media-control-background, var(--media-secondary-color, rgb(20 20 30 / 0.7)));
```
cjpillsbury, line 15: - Replacing `--media-listbox-background` with generic `--media-control-background` that conforms with other components
  - This includes changing the default value in the composition to be consistent with other components
  - Component-specific styling can be achieved by, e.g.:
```css
media-chrome-listbox {
   --media-control-background: rgb(10 10 15 / 0.8);
}
```
- May want to consider applying this background to the `media-chrome-listbox-item` instead, though this would require some additional changes to the default CSS

src/js/experimental/media-chrome-listbox.js
```diff
@@ -12,8 +12,8 @@ template.innerHTML = `
...
+    color: var(--media-control-font-color, var(--media-primary-color, rgb(238 238 238)));
```
cjpillsbury, line 16: - Renaming `--media-text-color` to `--media-controll-font-color` for consistency and replacing with consistent composition, including defaulting (which again changes the default value)

src/js/experimental/media-chrome-listbox.js
```diff
@@ -23,11 +23,11 @@ template.innerHTML = `
   }
 
   ::slotted(media-chrome-listitem[aria-selected="true"]) {
-    background-color: var(--media-listbox-selected-background, rgba(122,122,184, .8));
+    background: var(--media-listbox-selected-background, rgb(122 122 184 / 0.7));
```
cjpillsbury, line 26: - Changing to apply to `background` instead of `background-color` for consistency
- Changing default transparency/alpha to `0.7` for consistency
- Should consider renaming variable to have a more consistent naming convention.

src/js/experimental/media-chrome-listbox.js
```diff
@@ -23,11 +23,11 @@ template.innerHTML = `
...
   }
 
   ::slotted(media-chrome-listitem:hover) {
-    background-color: var(--media-listbox-hover-background, rgba(82,82,122, .8));
+    background: var(--media-listbox-hover-background, rgb(82 82 122 / 0.8));
```
cjpillsbury, line 30: Same as above.

src/js/media-chrome-button.js
```diff
@@ -11,15 +11,15 @@ template.innerHTML = `
...
+    background: var(--media-control-background, var(--media-secondary-color, rgb(20 20 30 / 0.7)));
 
     padding: var(--media-control-padding, 10px);
 
     ${/* Vertically center any text */''}
     font-size: 14px;
     line-height: var(--media-text-content-height, var(--media-control-height, 24px));
     font-weight: bold;
-    color: #fff;
+    color: var(--media-control-font-color, var(--media-primary-color, rgb(238 238 238)));
```
cjpillsbury, line 22: - Adding CSS variable for consistency, with standard composition, including defaults.

src/js/media-control-bar.js
```diff
@@ -14,7 +14,7 @@ template.innerHTML = `
       ${/* Need position to display above video for some reason */''}
       box-sizing: border-box;
       display: inline-flex;
-      color: var(--media-icon-color, #eee);
+      color: var(--media-control-font-color, var(--media-primary-color, rgb(238 238 238)));
       --media-loading-icon-width: 44px;
```
cjpillsbury, line 18: - Should consider adding `fill` for consistency

src/js/media-time-range.js
```diff
@@ -21,13 +21,9 @@ const template = document.createElement('template');
 template.innerHTML = `
   <style>
     :host {
-      --media-preview-background-color: rgba(20,20,30, .5);
```
cjpillsbury, line 24: - Removing values with defaults here and relying on well-defined compositions now defined in `media-chrome-range`.
  - This changes some default values here, which can be overridden if desired.

src/js/media-time-range.js
```diff
@@ -79,11 +75,10 @@ template.innerHTML = `
 
     media-preview-time-display,
     ::slotted(media-preview-time-display) {
-      color: unset;
```
cjpillsbury, line 82: - Removed for consistency

src/js/media-time-range.js
```diff
@@ -79,11 +75,10 @@ template.innerHTML = `
...
       min-width: 0;
       ${/* delay changing these CSS props until the preview box transition is ended */''}
       transition: min-width 0s .25s, border-radius 0s .25s;
-      background: var(--media-preview-time-background, var(--media-preview-background));
+      background: var(--media-preview-time-background, var(--media-preview-background, var(--media-control-background, var(--media-secondary-color, rgb(20 20 30 / 0.7)))));
```
cjpillsbury, line 81: - Same as comparable case, above

src/js/themes/media-theme-demuxed-2022.js
```diff
@@ -14,9 +14,13 @@ const template = document.createElement('template');
 template.innerHTML = `
 <style>
   :host {
-    --primary-color: black;
-    --secondary-color: transparent;
-    --tertiary-color: #7596CC;
+    --media-primary-color: black;
```
cjpillsbury, line 17: Example of using the new css vars

src/js/themes/media-theme-youtube.js
```diff
@@ -15,8 +15,8 @@ template.innerHTML = `
 <style>
   :host {
     display: inline-block;
-    --primary-color: #eee;
-    --secondary-color: transparent;
+    --media-secondary-color: transparent;
```
cjpillsbury, line 18: Another example of usage.

src/js/experimental/media-chrome-listbox.js
gkatsev: Changing the colors here changes the contrast ratios of the colors and text between the components. It might be fine, but we'd need to survey the contrast ratios.

src/js/experimental/media-captions-listbox.js
gkatsev, line 9: I hate that the default color value is copied to each file :(

That said, can be a future improvement. Definitely ties into the whole "do we have a global stylesheet" discussion.

heff: 👍 `--media-control`. It's more clear than "media icon". We could also move to `--media-chrome-icon-color` (or -`--media-ui-icon-color`), since `media-control` has an unclear scope. Does media-control apply to everything or just what's in the control bar?

heff: Feels like a pretty low cost for element independence considering it won't change much and it's easily find and replaceable.

src/js/experimental/media-chrome-listbox.js
heff: Is it worth adding a ~ `darker()` to the shared color in a situation like that? We ultimately _want_ people to change these colors, and not have it then be a game of whackamole, so if there's a middle ground somewhere in that it's worth finding.

src/js/experimental/media-chrome-listbox.js
heff: 👍

src/js/media-chrome-button.js
heff: 🎉

src/js/media-chrome-range.js
```diff
@@ -46,7 +46,7 @@ template.innerHTML = `
       box-sizing: border-box;
       display: inline-block;
       position: relative;
-      background: var(--media-control-background, rgba(20,20,30, 0.7));
+      background: var(--media-control-background, var(--media-secondary-color, rgb(20 20 30 / 0.7)));
```
heff, line 49: We seem to be using `control-background` and `secondary` as interchangeable, and that's feeling unnecessary since they're both _global_.

src/js/media-text-display.js
```diff
@@ -12,15 +12,15 @@ template.innerHTML = `
...
+      background: var(--media-control-background, var(--media-secondary-color, rgb(20 20 30 / 0.7)));
   
       padding: var(--media-control-padding, 10px);
 
       font-size: 14px;
       line-height: var(--media-text-content-height, var(--media-control-height, 24px));
       font-family: Arial, sans-serif;
       text-align: center;
-      color: #ffffff;
+      color: var(--media-control-font-color, var(--media-primary-color, rgb(238 238 238)));
```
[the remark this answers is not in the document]
heff: ffffff

heff (commented):
Great progress. 👍 

After reading through it, my s1 is:
* Move to a `--media-chrome` prefix for chrome-global settings. e.g. `--media-chrome-icon-color` rather than `--media-control-icon-color`. If we need to distinguish icons we should probably do it with a more clear word than "control".
* Kill "primary" and "secondary" and instead rely on `--media-chrome-foreground-color` and `--media-chrome-background-color`. "secondary" at least is really ambiguous. If we do need it I'd love to understand why better. And themes can still implement their own primary and secondary color concepts, like the demuxed one appears to be doing.

Otherwise there's a few decisions here probably worth documenting somewhere at the end of this:
* How we name these
* Choosing the shortcut property over the specific properties, e.g. `background` over `background-color`. That may not matter if we can provide variables and defaults for the specific properties without making the short cut properties confusing or harder to work with. But when I was creating the initial variables this seemed like the simpler path.

src/js/experimental/media-chrome-listbox.js
luwes: `font-color`, that feels off to me. a font:

> a file or set of characters (with a very specific set of characteristics, including weight, style, and width) used to create text in a specific style

it's the text that is colored, not the design of the letters.

src/js/experimental/media-captions-listbox.js
gkatsev: It's definitely possible to do a find and replace, but it's much nicer that if we decide to change the default primary color, we can change it in one place only.

Additionally, easy find and replace relies on copying the default color perfectly, though, could search for `--media-primary-color`, but that makes it a two-step thing.

I do think that this is a good *future* improvement. Especially, once we figure out whether we want a global stylesheet or do some build tooling to handle CSS for us.

Having the default be set in a single place shouldn't eliminate element independence. If we use build tooling, we can have a `defaults.css` that gets injected into each component at build time. With a global CSS, a user would need to include those extra styles, or they could skip it and set the primary-color directly.

We should strive to find a balance between improvements for users and improvement for developers. Some of this can be improved by having better developer-focus documentation, like listing out the default values for these things. When I was working on the select-menu and listbox stuff, I often had to look for other elements that might have a color set, or a background set to grab their styles. I wished that I could have a list of color variables that I could reference to use instead of copying the specific colors.

src/js/experimental/media-chrome-listbox.js
gkatsev: Assuming that these colors aren't good enough, we could add a second set of `-darker` color varilables.

At some point soon, oklch will be in all browsers which would make it a lot easier to do these kinds of things ad-hoc, too.

src/js/experimental/media-chrome-listbox.js
gkatsev: I don't think it's going to be a problem for us, but I'm always wary of changes of `background-color` to `background` because it could reset other props that `background` is a shorthand for as well.

src/js/media-control-bar.js
gkatsev: yeah, I think we should set `fill` as well.

src/js/themes/media-theme-youtube.js
gkatsev: just wanted to make a note here that if a select-menu/listbox was used, the text would become unreadable in most cases.

src/js/experimental/media-captions-listbox.js
gkatsev: Also, should the css var be called `--media-control-fill`?

Though, I think `--media-control-icon-color` and `--media-control-text-color` make sense to me.



> Does media-control apply to everything or just what's in the control bar?

yeah, that's a tricky one. Since they're global, would definitely work where ever. But the naming does make it seem like maybe it's required to be part of the control bar.

src/js/media-time-range.js
luwes: we have to be careful here, should be tested with Mux player's changeable colors

cjpillsbury:
@heff / @luwes / @gkatsev for a few of these questions, re: naming - we'll want to think about cases where there is a difference between those things that apply in fact to "everything" vs. those things that apply to "most things". The most glaring example of this in this spike/reference is the background - it applies to all UI "components"* (with exceptions currently in the listbox case, which we should also discuss), but does not apply to e.g. the `media-controller`, or the `media-poster-image`, or the `media-control-bar`. It would be good for us to come up with some convention that makes those sorts of things clear.

cjpillsbury:
@heff re:
> Choosing the shortcut property over the specific properties, e.g. `background` over `background-color`. That may not matter if we can provide variables and defaults for the specific properties without making the short cut properties confusing or harder to work with. But when I was creating the initial variables this seemed like the simpler path.

I believe we actually intentionally switched this so we could apply gradients to backgrounds cleanly. I think it may be advantageous for us to prefer the shorthands over the isolated variables (at least for v1), though this is definitely a good sub-topic for discussion.

src/js/experimental/media-captions-listbox.js
cjpillsbury: This is a good discussion that I'd like to have, though my S1 is it doesn't directly impact our primary goals for the CSS updates and should be treated as a follow up. I definitely don't want this conversation to "disappear into the ether" though.

gkatsev: Yeah, I think it definitely can be done as a follow-up.

Though, perhaps developer-focused documentation on the CSS vars should be part of this effort.

PR #474, opened 2023-02-15
fix: add lazy template prop to theme el

luwes:
this is required for (Mux Player) React because sometimes the `media-theme` element is not upgrade yet so the template property will be set on an HTMLElement. After the upgrade we have to re-set it on the instance of the custom element...

https://web.dev/custom-elements-best-practices/#make-properties-lazy

### DIFF
--- a/src/js/media-theme-element.js
+++ b/src/js/media-theme-element.js
@@ -57,9 +57,7 @@ export class MediaThemeElement extends window.HTMLElement {
     observer.observe(this.renderRoot, {
       attributes: true,
       subtree: true,
-    })
-
-    this.createRenderer();
+    });
   }
 
   get mediaController() {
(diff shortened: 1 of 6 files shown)


src/js/themes/media-theme-demuxed-2022.js
```diff
@@ -7,7 +7,7 @@
 </media-theme-demuxed-2022>
 */
 
-import { window } from '../utils/server-safe-globals.js';
+import { window, document } from '../utils/server-safe-globals.js';
```
cjpillsbury, line 10: praise: Good catches re: importing document ❤️

src/js/media-theme-element.js
```diff
@@ -60,6 +60,20 @@ export class MediaThemeElement extends window.HTMLElement {
     })
 
     this.createRenderer();
+
+    // In case the template prop was set before custom element upgrade.
+    // https://web.dev/custom-elements-best-practices/#make-properties-lazy
+    this.#upgradeProperty('template');
```
cjpillsbury, line 66: question(non-blocking): In the best practices example referenced by the link ☝️, this happens in `connectedCallback`. Any reason why we're doing this in the `constructor` instead?

cjpillsbury (approved):
This all makes sense to me. Just one non-blocking question. 🎉

src/js/media-theme-element.js
luwes: good call, I think I figured because it worked and there was no need to re-evaluate it on each connectedCallback but thinking about this again connectedCallback is called slightly later and it's needed there to safely access custom element attributes which we do in the props getter so I'll move it here 👍

PR #476, opened 2023-02-16
Feat/media UI extensions live et al

cjpillsbury:
This adds support for APIs currently being drafted in [media-ui-extensions](https://github.com/video-dev/media-ui-extensions/issues/3), including:

- [Stream Type (proposal issue)](https://github.com/video-dev/media-ui-extensions/issues/3)
- [DVR State (proposal issue)](https://github.com/video-dev/media-ui-extensions/issues/4)
- [Live Edge (proposal draft PR)](https://github.com/video-dev/media-ui-extensions/pull/7)

Additionally, it updates the behavior of the `media-live-button` per prior conversations around presentation and behavior related to the `media-paused` state.


src/js/media-controller.js
```diff
@@ -472,7 +472,16 @@ class MediaController extends MediaContainer {
...
+      },
+      'emptied,durationchange,loadedmetadata,streamtypechange': () => {
+        this.propagateMediaState(
+          MediaUIAttributes.MEDIA_STREAM_TYPE
+        );
+      },
+      'emptied,durationchange,loadedmetadata,targetlivewindowchange': () => {
+        this.propagateMediaState(
+          MediaUIAttributes.MEDIA_TARGET_LIVE_WINDOW
+        );
```
cjpillsbury, line 484: NOTE: On unchanged lines https://github.com/muxinc/media-chrome/pull/476/files#diff-f093c2ba0a65c5a566a4ad86659ffc5bee29bfbbc241f863bed8f634b7a0f9f7R508-R510 where we monitor for events to update `MEDIA_TIME_IS_LIVE`, this does not account for all cases where that value may change, nor is there any set of events on an `HTMLMediaElement` that would catch all cases. Given our current use case, we may want to instead model something like `MEDIA_IS_PLAYING_LIVE` that means:

1. media is currently playing
2. the media playing is within the live edge window

This more circumscribed attribute should be able to be reliably updated based on events and would simplify `media-live-button`.

package.json
```diff
@@ -24,7 +24,7 @@
     "watch": "run-p watch:types watch:esbuild",
     "watch:types": "yarn build:types -w",
     "watch:esbuild": "node ./scripts/esbuild.js",
-    "dev": "yarn watch",
+    "dev": "yarn build:esm && yarn build:types && yarn watch",
```
cjpillsbury, line 27: TODO: I need to revert this commit but was noticing some issues with dev watch behavior. Can circle back in one of our syncs.

src/js/media-controller.js
```diff
@@ -1007,66 +1016,37 @@ const Delegates = {
...
     
     if (!media) return false;
-
-    const streamIsLive = controller.getAttribute(MediaUIAttributes.MEDIA_STREAM_TYPE) === 'live';
-    const seekable = media.seekable;
-
-    // If there's no way to seek, assume the media element is keeping it "live"
-    if (streamIsLive && !seekable) {
-      return true;
+    if (typeof media.liveEdgeStart === 'number') {
```
cjpillsbury, line 1028: should this move into its own getter function for consistency?

src/js/media-controller.js
```diff
@@ -1154,6 +1134,33 @@ const getShowingCaptionTracks = (controller) => {
   });
 };
 
+const getStreamType = (controller) => {
+  const { media } = controller;
+  
+  if (!media) return undefined;
```
gkatsev, line 1140: Previously, this returned `null`, why the change?

src/js/media-controller.js
```diff
@@ -1007,66 +1016,37 @@ const Delegates = {
...
-
-    const currentTime = media.currentTime;
-    const seekableEnd = seekable.end(seekable.length - 1);
-
-    if (currentTime > seekableEnd - liveThreshold) {
-      return true;
-    }
-
-    return false;
+    const liveEdgeStartOffset = controller.hasAttribute('livethreshold')
```
gkatsev, line 1045: Should this attribute be with respect to the new proposal? Maybe to match the variable name?

src/js/media-controller.js
cjpillsbury: Just for consistency with the spec. If there's no slotted media, I'm treating this as equivalent to "unimplemented". From "the out side" (aka the attribute value), this should be identical regardless.

src/js/media-controller.js
cjpillsbury: Yeah I agree this attribute should get a rename. Since the proposal moved to `liveEdgeStart`, there isn't an explicit equivalent of `livethreshold` (it's effectively the variable name `liveEdgeStartOffset`). Do you think the attribute should be `liveedgestartoffset`? or `liveedgeoffset` (maybe a bit confusing, though that was the name of the property in the previous iteration of the proposal)? or something else? Def open to suggestions here (Names Are Hard™️)

src/js/media-controller.js
```diff
@@ -1007,66 +1016,37 @@ const Delegates = {
...
     }
 
-    // If the slotted media does not provide a `seekable` property or its length is falsey,
-    // assume the media is not live.
-    if (!seekable?.length) {
-      return false;
-    }
+    const live = getStreamType(controller) === 'live';
+    // Can't be playing live if it's not a live stream
+    if (!live) return false;
```
luwes, line 1035: nit (non-b): 1 empty line after a return statement might help reading the code

src/js/media-controller.js
```diff
@@ -1007,66 +1016,37 @@ const Delegates = {
...
+    const seekable = media.seekable;
+    // If the slotted media element is live but does not expose a 'seekable' `TimeRanges` object,
+    // always assume playing live
+    if (!seekable) return true;
+    // If there is an empty `seekable`, assume we are not playing live
+    if (!seekable.length) return false;
```
luwes, line 1041: 👍

src/js/media-controller.js
luwes: `liveedgeease` :) 

`liveedgeoffset` sounds pretty good to me

cjpillsbury: Cool I'll run with that unless folks have any concerns.

src/js/media-controller.js
heff: Interesting to look at this from either the media element or media chrome perspective. I could see this being "media chrome knows what stream type is, but it's unknown". But I'm good with matching the spec here.

heff (approved):

src/js/constants.js
```diff
@@ -84,6 +84,7 @@ export const MediaUIAttributes = {
   MEDIA_LOADING: 'media-loading',
   MEDIA_BUFFERED: 'media-buffered',
   MEDIA_STREAM_TYPE: 'media-stream-type',
+  MEDIA_TARGET_LIVE_WINDOW: 'media-target-live-window',
```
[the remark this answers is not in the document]
cjpillsbury: I don't believe so. I think "target live window" makes sense still in our "DVR" proposal. We can always amend it if necessary once we move https://github.com/video-dev/media-ui-extensions/issues/4 to an official proposal PR, since there isn't a custom/extended media element that supports this yet (and `mux-video` will be the first, so we can treat it as "internal")

src/js/media-controller.js
cjpillsbury: for the updated implementation of `media-live`button` in this PR, the edge case will not be a problem. However, the `media-time-is-live` will "lie" on a subset of initialization cases for live/dvr, particularly when autoplay is off. As long as we're ok with that attribute not always being accurate, we can definitely treat that as out of scope and a "we'll deal with it if/when it becomes an issue" scenario, since it doesn't impact any current media chrome use cases/components.

src/js/media-controller.js
```diff
@@ -1007,66 +1016,38 @@ const Delegates = {
...
-
-    const currentTime = media.currentTime;
-    const seekableEnd = seekable.end(seekable.length - 1);
-
-    if (currentTime > seekableEnd - liveThreshold) {
-      return true;
-    }
-
-    return false;
+    const liveEdgeStartOffset = controller.hasAttribute('liveedgeoffset')
```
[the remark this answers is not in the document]
cjpillsbury: I did a quick search and didn't see any docs, but I'll be sure to do a more thorough audit before merging.

src/js/media-controller.js
cjpillsbury: @heff yup was mulling on the same thing and this is where I (soft) landed. Open to rethinking that moving forward for this kinda stuff if we wanted to discuss in one of our weekly meetings.

luwes (approved):

PR #479, opened 2023-02-23
fix: catch play promise internally

gkatsev:
Was seeing a bunch of uncaught promise rejects in tests and realized it stemmed from here. We should be catching this rejected promise as it happens internally and isn't actionable by users of media-chrome.

### DIFF
--- a/src/js/media-controller.js
+++ b/src/js/media-controller.js
@@ -97,7 +97,7 @@ class MediaController extends MediaContainer {
           mediaUIEventHandlers['MEDIA_SEEK_TO_LIVE_REQUEST'](e, media);
         }
 
-        this.media.play();
+        this.media.play().catch(() => {});
       },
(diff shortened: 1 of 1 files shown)


src/js/media-controller.js
```diff
@@ -97,7 +97,7 @@ class MediaController extends MediaContainer {
           mediaUIEventHandlers['MEDIA_SEEK_TO_LIVE_REQUEST'](e, media);
         }
 
-        this.media.play();
+        this.media.play().catch(() => {});
```
cjpillsbury, line 100: suggestion(non-blocking): while you're in here, you might want to use `this.media?.play()` etc. or `this.media.play?.()` (since we haven't defined a "minimally valid slotted media element" yet) to catch those edge cases as well

cjpillsbury (approved):
👍

PR #480, opened 2023-02-27
fix: Use default-stream-type when slotted media streamType is unknown.

cjpillsbury:
Since `"unknown"` is supposed to represent cases where the slotted media element supports the `streamType` [media-ui-extension](https://github.com/video-dev/media-ui-extensions/issues/3) but cannot derive the stream type yet, updating impl to prefer `default-stream-type` in this case.

### DIFF
--- a/src/js/media-controller.js
+++ b/src/js/media-controller.js
@@ -1140,7 +1140,19 @@ const getStreamType = (controller) => {
   
   if (!media) return undefined;
   
-  if (media.streamType) return media.streamType;
+  if (media.streamType) {
+    // If the slotted media supports `streamType` but
+    // `streamType` is "unknown", prefer `default-stream-type`
+    // if set (CJP)
+    if (media.streamType === 'unknown' && controller.hasAttribute('default-stream-type')) {
+      const defaultType = controller.getAttribute('default-stream-type');
+
+      if (StreamTypeValues.includes(defaultType)) {
+        return defaultType;
+      }
+    }
+    return media.streamType;
+  }
   const duration = media.duration;
 
   if (duration === Infinity) {


src/js/media-controller.js
```diff
@@ -1140,7 +1140,19 @@ const getStreamType = (controller) => {
...
+    // if set (CJP)
+    if (media.streamType === 'unknown' && controller.hasAttribute('default-stream-type')) {
+      const defaultType = controller.getAttribute('default-stream-type');
+
+      if (StreamTypeValues.includes(defaultType)) {
+        return defaultType;
+      }
+    }
+    return media.streamType;
+  }
```
luwes, line 1155: just wondering why not gate it here and still try to derive it from the duration?

```suggestion
  if (media.streamType && media.streamType !== 'unknown') {
    return media.streamType;
  }
```

cjpillsbury: @luwes I think that's conflating two different things. The derived case is intended for slotted media elements that do not support the `streamType` api. `default-stream-type` is a media chrome concept representing "what we want to treat as the stream type when it is not known". If someone has implemented `streamType` in their extended media element, I think we should treat that as the "source of truth". So, in this case, the slotted media definitely has an implementation of `streamType` (and therefore we should not override their implementation in media chrome), but the stream type is currently unknown for their implementation (and therefore we should use `default-stream-type` if set).

luwes (approved):
LGTM

src/js/media-controller.js
```diff
@@ -1140,7 +1140,19 @@ const getStreamType = (controller) => {
...
-  if (media.streamType) return media.streamType;
+  if (media.streamType) {
+    // If the slotted media supports `streamType` but
+    // `streamType` is "unknown", prefer `default-stream-type`
+    // if set (CJP)
+    if (media.streamType === 'unknown' && controller.hasAttribute('default-stream-type')) {
+      const defaultType = controller.getAttribute('default-stream-type');
+
+      if (StreamTypeValues.includes(defaultType)) {
+        return defaultType;
```
heff, line 1151: I was considering if we should have a default default stream type, e.g. if no default stream type then it's on-demand. But I think ultimately that can be up to the player designer to decide how to treat the "no info" default. 👍

heff (approved):

PR #484, opened 2023-03-01
fix: nullish coalesce operator, improve process

luwes:
this change was mainly to fix an issue with the regex of the nullish coalesce operator that @cjpillsbury found.

instead of tweaking the regex I did a small refactor so all expressions are now run through the same tokenizer and evaluator.
this should be more robust, having 1 location for matching the tokens.


cjpillsbury (approved):
LGTM!

PR #490, opened 2023-03-03
fix: add default much used styles to theme

luwes:
while I was writing the theme docs I noticed the multi-layout theme has very little styles
https://media-chrome-docs-git-fork-luwes-theme-docs-mux.vercel.app/en/themes/responsive-themes

should it have any styles at all to get this functionality came to mind?

maybe proper defaults makes sense in the base theme, it's always possible to override them.

### DIFF
--- a/src/js/media-theme-element.js
+++ b/src/js/media-theme-element.js
@@ -11,6 +11,22 @@ const observedMediaAttributes = {
   'media-stream-type': 'streamType',
 };
 
+const prependTemplate = document.createElement('template');
+
+prependTemplate.innerHTML = `
+  <style>
+    :host {
+      display: block;
+      line-height: 0;
+    }
+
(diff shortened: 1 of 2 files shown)


src/js/media-theme-element.js
```diff
@@ -11,6 +11,22 @@ const observedMediaAttributes = {
...
 };
 
+const prependTemplate = document.createElement('template');
+
+prependTemplate.innerHTML = `
+  <style>
+    :host {
+      display: block;
+      line-height: 0;
+    }
```
gkatsev, line 21: this matches the defaults set on media-controller 👍🏻

gkatsev (approved):
Yeah, I guess this makes sense to me. It makes the theme element behave closer to what media-controller behaves when used directly, right?

PR #491, opened 2023-03-04
feat: support fetching theme HTML files

luwes:
Test URL: https://media-chrome-git-fork-luwes-theme-template-src-mux.vercel.app/examples/experimental/themes/micro-theme.html

this change adds support for fetching HTML files as theme templates.

this works better than I expected with the discovery of `link[rel=preload][as=fetch]`.
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/preload#what_types_of_content_can_be_preloaded

I initially thought it required the unsupported `as=document` or HTML imports.

The preloading is not required but it helps with showing the whole theme because only after the JS is loaded and executed can it start fetching the HTML file without any hints.

as you can see in the example both a local and remote url work.
it does require `access-control-allow-origin: *` in the response of the remote HTML url I believe.

### DIFF
--- a/examples/experimental/themes/micro-theme.html
+++ b/examples/experimental/themes/micro-theme.html
@@ -1,0 +1,78 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta name="viewport" content="width=device-width" />
+    <title>Media Chrome Micro Theme Example</title>
+
+    <script type="module" src="../../../../src/js/media-theme-element.js"></script>
+    <script type="module" src="../../../../dist/index.js"></script>
+
+    <!-- Preload theme HTML files so they're loaded in parallel with JS files  -->
+    <link rel="preload" href="./micro.html" as="fetch" crossorigin>
+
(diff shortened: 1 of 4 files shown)


examples/experimental/themes/micro-theme.html
```diff
@@ -0,0 +1,78 @@
...
+      }
+
+      video {
+        width: 100%;      /* prevents video to expand beyond its container */
+      }
+    </style>
+    <main>
+      <h1>Media Chrome Micro Theme Example</h1>
+
+      <media-theme template="https://gist.githubusercontent.com/luwes/5812c419830fee000d3463c496d18e19/raw/c295dad03a33ea8ad93870fa55de40a3308c8f45/media-theme-micro.html">
```
luwes, line 35: remote url

examples/experimental/themes/micro-theme.html
```diff
@@ -0,0 +1,78 @@
...
+            default
+            kind="metadata"
+            src="https://image.mux.com/O6LdRc0112FEJXH00bGsN9Q31yu5EIVHTgjTKRkKtEq1k/storyboard.vtt"
+          />
+        </video>
+      </media-theme>
+
+      <br>
+
+      <media-theme template="micro.html">
```
luwes, line 55: local url

PR #492, opened 2023-03-07
feat: Minimal theme

AdamJaggard:
A new minimal theme supporting live and VOD.

image

Todo:
- ~Show time range while in live DVR mode~
- ~Change the shape of the live button indicator~


src/js/themes/minimal.js
```diff
@@ -0,0 +1,230 @@
...
+const template = document.createElement('template');
+template.innerHTML = html`
+  <style>
+    media-controller {
+      --media-control-background: transparent;
+      --media-control-hover-background: transparent;
+    }
+
+    media-control-bar {
+      height: 42px;
```
luwes, line 15: going by the designs extra small is 30px high and the margins are smaller too

SCR-20230315-g0v

src/js/themes/minimal.js
```diff
@@ -0,0 +1,230 @@
...
+              <media-time-display></media-time-display>
+            </template>
+          </template>
+        </div>
+        <template if="targetLiveWindow > 0">
+          <template if="breakpointSm">{{>TimeRange}}</template>
+        </template>
+        <div class="live-controls-right">
+          {{>MuteButton}}{{>VolumeRange}}{{>CaptionsButton}}{{>FullscreenButton}}
+        </div>
```
luwes, line 214: cool to see how live and DVR is mixed here in one main live view.

suggestion (non-blocking): we've been using a "spacer" div to push out left and right parts of the control bar.
it removes one level of nesting. 

see for example 
https://codesandbox.io/s/media-chrome-multi-layout-theme-gwlon8?file=/index.html

src/js/themes/minimal.js
```diff
@@ -0,0 +1,230 @@
...
+import { MediaThemeElement } from '../media-theme-element.js';
+
+const html = (raw, ...keys) => String.raw({ raw }, ...keys);
+
+const template = document.createElement('template');
+template.innerHTML = html`
+  <style>
+    media-controller {
+      --media-control-background: transparent;
+      --media-control-hover-background: transparent;
```
luwes, line 11: the time range and volume range has a border-radius in design.
there's a CSS var for that `--media-range-track-border-radius`

luwes:
looking good.

we'd still need to add the extra buttons that are not turned on by default. like seek forward / backward, cast, airplay etc and add CSS vars to toggle them on / off 

see 
https://github.com/muxinc/media-chrome/blob/94ccd2ba6c44e6ee279f4dc5d8527456c5c0ac75/src/js/themes/micro.js#L359-L405
https://github.com/muxinc/media-chrome/blob/94ccd2ba6c44e6ee279f4dc5d8527456c5c0ac75/src/js/themes/micro.js#L224-L262

src/js/themes/minimal.js
```diff
@@ -0,0 +1,230 @@
...
+
+    .live-controls-left {
+      margin-right: auto;
+    }
+
+    [disabled],
+    [aria-disabled='true'] {
+      opacity: 60%;
+      cursor: not-allowed;
+    }
```
luwes, line 46: this CSS is also required for when features are not supported 
https://github.com/muxinc/media-chrome/blob/94ccd2ba6c44e6ee279f4dc5d8527456c5c0ac75/src/js/themes/micro.js#L18-L40

gkatsev (approved):
LGTM!

src/js/themes/minimal.js
AdamJaggard: Have added this and definitions for the optional extra buttons. Is there anything else that needs doing to accommodate the optional buttons in the theme templates? How do consumers of the theme enabled these extra buttons?

luwes: oh sorry about this for the default disabled buttons the syntax changed a little. this should only be added, you can copy / paste that 

https://github.com/muxinc/media-chrome/blob/main/src/js/themes/microvideo.js#L224-L236

PR #494, opened 2023-03-08
Fix/improve inferred media-ui-extension values

cjpillsbury:
This ensures that `<media-controller>`'s state representation for `streamType` and `targetLiveWindow` are more predictable, even for cases where the slotted media element has no/partial support for these media-ui-extension values. This will allow for more predictable UI design, including (especially) with Media Chrome themes.

### DIFF
--- a/src/js/media-controller.js
+++ b/src/js/media-controller.js
@@ -478,7 +478,7 @@ class MediaController extends MediaContainer {
           MediaUIAttributes.MEDIA_STREAM_TYPE
         );
       },
-      'emptied,durationchange,loadedmetadata,targetlivewindowchange': () => {
+      'emptied,durationchange,loadedmetadata,streamtypechange,targetlivewindowchange': () => {
         this.propagateMediaState(
           MediaUIAttributes.MEDIA_TARGET_LIVE_WINDOW
         );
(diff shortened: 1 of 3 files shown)


src/js/media-theme-element.js
```diff
@@ -1,4 +1,4 @@
-import { window } from './utils/server-safe-globals.js';
+import { window, document } from './utils/server-safe-globals.js';
```
cjpillsbury, line 1: NOTE: This is unrelated but got caught while testing. There's a module-level reference to `document` that was causing crashes for SSR usage.

src/js/media-controller.js
```diff
@@ -1137,14 +1137,17 @@ const getShowingCaptionTracks = (controller) => {
 
 const getStreamType = (controller) => {
   const { media } = controller;
-  
-  if (!media) return undefined;
-  
+
+  if (!media) return StreamTypes.UNKNOWN;
```
luwes, line 1141: if there is no media could the stream type algo have run? should this stay undefined?

will this cause a new attribute to be added `media-stream-type=unknown`?

src/js/media-controller.js
```diff
@@ -1167,11 +1170,23 @@ const getStreamType = (controller) => {
...
+
+  if (!media) return Number.NaN;
+  const { targetLiveWindow } = media;
+  // Since `NaN` represents either "unknown" or "inapplicable", need to check if `streamType`
+  // is `"live"`. If so, assume it's "standard live" (aka `targetLiveWindow === 0`) (CJP)
+  if (
+    (targetLiveWindow == null || Number.isNaN(targetLiveWindow)) &&
+    getStreamType(controller) === StreamTypes.LIVE
+  ) {
+    return 0;
```
luwes, line 1187: 👍

src/js/media-controller.js
cjpillsbury: > if there is no media could the stream type algo have run? 
Yes
> should this stay undefined?
Maybe (See below)
> will this cause a new attribute to be added `media-stream-type=unknown`?
Yes.

The current implementation strictly conforms with expected values from https://github.com/video-dev/media-ui-extensions/issues/3. I do think for this case there is a compelling argument to instead deviate (@heff called this out in the proposal discussion), but we want to be consistent here one way or another, either:

1. **_always_** translate `"unknown"` cases to `undefined`
2.. _**always**_ treat `undefined` and similar cases as `"unknown"`

Eventually, we may want to distinguish between these two scenarios in Media Chrome, but for now, this will only add complexity for UI devs as far as I can tell. I'll refactor to always use `undefined` (aka (1), above) per your / @heff's callout just to reduce the attributes footprint, though, in general, I think it's good to err on the side of predictable consistency with media-ui-extensions.

luwes (approved):
LGTM

PR #495, opened 2023-03-08
fix: preventClick on the button given to selectmenu unconditionally

gkatsev:
Currently, the click handling is only turned off if the button is enabled when slotted

but it should also be turned off if the button is disabled

### DIFF
--- a/src/js/experimental/media-chrome-selectmenu.js
+++ b/src/js/experimental/media-chrome-selectmenu.js
@@ -83,6 +83,9 @@ class MediaChromeSelectMenu extends window.HTMLElement {
       // update button reference if necessary
       this.#button = newButton;
 
+      // if it's a media-chrome-button, ask it to not handle the click event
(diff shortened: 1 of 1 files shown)


src/js/experimental/media-chrome-selectmenu.js
```diff
@@ -83,6 +83,9 @@ class MediaChromeSelectMenu extends window.HTMLElement {
       // update button reference if necessary
       this.#button = newButton;
 
+      // if it's a media-chrome-button, ask it to not handle the click event
+      this.#button.preventClick = true;
```
cjpillsbury, line 87: suggestion(non-blocking): as a followup, we may want to better account for buttons with a `preventClick` API vs. those that don't have it (tl;dr - this is unnecessarily mutative for non `media-chrome-button` slotted buttons.)

cjpillsbury (approved):
🏃

PR #496, opened 2023-03-08
fix: theme inline-block, improve responsive theme

luwes:
Made the responsive theme example show a similar state with the basic example for comparison.
`media-theme-element` actually needed a default of `display: inline-block` to match with theme-less out of the box.

Basic: https://media-chrome-git-fork-luwes-responsive-theme-tweaks-mux.vercel.app/examples/basic.html
Theme: https://media-chrome-git-fork-luwes-responsive-theme-tweaks-mux.vercel.app/examples/themes/responsive-theme.html

if the media-theme-element was `display: block` by default the audio player would expand 100% in the theme example

### DIFF
--- a/src/js/themes/responsive.js
+++ b/src/js/themes/responsive.js
@@ -13,19 +13,14 @@ import { MediaThemeElement } from '../media-theme-element.js';
 const template = document.createElement('template');
 template.innerHTML = `
 <style>
-  :host {
-    display: inline-block;
-  }
-  media-controller:not([audio]) {
-    display: block;
-    aspect-ratio: 16 / 9;   /* set container aspect ratio if preload=none */
-  }
-  video {
-    width: 100%;      /* prevents video to expand beyond its container */
-  }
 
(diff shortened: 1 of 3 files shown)


src/js/themes/responsive.js
```diff
@@ -13,19 +13,14 @@ import { MediaThemeElement } from '../media-theme-element.js';
 const template = document.createElement('template');
 template.innerHTML = `
 <style>
-  :host {
-    display: inline-block;
-  }
-  media-controller:not([audio]) {
-    display: block;
-    aspect-ratio: 16 / 9;   /* set container aspect ratio if preload=none */
```
gkatsev, line 21: I guess this moved to the html?

gkatsev (approved):

src/js/themes/responsive.js
luwes: yup, this should be handled by the consumer of the theme

PR #497, opened 2023-03-09
fix: add display CSS vars

luwes:
this adds CSS vars for the control's `display` property.
it's handy for when using a theme to show/hide specific controls.

```html
  <media-theme-responsive style="--media-fullscreen-button-display: none"></media-theme-responsive>
```

a theme or player can also easily create higher level CSS vars by forwarding

```css
media-controller::part(top) {
  --media-play-button-display: var(--top-controls, var(--top-play-button, inline-flex));
  --media-mute-button-display: var(--top-controls, var(--top-mute-button, inline-flex));
}
```

### DIFF
--- a/src/js/media-chrome-button.js
+++ b/src/js/media-chrome-button.js
@@ -1,4 +1,5 @@
 import { MediaStateReceiverAttributes } from './constants.js';
+import { getOrInsertCSSRule } from './utils/element-utils.js';
 import { window, document } from './utils/server-safe-globals.js';
 
 const template = document.createElement('template');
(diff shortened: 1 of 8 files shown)


src/js/media-chrome-button.js
```diff
@@ -91,7 +92,10 @@ class MediaChromeButton extends window.HTMLElement {
 
     this.nativeEl.appendChild(slotTemplate.content.cloneNode(true));
 
-    shadow.appendChild(buttonHTML);
+    this.shadowRoot.appendChild(buttonHTML);
+
+    const { style } = getOrInsertCSSRule(this.shadowRoot, ':host');
+    style.setProperty('display', `var(--${this.localName}-display, inline-flex)`);
```
gkatsev, line 98: I thought that relying on `localName` is an issue, but it's only an issue if you're trying to override it (had some issues previously, where the localName was getting set, but the switch to strict mode broke things because it disallows it, but loose mode still allows it and just makes it silently fail).

gkatsev (approved):

PR #498, opened 2023-03-09
example: remove unneeded media-loading attr

luwes:
this attribute is unneeded as far as I can tell. it will be propagated by the waiting event 

https://github.com/muxinc/media-chrome/blob/90281290051c00acc95013ed82c712c77f4229e9/src/js/media-controller.js#L504-L507


luwes:
going w/ bias for action here and merge. only touches examples and we don't use this attribute in Mux player. very sure this attribute is not needed.

PR #502, opened 2023-03-17
Add control display CSS var

luwes:
This change adds the `--media-control-display` css var and adds the display var also to the selectmenu elements.

Changes the Micro theme to make use of the new CSS vars

### DIFF
--- a/src/js/themes/micro.js
+++ b/src/js/themes/micro.js
@@ -221,44 +221,17 @@ template.innerHTML = html`
     width: 7px;
   }
 
-  media-play-button {
-    display: var(--controls, var(--play-button, inline-flex));
-  }
-
+  /* Turn some buttons off by default */
   media-seek-backward-button {
-    display: var(--controls, var(--seek-backward-button, none));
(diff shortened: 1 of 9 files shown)


src/js/themes/micro.js
```diff
@@ -221,44 +221,17 @@ template.innerHTML = html`
     width: 7px;
   }
 
-  media-play-button {
-    display: var(--controls, var(--play-button, inline-flex));
```
luwes, line 225: also note here that I deleted the Mux player "short" CSS vars. 
I figured that was best kept as a specific thing for Mux player.

themes will use the regular display vars defined in Media Chrome. e.g. `--media-seek-backward-button-display`

Mux player adds these short vars with a forwarding like defined here 
https://github.com/muxinc/elements/blob/883a648f9e84a241dc0c4364f8526954b0324429/packages/mux-player/src/styles.css

heff (approved):
👍

gkatsev (approved):

PR #503, opened 2023-03-17
refactor(experimental): don't put generated elements into the light DOM

gkatsev:
While it's nice that the data-driven elements put children into the light DOM, this can break usage in react if the element isn't included in the shadow DOM by itself. Therefore, we need to keep sprouted elements in the shadow DOM. This would mean that users who'd want to style these would need to use parts and may be limited in what they can do unless they create a Web Component that extends the components they need more control over. 

Uses the commit from #493

TODO:
- [ ] Add parts to the sprouted elements
- [ ] Fix up the assignedElements things so that the defaults are set up correctly, and we don't need to do a check and reach into the DOM each time.


luwes (changes requested):
there is something weird going on
https://media-chrome-git-fork-gkatsev-no-light-dom-generated-els-mux.vercel.app/examples/control-elements/media-chrome-selectmenu.html

try selecting the same option 3 times, eventually it toggles between off and English

luwes:
> make media-chrome-listitem more similar to <option>

I was digging in this recently. I do see them used mixed in the aria examples where they set `role=option` on li elements but since we own the naming of the element is it worth to name the `media-chrome-listitem` => `media-chrome-option` because a `listbox` holds `option`'s, a `list` holds `listitem`'s 

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listbox_role
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/list_role

gkatsev:
Thanks for reviewing!

> there is something weird going on https://media-chrome-git-fork-gkatsev-no-light-dom-generated-els-mux.vercel.app/examples/control-elements/media-chrome-selectmenu.html
> 
> try selecting the same option 3 times, eventually it toggles between off and English

In a way, that's by design, but you're right that we should probably turn it off for the listbox when set up as part of the selectmenu. Though, I think it's out-of-scope for this PR. Definitely something that should be addressed before it's no longer experimental.


> I was digging in this recently. I do see them used mixed in the aria examples where they set `role=option` on li elements but since we own the naming of the element is it worth to name the `media-chrome-listitem` => `media-chrome-option` because a `listbox` holds `option`'s, a `list` holds `listitem`'s
> 
> https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listbox_role https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/list_role

I think that is a reasonable, but I think it's also out-of-scope for this PR.

I added both of these to https://github.com/muxinc/media-chrome/issues/387 so they definitely won't get forgotten about.

docs/src/pages/en/media-captions-selectmenu.md
```diff
@@ -0,0 +1,110 @@
...
+      <span slot="off">CC</span>
+    </media-captions-button>
+</media-captions-selectmenu>
+```
+
+## Styling
+
+media-captions-listbox puts all the children in its shadow DOM, which makes it a bit harder to style. One alternative is to slot a button or listbox element and then style them directly. However, this component also [exposes three parts](#exposed-parts) that can be targeted for styling.
+
+Check out this example for usage, but please try and use better colors as this color scheme is for demonstration purposes only.
```
heff, line 57: There's only code here, how do I see this example?

heff (approved):
Nothing jumps out at me as an issue on first pass.

I think there should be an additional doc written to show the user experience here (for selectmenu in general) which would be along the lines of "How to style media-control-bar buttons/elements".
I'm assuming in this guide there will need to be some caveats related to selectmenus when you're trying to apply the same styles to all direct children of control-bar. Right now I don't have a clear vision of what those caveats will look like, and if we should/could do any more to address them.

docs/src/pages/en/media-captions-selectmenu.md
gkatsev: yeah, it's broken. Let me change how I import the module and if we figure out a better way, later on, we can change it up again.

src/js/experimental/media-chrome-selectmenu.js
```diff
@@ -138,13 +137,14 @@ class MediaChromeSelectMenu extends window.HTMLElement {
...
+    // one of the components should have a media-controller attribute.
+    // There isn't a good way now to differentiate between default buttons or
+    // or a slotted button but outside of the media-controller.
+    // So, a regular declarative selectmenu may default to open up rather than down.
     if (
       this.hasAttribute('media-controller') ||
       this.#button.hasAttribute('media-controller') ||
-      this.#listbox.hasAttribute('media-controller') ||
-      !this.#buttonSlotted
+      this.#listbox.hasAttribute('media-controller')
```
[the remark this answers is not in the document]
gkatsev: This example demonstrates it better: https://media-chrome-git-fork-gkatsev-no-light-dom-generated-els-mux.vercel.app/examples/control-elements/media-chrome-selectmenu.html
See buttons at the top. This makes it so that the menu opens down rather than up.
This top CC button is outside the media-controller and by default will open down.
image

docs/src/pages/en/media-captions-selectmenu.md
gkatsev: It's fixed now.

gkatsev:
Also, https://github.com/muxinc/media-chrome/pull/503/commits/2155a3f264195018ba452c608a03d4f3109b4018 now shows what needs to be done to make disabled styling work.

gkatsev:
I was going to update the disabled styles to be simple like
```css
[disabled],
[disabled]::part(button) {
  opacity: 60%;
  cursor: not-allowed;
}
```
but I realized that we're then doubling up the opacity. So, we need to do something like:
```css
[disabled]:not(media-captions-selectmenu),
[disabled]::part(button) {
  opacity: 60%;
  cursor: not-allowed;
}
```
If we had a way to mark all selectmenus in some manner, could use that rather than target specific ones.

luwes (approved):

PR #506, opened 2023-03-23
experiment: selectmenu use inheritance for some CSS props

luwes:
in reply to https://github.com/muxinc/media-chrome/issues/387#issuecomment-1481546061

for some CSS we could have this workaround
https://media-chrome-git-fork-luwes-css-experiment-inherit-mux.vercel.app/examples/control-elements/media-chrome-selectmenu.html

### DIFF
--- a/src/js/experimental/media-chrome-selectmenu.js
+++ b/src/js/experimental/media-chrome-selectmenu.js
@@ -12,6 +12,7 @@ template.innerHTML = /*html*/`
     position: relative;
     flex-shrink: .5;
     background: var(--media-control-background, rgba(20,20,30, 0.7));
+    padding: var(--media-control-padding, 10px);
   }
 
   [name="listbox"]::slotted(*),
@@ -25,6 +26,7 @@ template.innerHTML = /*html*/`
 
   [name="button"]::slotted(*),
   [part="button"] {
+    padding: 0;
     background-color: inherit;
   }
   </style>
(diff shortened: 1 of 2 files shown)


src/js/experimental/media-chrome-selectmenu.js
```diff
@@ -5,22 +5,30 @@ import { closestComposedNode, getOrInsertCSSRule } from '../utils/element-utils.
...
     left: 0;
     bottom: 100%;
     max-height: 300px;
     overflow: hidden auto;
   }
+
+  [name="button"]::slotted(*),
+  [part="button"] {
+    padding: 0;
+    background-color: inherit;
```
heff, line 30: Could probably just make this transparent, yeah? Any kind of gradient could we weird/missed otherwise, but maybe that's too different from how selectmenus are supposed to work? Good either way.

heff (approved):

src/js/experimental/media-chrome-selectmenu.js
luwes: I think yes, will test this soon

luwes:
just discovered this would not work nicely with styling the button CSS part.

the example could easier be written with CSS vars and we should document and promote users to use the CSS vars first and then the CSS parts if needed.

```css
          #cb > * {
            --media-control-background: #900;
            --media-control-padding: 5px;
            border: 1px solid #009;
          }
```

gkatsev:
Also, using the inherit piece is complicated. While [`all`](https://developer.mozilla.org/en-US/docs/Web/CSS/all) does exist, using `all: inherit` is likely to cause issues (I tried this approach earlier). This means that only the specific css we make inheritable, will apply, which, I think will make things more inconsistent. Since some things you could apply to the main element without issues but others you'd need to apply to the part.
I do think that at this point the best bet is to make sure that we have the appropriate well-documented CSS variables available (we have some, but may need more, and I don't think they're well documented yet) and the correct CSS parts available (I think button, listbox, and listitem (which should change to `option` if we rename the component) are all that's needed).

PR #510, opened 2023-03-28
fix: toggle time display on click from remaining / not remaining

dylanjha:
Discussion: https://github.com/muxinc/media-chrome/discussions/181

- [x] updates `<media-time-display>` behavior so that it toggles on click between current / remaining
- [x] handles keyboard interaction with Enter & Spacebar
- [x] Do we need `cursor: pointer:` here? What's the best place to put that style rule?
- [x] What's the right conventional commit here? `improv:` ? `fix:` ?

### DIFF
--- a/src/js/media-time-display.js
+++ b/src/js/media-time-display.js
@@ -64,7 +64,8 @@ class MediaTimeDisplay extends MediaTextDisplay {
 
   constructor() {
     super();
-    this.container.innerHTML = formatTimesLabel(this);
+    this.container.style.cursor = 'pointer';
+    this.container.innerHTML = `<span style="cursor: pointer">${formatTimesLabel(this)}</span>`;
   }
 
   connectedCallback() {
--- a/src/js/media-time-display.js
+++ b/src/js/media-time-display.js
@@ -4,6 +4,9 @@ import { formatAsTimePhrase, formatTime } from './utils/time.js';
 import { MediaUIAttributes } from './constants.js';
 import { nouns } from './labels/labels.js';
 // Todo: Use data locals: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleTimeString
+//
+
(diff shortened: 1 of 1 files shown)


src/js/media-time-display.js
```diff
@@ -4,6 +4,9 @@ import { formatAsTimePhrase, formatTime } from './utils/time.js';
 import { MediaUIAttributes } from './constants.js';
 import { nouns } from './labels/labels.js';
 // Todo: Use data locals: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleTimeString
+//
+
+const ButtonPressedKeys = ['Enter', ' '];
```
dylanjha, line 9: Should be somewhere else?

src/js/media-time-display.js
```diff
@@ -71,9 +74,39 @@ class MediaTimeDisplay extends MediaTextDisplay {
 
     this.setAttribute('role', 'progressbar');
     this.setAttribute('aria-label', nouns.PLAYBACK_TIME());
+
+    const keyUpHandler = (evt) => {
```
dylanjha, line 78: seems like some boilerplate that I had to add here to handle keypresses, is there a more generic util for this?

src/js/media-time-display.js
luwes: fine I think

src/js/media-time-display.js
luwes: it's fine here I think, copy a little code

luwes:
I'd go with `fix:` for this change, in the <v1.0.0 state. 
`cursor: pointer` could be an inline style on the span since it's in the shadow DOM

luwes (approved):
LGTM!

I'll follow up this with a few small changes around the `<slot>` logic of MediaTextDisplay,
it's weird that we remove the slot I think.

PR #511, opened 2023-03-29
fix: slot for text display, customizable content

luwes:
this change removes the unneeded container span and public container property from the text display elements and uses the slot and private slot property to add the content.

### DIFF
--- a/src/js/media-current-time-display.js
+++ b/src/js/media-current-time-display.js
@@ -5,18 +5,21 @@ import { MediaUIAttributes } from './constants.js';
 // Todo: Use data locals: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleTimeString
 
 class MediaCurrentTimeDisplay extends MediaTextDisplay {
+  #slot;
+
   static get observedAttributes() {
     return [...super.observedAttributes, MediaUIAttributes.MEDIA_CURRENT_TIME];
   }
 
   constructor() {
     super();
-    this.container.innerHTML = formatTime(0);
+    this.#slot = this.shadowRoot.querySelector('slot');
+    this.#slot.innerHTML = formatTime(0);
   }
 
   attributeChangedCallback(attrName, oldValue, newValue) {
     if (attrName === MediaUIAttributes.MEDIA_CURRENT_TIME) {
-      this.container.innerHTML = formatTime(newValue);
+      this.#slot.innerHTML = formatTime(newValue);
     }
     super.attributeChangedCallback(attrName, oldValue, newValue);
   }
--- a/src/js/media-current-time-display.js
+++ b/src/js/media-current-time-display.js
@@ -14,12 +14,12 @@ class MediaCurrentTimeDisplay extends MediaTextDisplay {
   constructor() {
     super();
     this.#slot = this.shadowRoot.querySelector('slot');
-    this.#slot.innerHTML = formatTime(0);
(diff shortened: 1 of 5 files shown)


src/js/media-current-time-display.js
```diff
@@ -5,18 +5,21 @@ import { MediaUIAttributes } from './constants.js';
...
+
   static get observedAttributes() {
     return [...super.observedAttributes, MediaUIAttributes.MEDIA_CURRENT_TIME];
   }
 
   constructor() {
     super();
-    this.container.innerHTML = formatTime(0);
+    this.#slot = this.shadowRoot.querySelector('slot');
+    this.#slot.innerHTML = formatTime(0);
```
gkatsev, line 17: should these use `textContent` instead?

luwes: def could. I'll update

src/js/media-time-display.js
```diff
@@ -64,8 +67,12 @@ class MediaTimeDisplay extends MediaTextDisplay {
 
   constructor() {
     super();
-    this.container.style.cursor = 'pointer';
-    this.container.innerHTML = `<span style="cursor: pointer">${formatTimesLabel(this)}</span>`;
+
+    this.#slot = this.shadowRoot.querySelector('slot');
+    this.#slot.innerHTML = `${formatTimesLabel(this)}`;
```
cjpillsbury, line 72: suggestion(non-blocking):

```suggestion
this.#slot.textContent = `${formatTimesLabel(this)}`;
```

src/js/media-time-display.js
```diff
@@ -125,7 +132,10 @@ class MediaTimeDisplay extends MediaTextDisplay {
     ) {
       const timesLabel = formatTimesLabel(this);
       updateAriaValueText(this);
-      this.container.innerHTML = timesLabel;
+      // Only update if it changed, timeupdate events are called a few times per second.
+      if (timesLabel !== this.#slot.innerHTML) {
+        this.#slot.innerHTML = timesLabel;
+      }
```
cjpillsbury, line 138: suggestion(non-blocking):
```suggestion
      if (timesLabel !== this.#slot.textContent) {
        this.#slot.textContent = timesLabel;
      }
```

src/js/media-text-display.js
```diff
@@ -42,9 +42,7 @@ template.innerHTML = `
       outline: 0;
     }
   </style>
-  <span id="container">
```
cjpillsbury, line 45: question(non-blocking): Were there no ripple effects on styling when removing this? Any concerns over how this may impact subclasses ability to specify roles and related a11y attributes without sprouting?

src/js/media-current-time-display.js
```diff
@@ -5,18 +5,21 @@ import { MediaUIAttributes } from './constants.js';
 // Todo: Use data locals: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleTimeString
 
 class MediaCurrentTimeDisplay extends MediaTextDisplay {
+  #slot;
```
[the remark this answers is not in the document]
luwes: it feels like such a random public property and a query in the shadow DOM is almost as easy.
yes there is some duplication but it's not too bad

src/js/media-text-display.js
luwes: no styling issues that I saw. it's possible to set any HTML in the slot so should be fine

PR #513, opened 2023-03-30
feat: introduce a playback rates selectmenu

gkatsev:
This uses a playback-rate-listbox under the hood which accepts items *declaratively* only.
```html
<media-playback-rate-listbox media-controller="mc">
  <media-chrome-listitem value="1">1x</media-chrome-listitem>
  <media-chrome-listitem value="1.5">1.5x</media-chrome-listitem>
  <media-chrome-listitem value="2">2x</media-chrome-listitem>
  <media-chrome-listitem value="3">3x</media-chrome-listitem>
</media-playback-rate-listbox>
```

The selectmenu itself accepts a `rates` property like the `playback-rate-button` does. It also defaults to the same defaults.

```html
<media-playback-rate-selectmenu rates="1 2 3 4"></media-playback-rate-selectmenu>
```

This also has some updates to the main listbox class such has the `value` getter/setter, which is available on the HTMLSelectElement as a both as well.

In addition, it changes the behavior of the listitem toggling to only toggle on ctrl-click or command-click, which @luwes pointed out may be confusing in a selectmenu context. Thus, it's still available on a multi-select listbox, but essentially not available in selectmenu.

### DIFF
--- a/src/js/experimental/media-chrome-listbox.js
+++ b/src/js/experimental/media-chrome-listbox.js
@@ -135,8 +135,9 @@ class MediaChromeListbox extends window.HTMLElement {
 
   set value(newValue) {
     const item = this.#items.filter(el => el.value === newValue || el.textContent === newValue)[0];
-    console.log(item);
+
     if (!item) return;
+
     this.#selectItem(item);
   }
 
--- a/src/js/experimental/media-chrome-listbox.js
+++ b/src/js/experimental/media-chrome-listbox.js
@@ -154,7 +154,7 @@ class MediaChromeListbox extends window.HTMLElement {
     const { key } = e;
 
     if (key === 'Enter' || key === ' ') {
-      this.handleSelection(e, this.hasAttribute('aria-multiselectable'));
+      this.handleSelection(e, this.hasAttribute('aria-multiselectable') && this.getAttribute('aria-multiselectable') === 'true');
     } else {
       this.handleMovement(e);
     }
@@ -286,7 +286,7 @@ class MediaChromeListbox extends window.HTMLElement {
   }
 
   #selectItem(item, toggle) {
-    if (this.getAttribute('aria-multiselectable') !== 'true') {
+    if (!this.hasAttribute('aria-multiselectable') || this.getAttribute('aria-multiselectable') !== 'true') {
       this.#assignedElements.forEach(el => el.setAttribute('aria-selected', 'false'));
     }
 
@@ -359,7 +359,7 @@ class MediaChromeListbox extends window.HTMLElement {
     this.#items.forEach(el => el.setAttribute('tabindex', '-1'));
     item.setAttribute('tabindex', '0');
 
-    this.handleSelection(e, e.metaKey || e.ctrlKey);
+    this.handleSelection(e, this.hasAttribute('aria-multiselectable') && this.getAttribute('aria-multiselectable') === 'true');
   }
 
   #searchItem(key) {
--- a/src/js/experimental/media-chrome-listbox.js
+++ b/src/js/experimental/media-chrome-listbox.js
@@ -179,7 +179,7 @@ class MediaChromeListbox extends window.HTMLElement {
   }
 
   #keydownListener = (e) => {
-    const { key, metaKey, altKey } = e;
+    const { key, altKey } = e;
 
     if (altKey) {
       this.removeEventListener('keyup', this.#keyupListener);
@@ -188,6 +188,7 @@ class MediaChromeListbox extends window.HTMLElement {
 
     if (key === 'Meta') {
       this.#metaPressed = true;
+      return;
     }
 
     // only prevent default on used keys
--- a/src/js/experimental/media-chrome-listbox.js
+++ b/src/js/experimental/media-chrome-listbox.js
@@ -129,6 +129,17 @@ class MediaChromeListbox extends window.HTMLElement {
     return this.#items.filter(el => el.getAttribute('aria-selected') === 'true');
   }
 
+  get value() {
+    return this.selectedOptions[0].value || this.selectedOptions[0].textContent;
+  }
+
+  set value(newValue) {
+    const item = this.#items.filter(el => el.value === newValue || el.textContent === newValue)[0];
+    console.log(item);
+    if (!item) return;
+    this.#selectItem(item);
+  }
+
   focus() {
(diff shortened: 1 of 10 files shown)


src/js/experimental/media-chrome-listbox.js
```diff
@@ -234,19 +246,27 @@ class MediaChromeListbox extends window.HTMLElement {
...
+  handleSelection(e, toggle) {
     const item = this.#getItem(e);
 
     if (!item) return;
 
-    const selected = item.getAttribute('aria-selected') === 'true';
+    this.#selectItem(item, toggle);
+  }
 
+  #selectItem(item, toggle) {
```
gkatsev, line 257: allows to *always* select an item, or toggle its selected state. This means that via the mouse, it'll only toggle with ctrl/command, but will always toggle with space/enter. Probably still needs to be tweaked some more, but good enough for this PR.

src/js/experimental/media-chrome-selectmenu.js
```diff
@@ -115,17 +115,17 @@ class MediaChromeSelectMenu extends window.HTMLElement {
   }
 
   #handleChange_() {
-    this.#toggle();
+    this.#toggle(true);
```
gkatsev, line 118: if we get a change event, we just want to close the selectmenu as a new item has been selected. We don't want to accidentally open it if it's been programmatically set.

src/js/experimental/media-playback-rate-listbox.js
```diff
@@ -0,0 +1,77 @@
...
+    ];
+  }
+
+  constructor() {
+    super({slotTemplate});
+  }
+
+  attributeChangedCallback(attrName, oldValue, newValue) {
+    if (attrName === MediaUIAttributes.MEDIA_PLAYBACK_RATE && oldValue !== newValue) {
+      this.value = newValue;
```
gkatsev, line 34: this is the magic that updates the selected item when the playback rate changes.

src/js/experimental/media-playback-rate-selectmenu.js
```diff
@@ -0,0 +1,74 @@
...
+
+    playbackrateButton.preventClick = true;
+
+    const playbackrateListbox = document.createElement('media-playback-rate-listbox');
+    playbackrateListbox.part.add('listbox');
+    playbackrateListbox.setAttribute('exportparts', 'listitem');
+
+    DEFAULT_RATES.forEach(rate => {
+      playbackrateListbox.append(createItem(rate));
+    });
```
gkatsev, line 39: setup the default playback rates

src/js/experimental/media-playback-rate-selectmenu.js
```diff
@@ -0,0 +1,74 @@
...
+
+    if (attrName === 'rates' && oldValue !== newValue) {
+      const listbox = this.shadowRoot.querySelector('media-playback-rate-listbox');
+
+      listbox.textContent = '';
+
+      const rates = newValue.trim().split(' ');
+
+      rates.forEach(rate => {
+        listbox.append(createItem(rate));
```
gkatsev, line 61: update to the requested playback rates

src/js/experimental/media-chrome-listbox.js
gkatsev: I've updated the keypress handling to only allow toggling with the keyboard when the listbox is multi-selectable.

devilshezi0 (approved):

devilshezi0:
> This uses a playback-rate-listbox under the hood which accepts items *declaratively* only.
> ```html
> <media-playback-rate-listbox media-controller="mc">
>   <media-chrome-listitem value="1">1x</media-chrome-listitem>
>   <media-chrome-listitem value="1.5">1.5x</media-chrome-listitem>
>   <media-chrome-listitem value="2">2x</media-chrome-listitem>
>   <media-chrome-listitem value="3">3x</media-chrome-listitem>
> </media-playback-rate-listbox>
> ```
> 
> The selectmenu itself accepts a `rates` property like the `playback-rate-button` does. It also defaults to the same defaults.
> 
> ```html
> <media-playback-rate-selectmenu rates="1 2 3 4"></media-playback-rate-selectmenu>
> ```
> 
> This also has some updates to the main listbox class such has the `value` getter/setter, which is available on the HTMLSelectElement as a both as well.
> 
> In addition, it changes the behavior of the listitem toggling to only toggle on ctrl-click or command-click, which @luwes pointed out may be confusing in a selectmenu context. Thus, it's still available on a multi-select listbox, but essentially not available in selectmenu.

gkatsev:
I actually updated it to only allow toggling when multi-selectable. Also, added a flag to `toggleSubsCaps` to force only turning off.

src/js/experimental/media-chrome-listbox.js
```diff
@@ -126,6 +127,18 @@ class MediaChromeListbox extends window.HTMLElement {
     return this.#items.filter(el => el.getAttribute('aria-selected') === 'true');
   }
 
+  get value() {
+    return this.selectedOptions[0].value || this.selectedOptions[0].textContent;
+  }
+
+  set value(newValue) {
+    const item = this.#items.filter(el => el.value === newValue || el.textContent === newValue)[0];
```
luwes, line 135: nit (non-b): cause use Array.find

src/js/experimental/media-playback-rate-listbox.js
```diff
@@ -0,0 +1,73 @@
...
+    } else if (attrName === 'aria-multiselectable') {
+      // diallow aria-multiselectable
+      this.removeAttribute('aria-multiselectable');
+      console.warn("Captions List doesn't currently support multiple selections. You can enable multiple items via the media.textTrack API.");
```
luwes, line 34: could this comment be removed for playbackrate? that's a string of 0.12kB 😛

luwes:
looks good!

src/js/experimental/media-chrome-listbox.js
gkatsev: I keep foregtting that's a thing, haha

src/js/experimental/media-playback-rate-listbox.js
gkatsev: Oops, this can definitely be trimmed, at the very least.

luwes (approved):
LGTM 👍

PR #514, opened 2023-03-30
fix: toggle selectmenu via keyboard

gkatsev:
This makes preventClick apply to keypresses in media-chrome-button.

It also has selectmenu listen to keypress events on the button and the listbox so that it can open and close the menu as expected via Enter/Space and Escape.

### DIFF
--- a/src/js/experimental/media-chrome-selectmenu.js
+++ b/src/js/experimental/media-chrome-selectmenu.js
@@ -103,10 +103,10 @@ class MediaChromeSelectMenu extends window.HTMLElement {
 
     this.#listboxSlot = this.shadowRoot.querySelector('slot[name=listbox]');
     this.#listboxSlot.addEventListener('slotchange', () => {
-      this.#listbox.removeEventListener('change', this.#handleChange);
+      this.disable();
       // update listbox reference if necessary
       this.#listbox = this.#listboxSlot.assignedElements()[0] || this.#listbox;
-      this.#listbox.addEventListener('change', this.#handleChange);
+      this.enable();
     });
   }
 
--- a/src/js/experimental/media-chrome-selectmenu.js
+++ b/src/js/experimental/media-chrome-selectmenu.js
@@ -110,6 +110,36 @@ class MediaChromeSelectMenu extends window.HTMLElement {
     });
   }
 
+  // NOTE: There are definitely some "false positive" cases with multi-key pressing,
+  // but this should be good enough for most use cases.
+  #keyupListener = (e) => {
+    const { key } = e;
+
+    if (!this.keysUsed.includes(key)) {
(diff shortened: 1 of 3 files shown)


src/js/experimental/media-chrome-selectmenu.js
```diff
@@ -110,6 +110,36 @@ class MediaChromeSelectMenu extends window.HTMLElement {
...
+    if (!this.keysUsed.includes(key)) {
+      this.removeEventListener('keyup', this.#keyupListener);
+      return;
+    }
+
+    const isButton = e.composedPath().includes(this.#button);
+
+    // only allow Enter/Space on the button itself and not on the listbox
+    // and allow hiding the menu when pressing Escape when focused on the listbox
+    if (isButton && (key === 'Enter' || key === ' ')) {
```
luwes, line 127: if I use space to open the page scrolls down so we need a event.preventDefault() here

gkatsev: hm... not happening for me locally, but happy to add one. Has to go in the `keydownListener`, though.

gkatsev: updated.

luwes (changes requested):
these still need fixing I think

- escape doesn't work if the focus is in the listbox
- the whole listbox can get focus somehow, see screenshot 
SCR-20230331-d6f
- I expect the listbox to close if I click out of the listbox, close on blur basically. but the video shouldn't start playing if the click is on the gesture receiver or the play button
- the bug of selecting 3 times Japanese for example is still there, it eventually selects English?

gkatsev:
> * escape doesn't work if the focus is in the listbox

Hm... weird, I'll take another look but it should've been accounted for.

> * the whole listbox can get focus somehow, see screenshot

This is expected. When the listbox opens, it gets focus.

> I expect the listbox to close if I click out of the listbox, close on blur basically. but the video shouldn't start playing if the click is on the gesture receiver or the play button

This is a tricky one, since we can't prevent the action on other components from within the selectmenu itself. Though, we definitely can make it close on a click outside the component.
This probably deserves a larger media-chrome discussion, as we may get more things that we may want that handle this.

> the bug of selecting 3 times Japanese for example is still there, it eventually selects English?

This is being handled elsewhere. There's some work on this in #513.

gkatsev:
> > * escape doesn't work if the focus is in the listbox
> 
> Hm... weird, I'll take another look but it should've been accounted for.

Weird, so, it works for the elements in the control bar but not the other ones. It's because that first CC button is still a chrome-selectmenu with a slotted captions-listbox. So, the fix is to make sure that the keypress listener is being re-added on a slotchange.

luwes (approved):

PR #517, opened 2023-04-02
Media Controller Refactor 

heff:
No functional changes, just reorganizing and moving things out of the file to make it easier to split out controller logic from the custom element.

The commits are intentional steps in the process (so far).

* Moved util and platform tests to util files
* Refactored _mediaStatePropagators, delegates, and some floating functions into one MediaUIStates object with a clear mapping between the state name, the getter, and related events.
* Moved UI event handlers to its own MediaUIRequestHandlers object outside of MediaCotroller
* Reapplied state & request objects back to the controller


src/js/media-controller.js
```diff
@@ -38,6 +46,394 @@ const ButtonPressedKeys = ['ArrowLeft', 'ArrowRight', 'Enter', ' ', 'f', 'm', 'k
 const DEFAULT_SEEK_OFFSET = 10;
 const DEFAULT_TIME = 0;
 
+const MediaUIStates = {
+  paused: {
+    get: function(controller){
+      const { media } = controller;
+
+      return (media) ? media.paused : true;
```
luwes, line 54: nit (non-b): unneeded parentheses

luwes:
LGTM so far 👍

heff:
At a decent merge point now, with other 1.0 work looming.

Followup steps:
* Wrap up [controller/container decision](https://github.com/muxinc/media-chrome/discussions) to guide next move here
* Build the MediaController object on its own, then reapply based on the previous bullet.

src/js/utils/platform-tests.js
```diff
@@ -0,0 +1,63 @@
+import { window, document } from './server-safe-globals.js';
```
cjpillsbury, line 1: praise: 👌 clean.

heff: 🙏

src/js/media-controller.js
```diff
@@ -38,357 +52,724 @@ const ButtonPressedKeys = ['ArrowLeft', 'ArrowRight', 'Enter', ' ', 'f', 'm', 'k
...
+      return media.currentTime;
+    },
+    mediaEvents: ['timeupdate', 'loadedmetadata']
+  },
+  MEDIA_DURATION: {
+    get: function(controller){
+      const { media } = controller;
+
+      // TODO: What if duration is infinity/live? (heff)
+      if (!media || !Number.isFinite(media.duration)) {
```
heff, line 156: @cjpillsbury we shouldn't return NaN here when duration is Infinity (live), right?

src/js/media-controller.js
```diff
@@ -629,36 +836,33 @@ class MediaController extends MediaContainer {
   mediaSetCallback(media) {
     super.mediaSetCallback(media);
 
+    // TODO: What does this do? At least add comment, maybe move to media-container
```
heff, line 839: @gkatsev what does this do?

src/js/media-controller.js
```diff
@@ -675,39 +879,38 @@ class MediaController extends MediaContainer {
...
     this.propagateMediaState(MediaUIAttributes.MEDIA_PAUSED, true);
   }
 
   propagateMediaState(stateName, state) {
-    if (arguments.length === 1) {
-      state = Delegates[stateName](this);
-    }
-
     propagateMediaState(this.mediaStateReceivers, stateName, state);
+    // TODO: I don't think we want these events to bubble? Video element states don't. (heff)
```
heff, line 913: @cjpillsbury confirming it's ok to turn off bubbling here. I don't have a strong defense but it feels wrong to bubble these state change events

src/js/media-controller.js
cjpillsbury: There are some assumptions in our time-related UI elements/Media State Receivers that I believe would need to be cleaned up. tl;dr - we prefer `seekable` but aren't guaranteed that the slotted media element supports `seekable`

src/js/media-controller.js
```diff
@@ -38,357 +52,724 @@ const ButtonPressedKeys = ['ArrowLeft', 'ArrowRight', 'Enter', ' ', 'f', 'm', 'k
...
+    // We want to let the user know that the media started playing at any point (`media-has-played`).
+    // Since these propagators are all called when boostrapping state, let's verify this is
+    // a real playing event by checking that 1) there's media and 2) it isn't currently paused.
+    get: function(controller) {
+      const { media } = controller;
+
+      if (!media) return false;
+
+      // TODO: Would seem to leave room for issue if this is called
+      // after a video has played but is currently paused.
```
heff, line 74: @cjpillsbury I'm not gonna try to fix this as part of this PR, but can you confirm this is a potential issue? e.g.
* Load a player
* press play
* press pause
* Add a poster element

The poster element would get has_played = false

src/js/utils/utils.js
```diff
@@ -34,3 +34,10 @@ export function isNumericString(str) {
   // @ts-ignore
   return !isNaN(str) && !isNaN(parseFloat(str));
 }
+
+/**
+ * Returns a promise that will resolve after passed ms.
+ * @param  {number} ms
+ * @return {Promise}
+ */
+export const delay = (ms) => new Promise((resolve) => setTimeout(resolve, ms));
```
cjpillsbury, line 43: nitpick(blocking): missing EOF newline

src/js/media-controller.js
heff: Got it. It feels odd to change what the media element is providing, but I guess it's fair to say at the UI level, if we're already using Infinity to set StreamType otherwise, then it simplifies UI logic to just return a falsey value here.

cjpillsbury (commented):
question(non-blocking): any concerns with taking the code you split out and moving it into its own JS module? Even if we plan on doing non-trivial refactors to it during our v1.0 effort, this would likely make merge conflicts less of an issue.

cjpillsbury:
Aside from us banging on this for smoke testing around the edges (particularly in non-desktop & streaming cases), this looks great as an initial step to create some architecture seams!

src/js/media-controller.js
cjpillsbury: Let's make sure we don't conflate "has loaded" with "has played". I believe this is an example where we need to "locally store and reset state", since it's not something that can be deterministically derived from the media element and instead is "historic" in nature. lmk if you want a "reliable but maybe a quick hack/not final form" solution. Also, glad you called this out, as this is likely broken. Making me a bit nervous on not having a simple diff to look at for these cases... 😅

src/js/media-controller.js
cjpillsbury: I actually agree 💯. We can instead consider an "appropriate response/UI update" in the components themselves? I'm down for that.

src/js/media-controller.js
```diff
@@ -38,357 +52,724 @@ const ButtonPressedKeys = ['ArrowLeft', 'ArrowRight', 'Enter', ' ', 'f', 'm', 'k
...
+      return castState;
+    },
+    mediaEvents: ['entercast','leavecast','castchange']
+  },
+  MEDIA_AIRPLAY_UNAVAILABLE: {
+    // NOTE: only adding this if airplay is supported, in part to avoid unnecessary battery consumption per
+    // Apple docs recommendations (See: https://developer.apple.com/documentation/webkitjs/adding_an_airplay_button_to_your_safari_media_controls)
+    // For a more advanced solution, we could monitor for media state receivers that "care" about airplay support and add/remove
+    // whenever these are added/removed. (CJP)
+    // NOTE: I don't think only adding this if supported helps, so making this match others.
```
heff, line 344: @cjpillsbury see this specific update, allowing it not be selectively added as a listener like it was before.

src/js/media-controller.js
```diff
@@ -38,357 +52,724 @@ const ButtonPressedKeys = ['ArrowLeft', 'ArrowRight', 'Enter', ' ', 'f', 'm', 'k
...
+
+      // Cast state: NO_DEVICES_AVAILABLE, NOT_CONNECTED, CONNECTING, CONNECTED
+      if (castState.includes('CONNECT')) {
+        // TODO: Move to non attr specific value
+        return undefined;
+      } else {
+        return AvailabilityStates.UNAVAILABLE;
+      }
+    },
+    mediaEvents: ['castchange']
```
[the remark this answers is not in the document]
cjpillsbury: may also consider having a shorthand `chromecastSupported` like `airplaySupported` solution, ☝️

src/js/media-controller.js
```diff
@@ -38,357 +52,724 @@ const ButtonPressedKeys = ['ArrowLeft', 'ArrowRight', 'Enter', ' ', 'f', 'm', 'k
...
+      if (!e) return undefined;
+      // TODO: Switch to non attr specific value
+
+      if (e.availability === 'available') {
+        return undefined;
+      } else if (e.availability === 'not-available') {
+        return AvailabilityStates.UNAVAILABLE;
+      }
+    },
+    mediaEvents: ['webkitplaybacktargetavailabilitychanged']
```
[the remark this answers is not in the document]
heff: That's not bad. 👍 Does it matter though? If airplay isn't supported then adding the listener shouldn't hurt anything, right? The hypothetical performance issue comes from attaching the listener to something that does support airplay and leaving it there.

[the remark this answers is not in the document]
cjpillsbury: you're likely right that that's the biggest concern and the reason for the callout (and one that we don't really have a good solution for without being overly presumptuous?). 😎 if you want to reject suggestion for simplicity (even if it means we're listening for things we know will never happen)

src/js/media-controller.js
gkatsev: This makes the media element focusable if a user didn't specify a tabindex already. -1 means that it'll be focusable via a mouse click or programmatically, but not be part of the tab order.

I believe it's necessary to keep focus inside the player for things like keyboard shortcuts.

src/js/media-controller.js
```diff
@@ -38,357 +52,724 @@ const ButtonPressedKeys = ['ArrowLeft', 'ArrowRight', 'Enter', ' ', 'f', 'm', 'k
...
+
+      // TODO: move to non attr specific value
+      return undefined;
+    },
+    // Give a little time for the volume support promise to run
+    mediaEvents: ['loadstart']
+  },
+  MEDIA_CAPTIONS_LIST: {
+    get: function(controller) {
+      // TODO: Move to non attr specific values
```
gkatsev, line 415: Should be fine to leave as is for now. These would get updated when we combine captions and subtitles lists.

gkatsev:
Looks like there are conflicts now. Hopefully, they're easy to resolve.

cjpillsbury (changes requested):
issue(blocker): We should move the "abstracted, javascript only" parts of the state management into its own module to mitigate merge conflict issues going forward (particularly on our Media Chrome v1.0 push).

src/js/media-controller.js
cjpillsbury: @heff since this appears to be a regression edge case, should we treat this as out of scope for this initial pass/PR?

src/js/controller.js
```diff
@@ -0,0 +1,771 @@
...
+} from './constants.js';
+
+import {
+  stringifyTextTrackList,
+  getTextTracksList,
+  updateTracksModeTo,
+} from './utils/captions.js';
+
+let volumeSupported;
+export const volumeSupportPromise = hasVolumeSupportAsync().then((supported) => {
```
cjpillsbury, line 27: NOTE: Currently exporting this based on our status quo impl of checking/propagating volume support in MediaController.

gkatsev (approved):

luwes (approved):
LGTM 👍

PR #519, opened 2023-04-04
fix: media-theme element lazy doc append

luwes:
this change fixes a bug where if the media-theme element was not added to the document yet it would throw a type error.

- adds support for rendering the template on connectedCallback if it didn't happen yet
- makes the template URL fetching a bit more strict on valid URL's so it doesn't try to request on just word URL's
  it requires a `/` somewhere in the URL. `./`, `/`, `https://`

### DIFF
--- a/src/js/media-theme-element.js
+++ b/src/js/media-theme-element.js
@@ -181,14 +181,8 @@ export class MediaThemeElement extends window.HTMLElement {
           this.#template = template;
           this.createRenderer();
         })
-        .catch(() => {
-          console.warn(`"${templateId}" is not a valid URL to a template file.`);
-        });
-
-      return;
+        .catch(console.error);
     }
-
-    console.warn(`"${templateId}" is not a valid template element ID or a valid URL to a template file.`);
   }
 
   createRenderer() {
--- a/src/js/media-theme-element.js
+++ b/src/js/media-theme-element.js
@@ -39,6 +39,7 @@ export class MediaThemeElement extends window.HTMLElement {
   renderer;
   #template;
   #prevTemplate;
+  #prevTemplateId;
 
   constructor() {
     super();
@@ -103,6 +104,7 @@ export class MediaThemeElement extends window.HTMLElement {
   }
 
   set template(element) {
+    this.#prevTemplateId = null;
     this.#template = element;
     this.createRenderer();
   }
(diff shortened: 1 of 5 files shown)


src/js/media-theme-element.js
```diff
@@ -140,30 +140,43 @@ export class MediaThemeElement extends window.HTMLElement {
     return props;
   }
 
-  async attributeChangedCallback(attrName, oldValue, newValue) {
+  attributeChangedCallback(attrName, oldValue, newValue) {
```
cjpillsbury, line 143: 👍

src/js/media-theme-element.js
```diff
@@ -140,30 +140,43 @@ export class MediaThemeElement extends window.HTMLElement {
...
+      request(templateId)
+        .then((data) => {
+          const template = document.createElement('template');
+          template.innerHTML = data;
+
+          this.#template = template;
+          this.createRenderer();
+        })
+        .catch(() => {
+          console.warn(`"${templateId}" is not a valid template element ID or a valid URL to a template file.`);
```
cjpillsbury, line 178: nitpick(non-blocking): This will not happen when relying on templateId as element ID. Consider removing that text for clarity/precision.

src/js/media-theme-element.js
```diff
@@ -140,30 +140,43 @@ export class MediaThemeElement extends window.HTMLElement {
...
+        });
```
cjpillsbury, line 179: suggestion(non-blocking): If neither condition above is met, consider adding an additional warning here.

src/js/themes/demuxed-2022.js
```diff
@@ -387,7 +387,7 @@ class MediaThemeDemuxed extends MediaThemeElement {
     resizeObserver.observe(this.shadowRoot.querySelector('media-controller'));
   }
 
-  async attributeChangedCallback(name, _oldValue, newValue) {
+  attributeChangedCallback(name, _oldValue, newValue) {
```
cjpillsbury, line 390: 👍

test/unit/media-theme.spec.js
```diff
@@ -3,6 +3,27 @@ import '../../src/js/media-theme-element.js';
 import '../../src/js/index.js';
 
 describe('<media-theme>', () => {
+  it(`<media-theme> with template attribute works w/ delayed document append`, async () => {
```
cjpillsbury, line 6: praise: regression test 🎉

cjpillsbury (approved):
Just a couple of minor non-blocking callouts. Otherwise LGTM!

cjpillsbury (approved):
👌

PR #524, opened 2023-04-06
fix: make DSD (declarative shadow dom) compatible

luwes:
Related #184

This makes Media Chrome DSD compatible in case anyone is putting in declarative shadow dom.
It creates shadow DOM without any JS needed. See example:

```html
<host-element>
  <template shadowrootmode="open">
    <slot></slot>
  </template>
  <h2>Light content</h2>
</host-element>
```

Supported now by Chrome and Safari 16.4! 

https://developer.chrome.com/en/articles/declarative-shadow-dom/
https://webkit.org/blog/13966/webkit-features-in-safari-16-4/#web-components

### DIFF
--- a/src/js/experimental/media-chrome-listbox.js
+++ b/src/js/experimental/media-chrome-listbox.js
@@ -55,21 +55,24 @@ class MediaChromeListbox extends window.HTMLElement {
   constructor(options = {}) {
     super();
 
-    const shadow = this.attachShadow({ mode: 'open' });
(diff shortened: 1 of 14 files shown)


AdamJaggard (approved):
LGTM. As this is only a stepping stone towards full SSR support I can't see any immediate issue with it.

cjpillsbury (approved):
I still think it would be a good idea to add simple regression tests here, but I won't treat that as a blocker.

PR #525, opened 2023-04-06
feat: remove deprecated, move experimental files

luwes:
it's a feat version because "breaking" below v1.
I have my doubts anyone is using those files but better safe.


cjpillsbury (approved):
🚢 🇮🇹

PR #528, opened 2023-04-11
Improve CSS vars

### LINKED ISSUE #453: Add CSS var to set the font family on media-time-display
Seems like this is missing and it should be a thing. I don't see why it wouldn't be?

cc: @davekiss

luwes:
Closes #473 
Fixes #453 

- Fixes some outdated documenting 
- Makes all colors the same uniform rgb() syntax
- Adds `--media-primary-color`
- Adds `--media-secondary-color`
- Adds `--media-tertiary-color`
- Adds `--media-text-color`
- Adds `--media-font`, `--media-font-weight`, `--media-font-size`, `--media-font-family`


src/js/media-loading-indicator.js
```diff
@@ -52,7 +52,7 @@ template.innerHTML = `
 svg, img, ::slotted(svg), ::slotted(img) {
   width: var(--media-loading-icon-width, 100px);
   height: var(--media-loading-icon-height);
-  fill: var(--media-icon-color, #fff);
+  fill: var(--media-icon-color, var(--media-primary-color, rgb(238 238 238)));
```
gkatsev, line 55: is it an issue that we're changing the default from white to a greyish color?

docs/src/pages/en/styling.md
```diff
@@ -23,18 +23,19 @@ Our current styling architecture is still quite nascent and is very likely to un
...
+| Name                               | CSS Property  | Default Value                                      | Description                                                                                       | Notes                                                                                                                                                                |
+| ---------------------------------- | ------------- | -------------------------------------------------- | ------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `--media-control-background`       | `background`  | `var(--media-secondary-color, rgb(20 20 30 / .7))` | background color of the component                                                                 | Applies to other components as well ([See notes below \*\*](#notes))                                                                                                 |
```
gkatsev, line 28: should `--media-primary-color` and `--media-secondary-color` get their own section at the top?

gkatsev: Looks like the new font css vars aren't documented either. Thinking about this, might be fine to skip for now since we have an upcoming effort to update documentation for this.

src/js/media-chrome-button.js
```diff
@@ -4,37 +4,33 @@ import { window, document } from './utils/server-safe-globals.js';
 
 const template = document.createElement('template');
 
-template.innerHTML = `
+template.innerHTML = /*html*/`
 <style>
   :host {
+    font: var(--media-font,
```
gkatsev, line 10: At first, I was worried that moving into the shorthand property it'll cause issues if someone were to set a specific property or the font property on the element, but after trying it out, that seems unfounded and working as expected.

gkatsev, line 10: I think I would still slightly prefer individual properties over the font shorthand. It's more explicit and would help with searchability, since you could find it via grepping for say `line-height`.

gkatsev (commented):
Overall, 👍🏻

src/js/media-chrome-button.js
luwes: cool will update. I don't feel so strongly about it

luwes: oh actually then we can't use `--media-font` as a CSS var for setting all at once which includes some more things like 

[font-stretch](https://developer.mozilla.org/en-US/docs/Web/CSS/font-stretch)
[font-style](https://developer.mozilla.org/en-US/docs/Web/CSS/font-style)
[font-variant](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant)

I think that extra function and it taking less CSS makes it a better choice. what do you think?

gkatsev: Theoretically, we could do some fancy var stuff to unset the other properties if --media-font is set, but probably too much work. Is --media-font needed?
While, I do think the individual properties are a bit better, I'm OK with keeping it as the shorthand.

luwes: > Is --media-font needed?

it would give devs the full flexibility of all font css props if needed with small addition in code size. 
seems like a good devex

cjpillsbury (approved):
This is awesome. One non-blocking callout - looks like the new font css vars aren't applied to all places where we have text: Immediately comes to mind: media-playback-rate-button, media-live-button (though there may be others I'm missing?)

luwes:
the buttons are def covered via `media-chrome-button`. should be good

PR #530, opened 2023-04-12
example: improve captions experience w/ CSS snippet

luwes:
TEST URL: https://media-chrome-git-fork-luwes-improve-cc-example-mux.vercel.app/examples/vanilla/advanced.html

- this makes the native captions look better and make them less intrusive on Chrome. More uniform w/ Safari.
- moves the captions up when the control bar is visible for Chrome and Safari.

this seemed like a nice quick thing to do to improve the experience and covers +85% of browser usage!
this could also be a guide in the docs maybe.

SCR-20230412-jx2

### DIFF
--- a/examples/vanilla/advanced.html
+++ b/examples/vanilla/advanced.html
@@ -23,6 +23,21 @@
         display: none;
       }
 
+      video::-webkit-media-text-track-container {
+        transition: transform .25s;
+      }
+
+      media-controller:is([media-paused],:not([user-inactive])) video::-webkit-media-text-track-container {
+        transform: translateY(-50px);
+      }
+
+      video::-webkit-media-text-track-display {
+        width: auto !important;
+        left: 50% !important;
+        transform: translateX(-50%);
+        padding: .4em;
+      }
+
       .examples {
         margin-top: 20px;
       }


examples/vanilla/advanced.html
```diff
@@ -23,6 +23,21 @@
...
 
+      video::-webkit-media-text-track-container {
+        transition: transform .25s;
+      }
+
+      media-controller:is([media-paused],:not([user-inactive])) video::-webkit-media-text-track-container {
+        transform: translateY(-50px);
+      }
+
+      video::-webkit-media-text-track-display {
```
gkatsev, line 34: I think generally, I wouldn't recommend doing this. Particularly if users make changes to how they want their captions displayed. Chrome respects the macOS settings.


without these changes
Screen Shot 2023-04-12 at 15 46 15

with these changes
Screen Shot 2023-04-12 at 15 46 09

having the transform translateY is good, though.

heff:
Nice solve. 👍 

There's a bit of an aesthetic disconnect between the controls fading and the captions movement. I think it'd ultimately look best if they were moving up and down together, but that'd be a bigger change. Second best would be to wait until the controls were fully faded out before moving the captions down. If we bake the translateY part in to media-container could we use the `autohide` setting to do that?

> could also be a guide in the docs maybe

Yeah, I don't love this in the advanced example. i.e. I want to try to get rid of the needed CSS in the examples rather than add more to it.  I feel like this would be ok to add in the media-container CSS and bake in by default. At least the translateY part.

PR #535, opened 2023-04-17
fix: inactive live button

luwes:
- adds aria-disabled="true" to media-live-button when it's playing live
- adds aria-label's based on live status


cjpillsbury (approved):
👌 LGTM

PR #537, opened 2023-04-17
feat: Migrate all attributes to lowercase ('smushedcase').

cjpillsbury:
Migrates all well-defined attributes in Media Chrome elements from kebab-case to lowercase (aka "smushedcase").


examples/nextjs-with-typescript/pages/index.tsx
```diff
@@ -1,31 +1,16 @@
 import type { NextPage } from 'next';
```
cjpillsbury, line 1: NOTE: Need to revert this. Hold over from presentation testing.

src/js/index.d.ts
```diff
@@ -0,0 +1,200 @@
+export * as constants from "./constants.js";
```
cjpillsbury, line 1: NOTE: Going to drop this in this PR, but this was the quick fix solution to get all of the TypeScript types in place for React. We should instead just completely revamp how the React wrappers work in our v1.0 effort.

src/js/media-chrome-button.js
```diff
@@ -129,14 +129,14 @@ class MediaChromeButton extends window.HTMLElement {
   enable() {
     this.addEventListener('click', this.#clickListener);
     this.addEventListener('keydown', this.#keydownListener);
-    this.setAttribute('tabindex', '0');
+    this.tabIndex = 0;
```
cjpillsbury, line 132: NOTE: Moving to using the `tabIndex` prop instead for forward-looking dreams.

src/js/media-container.js
```diff
@@ -47,37 +63,51 @@ template.innerHTML = `
...
-     * if gestures are disabled, don't accept pointer-events
-     */''}
-    :host(:not([audio])[gestures-disabled]) ::slotted([slot=gestures-chrome]),
-    :host(:not([audio])[gestures-disabled]) media-gesture-receiver[slot=gestures-chrome] {
+    ${
+      /*
+       * if gestures are disabled, don't accept pointer-events
+       */ ''
+    }
+    :host(:not([${Attributes.AUDIO}])[${
```
cjpillsbury, line 81: oof formatter really did a number on this one. Will unmangle.

src/js/media-container.js
```diff
@@ -89,33 +119,39 @@ template.innerHTML = `
...
     }
 
-    ${/* Hide controls when inactive, not paused, not audio and auto hide not disabled */''}
-    :host([user-inactive]:not([${MediaUIAttributes.MEDIA_PAUSED}]):not([${MediaUIAttributes.MEDIA_IS_CASTING}]):not([audio])) ::slotted(:not([slot=media]):not([no-auto-hide])) {
+    ${
+      /* Hide controls when inactive, not paused, not audio and auto hide not disabled */ ''
+    }
+    :host([${Attributes.USER_INACTIVE}]:not([${
+  MediaUIAttributes.MEDIA_PAUSED
+}]):not([${MediaUIAttributes.MEDIA_IS_CASTING}]):not([${
```
cjpillsbury, line 152: same for this uggo.

src/js/media-time-display.js
```diff
@@ -12,8 +19,8 @@ const ButtonPressedKeys = ['Enter', ' '];
 const DEFAULT_TIMES_SEP = '&nbsp;/&nbsp;';
 
 const formatTimesLabel = (el, { timesSep = DEFAULT_TIMES_SEP } = {}) => {
-  const showRemaining = el.getAttribute('remaining') != null;
-  const showDuration = el.getAttribute('show-duration') != null;
+  const showRemaining = el.showRemaining;
```
cjpillsbury, line 22: NOTE: This feels like scope creep for this round. Hoping for a gut check from others if you think I should drop it (still around in other branches).

src/js/themes/microvideo.js
```diff
@@ -102,7 +102,7 @@ template.innerHTML = /*html*/`
     align-items: start;
   }
 
-  :host([control-bar-place$="end"]) media-control-bar {
+  :host([controlbarplace$="end"]) media-control-bar {
```
cjpillsbury, line 105: NOTE: Should discuss - normalize so official attrs for official themes are also lowercase/"smushedcase" (See below for at least one particular friction point/tradeoff).

src/js/themes/microvideo.js
```diff
@@ -440,7 +440,7 @@ template.innerHTML = /*html*/`
 </template>
 
 <media-controller
-  style="--_control-bar-place-self:{{controlBarPlace ?? 'unset'}}"
+  style="--_control-bar-place-self:{{controlbarplace ?? 'unset'}}"
```
cjpillsbury, line 443: NOTE: Because we use JS to translate attr -> prop, using all lowercase/"smushedcase" here doesn't allow us to use `camelCase` to translate (since there are no delimiters to match against with the new names).

cjpillsbury: We may want to treat this as a separate effort and descope it for this initial round of changes.

src/js/media-playback-rate-button.js
```diff
@@ -20,32 +20,35 @@ class MediaPlaybackRateButton extends MediaChromeButton {
     return [
       ...super.observedAttributes,
       MediaUIAttributes.MEDIA_PLAYBACK_RATE,
-      'rates',
+      Attributes.RATES,
     ];
   }
 
+  /** @type number[] | undefined */
+  #_rates;
```
luwes, line 28: can we leave the underscore, use `#rates` instead? 

don't really see a reason for an added underscore.

cjpillsbury: Yup we can def drop the underscore. ❤️

src/js/media-time-display.js
```diff
@@ -59,12 +66,17 @@ class MediaTimeDisplay extends MediaTextDisplay {
...
+      Attributes.REMAINING,
+      Attributes.SHOW_DURATION,
       'disabled',
     ];
   }
 
+  /** @type boolean | undefined */
+  #_showDuration;
+  /** @type boolean | undefined */
+  #_remaining;
```
luwes, line 78: here too the underscores

src/js/media-playback-rate-button.js
```diff
@@ -60,12 +63,22 @@ class MediaPlaybackRateButton extends MediaChromeButton {
     super.attributeChangedCallback(attrName, oldValue, newValue);
   }
 
+  get rates() {
```
gkatsev, line 66: this seems like scope-creep.
This getter should return a DOMTokenList, while the setter should accept an array.

cjpillsbury: Cool I agree let's just drop all of the property changes in this PR. @luwes when we re-add we'll be sure to account for your private variable naming convention prefs.

package.json
```diff
@@ -25,7 +25,7 @@
     "build:iife": "esbuild src/js/index.js --bundle --target=es2019 --format=iife --outdir=dist/iife --minify --sourcemap --global-name=MediaChrome && yarn size",
     "build:react": "node ./scripts/react/build.js",
     "build:manifest": "npx @custom-elements-manifest/analyzer analyze --config scripts/custom-elements-manifest.config.js",
-    "build": "yarn build:esm && yarn build:cjs && yarn build:iife && yarn build:react && yarn build:types && yarn build:manifest",
+    "build": "yarn build:types && yarn build:esm && yarn build:cjs && yarn build:iife && yarn build:react && yarn build:manifest",
```
cjpillsbury, line 28: NOTE: switch the order to ensure types are available before build (for dev watch behavior/expectations)

src/js/media-time-display.js
cjpillsbury: Will either remove or rename before merging, depending on team decision here.

src/js/media-container.js
luwes: yup

src/js/media-container.js
cjpillsbury: yessir!

gkatsev: which formatter did this? `yarn format`?

src/js/media-playback-rate-button.js
```diff
@@ -20,25 +20,29 @@ class MediaPlaybackRateButton extends MediaChromeButton {
...
       ...super.observedAttributes,
       MediaUIAttributes.MEDIA_PLAYBACK_RATE,
-      'rates',
+      Attributes.RATES,
     ];
   }
 
+  // NOTE: Adding for TypeScript Errors. Followup should add correct getter/setter & private var (CJP)
+  /** @type number[] | undefined */
+  _rates;
```
gkatsev, line 29: is this used anywhere?

src/js/media-poster-image.js
```diff
@@ -35,7 +40,7 @@ const setBackgroundImage = (el, image) => {
 
 class MediaPosterImage extends window.HTMLElement {
   static get observedAttributes() {
-    return ['placeholder-src', 'src'];
+    return [Attributes.PLACEHOLDER_SRC, Attributes.SRC];
```
gkatsev, line 43: non-blocker: alternative is to do `Object.values(Attributes)`

src/js/media-container.js
cjpillsbury: I think it was some default prettier line length creeping in? Haven't investigated.

src/js/media-poster-image.js
cjpillsbury: decided to not use values just because that's "incidental" and not "definitional" (aka all of the values in `Attributes` _**happen to**_ be `observedAttributes`, but that could change and isn't inherent to their relationship).

gkatsev: Ah, right. Worked here incidentally. Wasn't a blocker or necessary anyway.

PR #539, opened 2023-04-18
fix: add default unavailable CSS to themes

luwes:
I had hoped we could add this CSS to media-container but it's not possible :( 

because those control elements are in the light DOM which could be nested in divs and not reachable via CSS in the shadow DOM.

still nice to have out of the box for themes I think.

### DIFF
--- a/src/js/media-theme-element.js
+++ b/src/js/media-theme-element.js
@@ -13,7 +13,7 @@ const observedMediaAttributes = {
 
 const prependTemplate = document.createElement('template');
 
-prependTemplate.innerHTML = `
+prependTemplate.innerHTML = /*html*/`
   <style>
     :host {
       display: inline-block;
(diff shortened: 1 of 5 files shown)


src/js/media-theme-element.js
```diff
@@ -24,6 +24,16 @@ prependTemplate.innerHTML = `
       width: 100%;
       height: 100%;
     }
+
+    media-controller:not(:is([mediacaptionslist], [mediasubtitleslist])) media-captions-selectmenu,
+    media-captions-button:not(:is([mediacaptionslist], [mediasubtitleslist])),
```
gkatsev, line 29: though(non-blocking): Should we use `:where` instead to limit the selector specificity?

gkatsev (approved):
I think this is reasonable to have in Themes. I don't think we'd want this in the default container CSS, even if there weren't issues around it because we're aiming to have the default components be as unassuming as possible.

src/js/media-theme-element.js
luwes: yes, that's a good idea. will update

PR #545, opened 2023-04-19
docs: custom element docs first draft

luwes:


package.json
```diff
@@ -20,12 +20,12 @@
     "clean": "rimraf dist",
     "lint": "eslint src/js",
     "build:types": "tsc",
-    "build:esm": "esbuild src/js/*.js src/js/*/*.js src/js/*/*/*.js --target=es2019 --format=esm --outdir=dist",
+    "build:esm": "cp -R src/js dist",
```
gkatsev, line 23: I think it's still worth using esbuild to compile away the newest JS syntax to be a bit more compatible.

luwes: yes, this is done because esbuild will always strip the comments (no way around that :() 
which are needed for customelement manifest analyzer. 

I guess we can copy / build manifest and then build esm. might be an ok workaround.

gkatsev: yeah, jsdoc generally ignores if it isn't `/**`. And not sure if tsc and the web component analyzer will pick it up as well.

gkatsev: nice!

package.json
```diff
@@ -19,13 +19,14 @@
...
+    "premanifest": "cp -R src/js/ dist/",
+    "manifest": "npx @custom-elements-manifest/analyzer analyze --config scripts/custom-elements-manifest.config.js",
     "build:types": "tsc",
     "build:esm": "esbuild src/js/*.js src/js/*/*.js src/js/*/*/*.js --target=es2019 --format=esm --outdir=dist",
     "build:cjs": "esbuild src/js/*.js src/js/*/*.js src/js/*/*/*.js --target=es2019 --format=cjs --outdir=dist/cjs",
     "build:iife": "esbuild src/js/index.js --bundle --target=es2019 --format=iife --outdir=dist/iife --minify --sourcemap --global-name=MediaChrome && yarn size",
     "build:react": "node ./scripts/react/build.js",
-    "build:manifest": "npx @custom-elements-manifest/analyzer analyze --config scripts/custom-elements-manifest.config.js",
-    "build": "yarn build:types && yarn build:esm && yarn build:cjs && yarn build:iife && yarn build:react && yarn build:manifest",
+    "build": "run-s build:types build:esm build:cjs build:iife build:react",
```
gkatsev, line 29: should we `run-p` these instead?

src/js/experimental/media-playback-rate-listbox.js
```diff
@@ -3,15 +3,18 @@ import { window, document } from '../utils/server-safe-globals.js';
 import { MediaUIAttributes, MediaUIEvents } from '../constants.js';
 
 const slotTemplate = document.createElement('template');
-slotTemplate.innerHTML = `
+slotTemplate.innerHTML = /*html*/`
   <style>
     media-chrome-listitem {
-      white-space: var(--media-captions-listbox-white-space, nowrap);
+      white-space: var(--media-playback-rate-listbox-white-space, nowrap);
```
gkatsev, line 9: oops. Great find!

gkatsev: We'd probably want to port this over to 0.x

src/js/media-seek-backward-button.js
```diff
@@ -33,9 +33,21 @@ const updateSeekIconValue = (el) => {
...
+ *
+ * @cssproperty [--media-seek-backward-button-display = inline-flex] - `display` property of button.
+ */
 class MediaSeekBackwardButton extends MediaChromeButton {
   static get observedAttributes() {
-    return [...super.observedAttributes, MediaUIAttributes.MEDIA_CURRENT_TIME];
+    return [
+      ...super.observedAttributes,
+      MediaUIAttributes.MEDIA_CURRENT_TIME,
+      Attributes.SEEK_OFFSET,
```
gkatsev, line 49: whoops. Potentially another candidate to port to 0.x branch.

package.json
```diff
@@ -19,13 +19,14 @@
     "format": "prettier --write .",
     "clean": "rimraf dist",
     "lint": "eslint src/js",
+    "premanifest": "cp -R src/js/ dist/",
+    "manifest": "npx @custom-elements-manifest/analyzer analyze --config scripts/custom-elements-manifest.config.js",
```
gkatsev, line 23: we should be able to use `cem` or `custom-elements-manifest` as the bin name instead of going through npx here.
```suggestion
    "manifest": "cem analyze --config scripts/custom-elements-manifest.config.js",
```

gkatsev (approved):
LGTM.

I think the only thing I would've liked is more descriptions for the slots and attributes, but it's definitely not a blocker or anything.

package.json
luwes: this should be possible now since we don't run `yarn manifest` from the docs root. 
I'll follow up on this one.

PR #548, opened 2023-04-21
feat: Add missing attr getters and setters for media-time-display

AdamJaggard:
This is a first pass at defining getters and setters for all defined attributes. There's also JSDoc comments with the correct types.

I've used the attribute values themselves as the source of truth of the values which simplifies copying values back and forth between attributes and props.

Additionally, the public `mediaSeekableEnd` and `mediaSeekableStart` methods have been removed. These where undocumented and added to the public API of the component but where only used as utilities internally. We can add them back if we think they would be useful and document their usage.

## What should happen if you set a media UI attribute via one of the new setters?

If you set `remaining` or `showDuration` via a setter, the UI updates expectedly. If you set `mediaCurrentTime` the UI does update but the media doesn't actually seek. Clicking play for instance will reset the current time to it's previous value. Is that expected? or should an event be created and fired.

### DIFF
--- a/src/js/media-time-display.js
+++ b/src/js/media-time-display.js
@@ -234,6 +234,10 @@ class MediaTimeDisplay extends MediaTextDisplay {
   }
 
   set mediaSeekable(range) {
+    if (range === undefined) {
+      this.removeAttribute(MediaUIAttributes.MEDIA_SEEKABLE);
+      return;
+    }
     this.setAttribute(range.join(':'));
   }
 
--- a/src/js/media-time-display.js
+++ b/src/js/media-time-display.js
@@ -179,6 +179,8 @@ class MediaTimeDisplay extends MediaTextDisplay {
    * @param {boolean} show
    */
   set remaining(show) {
+    // don't set unless needed, could trigger an attr change event
+    if (show === this.remaining) return;
     this.toggleAttribute(Attributes.REMAINING, show);
   }
 
@@ -195,6 +197,8 @@ class MediaTimeDisplay extends MediaTextDisplay {
    * @param {boolean} show
    */
   set showDuration(show) {
+    // don't set unless needed, could trigger an attr change event
+    if (show === this.showDuration) return;
     this.toggleAttribute(Attributes.SHOW_DURATION, show);
   }
 
--- a/src/js/media-time-display.js
+++ b/src/js/media-time-display.js
@@ -76,6 +76,7 @@ const updateAriaValueText = (el) => {
  * @cssproperty --media-control-hover-background - `background` of control hover state.
  */
 class MediaTimeDisplay extends MediaTextDisplay {
+  /** @type {HTMLSlotElement} */
   #slot;
 
   static get observedAttributes() {
--- a/src/js/media-time-display.js
+++ b/src/js/media-time-display.js
@@ -215,15 +215,15 @@ class MediaTimeDisplay extends MediaTextDisplay {
    */
   get mediaCurrentTime() {
     const attrVal = this.getAttribute(MediaUIAttributes.MEDIA_CURRENT_TIME);
-    return attrVal != null ? +attrVal : undefined;
+    return attrVal ? +attrVal : undefined;
   }
 
   set mediaCurrentTime(time) {
-    if (time === undefined) {
+    if (time == null) {
       this.removeAttribute(MediaUIAttributes.MEDIA_CURRENT_TIME);
       return;
     }
-    this.setAttribute(MediaUIAttributes.MEDIA_CURRENT_TIME, time.toString(10));
+    this.setAttribute(MediaUIAttributes.MEDIA_CURRENT_TIME, `${+time}`);
   }
 
   /**
--- a/src/js/media-time-display.js
+++ b/src/js/media-time-display.js
@@ -163,21 +163,104 @@ class MediaTimeDisplay extends MediaTextDisplay {
     this.tabIndex = -1;
   }
 
+  // Own props
+
+  /**
+   * Whether to show the remaining time
+   * @returns {boolean}
+   */
+  get remaining() {
+    return this.hasAttribute(Attributes.REMAINING);
+  }
+
+  /**
+   * Whether to show the remaining time
+   * @param {boolean} show
+   */
+  set remaining(show) {
+    if (show && !this.hasAttribute(Attributes.REMAINING)) {
+      this.setAttribute(Attributes.REMAINING, '');
+    } else if (!show && this.hasAttribute(Attributes.REMAINING)) {
+      this.removeAttribute(Attributes.REMAINING);
+    }
+    this.update();
+  }
+
+  /**
+   * Whether to show the duration
+   * @returns {boolean}
+   */
+  get showDuration() {
+    return this.hasAttribute(Attributes.SHOW_DURATION);
+  }
+
+  /**
+   * Whether to show the duration
+   * @param {boolean} show
+   */
+  set showDuration(show) {
+    if (show && !this.hasAttribute(Attributes.SHOW_DURATION)) {
+      this.setAttribute(Attributes.SHOW_DURATION, '');
+    } else if (!show && this.hasAttribute(Attributes.SHOW_DURATION)) {
+      this.removeAttribute(Attributes.SHOW_DURATION);
+    }
+    this.update();
+  }
+
+  // Props derived from media UI attributes
+
+  /**
+   * Get the duration
+   * @returns {number | undefined}
+   */
   get mediaDuration() {
     const attrVal = this.getAttribute(MediaUIAttributes.MEDIA_DURATION);
     return attrVal != null ? +attrVal : undefined;
   }
 
+  /**
+   * Set the duration
+   * @param {number} time the new duration in seconds
+   */
+  set setMediaDuration(time) {
+    this.setAttribute(MediaUIAttributes.MEDIA_DURATION, time.toString(10));
+  }
+
+  /**
+   * The current time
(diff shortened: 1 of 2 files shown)


src/js/media-time-display.js
```diff
@@ -157,31 +149,104 @@ class MediaTimeDisplay extends MediaTextDisplay {
...
+  get showDuration() {
+    return this.hasAttribute(Attributes.SHOW_DURATION);
+  }
+
+  /**
+   * Whether to show the duration
+   * @param {boolean} show
+   */
+  set showDuration(show) {
+    if (show && !this.hasAttribute(Attributes.SHOW_DURATION)) {
```
cjpillsbury, line 188: suggestion(non-blocking): I think we probably want to do a simple check/early bail for these. Since `attributeChangedCallback()` will be invoked whenever an attribute is set, even if it hasn't actually changed values. Something like:
```js
if (show === this.showDuration) return;
```

src/js/media-time-display.js
```diff
@@ -157,31 +149,104 @@ class MediaTimeDisplay extends MediaTextDisplay {
...
+      this.setAttribute(Attributes.SHOW_DURATION, '');
+    } else if (!show && this.hasAttribute(Attributes.SHOW_DURATION)) {
+      this.removeAttribute(Attributes.SHOW_DURATION);
+    }
+    this.update();
```
cjpillsbury, line 193: nitpick(blocking): Remove all of the property-based invocations of `this.update()`. Since we set attributes in our setters, the method will already get invoked via `attributeChangedCallback()`, so redundant.

src/js/media-time-display.js
```diff
@@ -157,31 +149,104 @@ class MediaTimeDisplay extends MediaTextDisplay {
...
+   * Whether to show the remaining time
+   * @param {boolean} show
+   */
+  set remaining(show) {
+    if (show && !this.hasAttribute(Attributes.REMAINING)) {
+      this.setAttribute(Attributes.REMAINING, '');
+    } else if (!show && this.hasAttribute(Attributes.REMAINING)) {
+      this.removeAttribute(Attributes.REMAINING);
+    }
+    this.update();
```
luwes, line 172:
```suggestion
this.toggleAttribute(Attributes.REMAINING, Boolean(show));
```

I don't think it's really worth the extra code for the attribute exist check. what do you think?

`this.update();` should not be required I think because the attributeChangedCallback will be called and there it should have the `this.update();` call

src/js/media-time-display.js
luwes, line 193: same here

src/js/media-time-display.js
AdamJaggard: > don't think it's really worth the extra code for the attribute exist check. what do you think?

I was wondering the same thing! this is cleaner.

I think you're right about the update() it's going to run twice

cjpillsbury (changes requested):

src/js/media-time-display.js
```diff
@@ -157,31 +149,98 @@ class MediaTimeDisplay extends MediaTextDisplay {
...
   get mediaDuration() {
     const attrVal = this.getAttribute(MediaUIAttributes.MEDIA_DURATION);
     return attrVal != null ? +attrVal : undefined;
   }
 
+  /**
+   * Set the duration
+   * @param {number} time the new duration in seconds
+   */
+  set setMediaDuration(time) {
```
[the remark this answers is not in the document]
AdamJaggard: good catch

luwes (approved):
lgtm!

src/js/media-time-display.js
```diff
@@ -157,31 +149,85 @@ class MediaTimeDisplay extends MediaTextDisplay {
...
+   * Get the duration
+   * @type {number | undefined} In seconds
+   */
   get mediaDuration() {
     const attrVal = this.getAttribute(MediaUIAttributes.MEDIA_DURATION);
     return attrVal != null ? +attrVal : undefined;
   }
 
+  set mediaDuration(time) {
+    this.setAttribute(MediaUIAttributes.MEDIA_DURATION, time.toString(10));
```
luwes, line 194: what happens if time is undefined or a string value here?

it would be good I think if string numbers would still work and don't break the functionality

gkatsev: Good call. It seems strings should still work (I tried `'10'.toString(10)` and it gave me back the string `'10'`), but worth addressing `time` being undefined.

Probably removing the attribute when a falsey value is provided?

AdamJaggard: I'm thinking of accepting undefined for all the new prop setters instead of just the ones that explicitly have undefined as a return type on the getter (what I was doing currently), I think that makes sense?

src/js/media-time-display.js
```diff
@@ -187,11 +187,15 @@ class MediaTimeDisplay extends MediaTextDisplay {
    */
   get mediaDuration() {
     const attrVal = this.getAttribute(MediaUIAttributes.MEDIA_DURATION);
-    return attrVal != null ? +attrVal : undefined;
+    return attrVal ? +attrVal : undefined;
```
luwes, line 190: should be fine, `NaN` came to mind or `0` but the attribute is not set to those I believe. would be strange.

AdamJaggard: Good point, `0` is a valid response even if in this case it is unexpected. I'm planning to move this logic into a utility so that going forward this logic is easy to tweak and test. I think I will do that in this PR as a proof of concept.

src/js/utils/element-utils.js
```diff
@@ -59,11 +63,69 @@ export function getOrInsertCSSRule(styleParent, selectorText) {
...
+  // avoid setting a value that hasn't changed
+  if (getBooleanAttr(el, attrName) == show) return;
+
+  // also handles undefined
+  if (show == null) {
+    el.removeAttribute(attrName);
+    return;
+  }
+
+  el.toggleAttribute(attrName, show);
```
heff, line 130: I don't think this is right? And the logic is kind of confusing. The second arg of toggleAttribute is `force`, so I'm not clear how this is intended to work.

I'd recommend changing "show" to "value", and then if value is falsey removeAttribute, if truthy `setAttribute(attrName, attrName).

src/js/utils/element-utils.js
```diff
@@ -59,11 +63,69 @@ export function getOrInsertCSSRule(styleParent, selectorText) {
...
+ * @param {HTMLElement} el
+ * @param {string} attrName
+ * @param {number} value
+ */
+export function setNumericAttr(el, attrName, value) {
+  // avoid setting a value that hasn't changed
+  if (getNumericAttr(el, attrName) == value) return;
+
+  // also handles undefined
+  if (value == null) {
```
heff, line 96: Also need to handle `NaN` here somehow.

src/js/utils/element-utils.js
```diff
@@ -59,11 +63,69 @@ export function getOrInsertCSSRule(styleParent, selectorText) {
...
+
+/**
+ * Gets the number represented by the attribute
+ * @param {HTMLElement} el
+ * @param {string} attrName
+ * @returns {number | undefined} Will return undefined if no attribute set
+ */
+export function getNumericAttr(el, attrName) {
+  const attrVal = el.getAttribute(attrName);
+  return attrVal != null ? +attrVal : undefined;
```
heff, line 83: Undefined feels like the wrong value to return here. We're calling this a number type so I think `NaN` could/should be the falsey value, like the duration prop on the video el.

(I also think the `mediaDuration` prop where we currently return `undefined` is wrong)

src/js/utils/element-utils.js
gkatsev: should this call it `force` to match the name of `toggleAttribute`? I think `show` would be a fine name for the setters that use this helper, though.

Might be worth also adding a note in the jsdoc about what this means (or even linking to mdn or something)

src/js/utils/element-utils.js
AdamJaggard: As it currently is implemented `NaN` is handled implicitly as any other number. Setting `NaN` stores it in the DOM attribute as a string and using the getter retrieves it as a real `NaN` value.

Would you prefer it to be treated the same as `null` and `undefined`? that is: if someone sets the prop to `NaN` the attribute is removed from the element, and absence of the attribute on the getter side would return `NaN`? I can see the benefit of narrowing the prop type from `number | undefined` to simply `number` as `typeof NaN === 'number'`. As long as that makes sense from a consumers perspective i.e. the value can never have a state of "undefined" it is always a value, with the "default" value being NaN.

AdamJaggard: If I use this component in isolation, initialise it in a web page and access the prop `element.mediaDuration` and it returns `NaN` (because no value has been set), I think that is counter-intuitive from a developers perspective? I think I would expect a default like `0` (if it's important to maintain the purity of the return type) or `undefined` to signal that a value as never been set, so there is no value to return.

src/js/utils/element-utils.js
luwes: I'm torn on this one. I agree for sure it's easier to work with `undefined` and it would be handy to use this util as 

`getNumericAttr(...) ?? 0` or `getNumericAttr(...) ?? NaN`

which we do in Mux elements.

I'm leaning to also return `undefined` for the Element property because of the reflection of attr / prop.
it seems easier they would be more interchangeable for checking nullish  `el.mediaDuration == null` and `el.getAttribute('mediaduration') == null`

cjpillsbury (approved):
pre-approving pending build fail resolution

PR #549, opened 2023-04-21
Experiment/media store

cjpillsbury:


src/js/media-controller.js
```diff
@@ -292,6 +319,11 @@ class MediaController extends MediaContainer {
         stateDetails.get(this)
       );
     });
+    if (this._mediaStore) {
+      Object.entries(this._mediaStore.getState()).forEach(([stateName, stateValue]) => {
+        propagateMediaState([el], stateName.toLowerCase(), stateValue);
```
heff, line 324: Is this just a simplified implementation or a choice to move to pushing all state to a stateReceiver, not just what it cares about? Or does propagateMediaState just work that out under the hood before pushing to the receiver?

src/js/mediaStore.js
```diff
@@ -0,0 +1,108 @@
+import { MediaUIEvents } from './constants.js';
+
+// Very similar to "MediaUIStates" except:
+export const stateFacade = {
```
heff, line 4: What do you mean by "facade" here?

src/js/mediaStore.js
```diff
@@ -0,0 +1,108 @@
...
+  mediaPaused: {
+    // renaming to "stateOwners" to indicate it is generic
```
heff, line 6: This terminology threw me off at first but I think I get it. It's not the final state value at this point, it's essentially the state ingredients or sources. stateOwner might not be the most accurate description because of that relationship but I'll sit with it.

I can see how it can make sense to call fullscreenElement a stateOwner, and update that in the store whenever the reference changes in the same way the media element is handled, but what about other controller settings like default-stream-type? What would be the owner of that?

src/js/mediaStore.js
```diff
@@ -0,0 +1,108 @@
...
+      };
+      mediaEvents.forEach((eventType) => {
+        stateOwners.media.addEventListener(eventType, handler);
+      });
+    });
+  }
+
+  return {
+    // note these "actions" could just be the events if we wanted, especially if we normalize on "detail" for
+    // any payload-relevant values
```
heff, line 73: That's my S1. Avoid adding new concepts/naming unless necessary.

src/js/mediaStore.js
```diff
@@ -0,0 +1,108 @@
...
+          MediaUIEvents.MEDIA_PLAY_REQUEST,
+        ].includes(type)
+      ) {
+        // Most if not all state change requests can be modeled as UPDATE requests (in the generic "CRUD" sense)
+        // 1. Name of state to change (in this case, "paused")
+        // 2. Value you want the state to be (in this case, true if MEDIA_PAUSE_REQUEST, false if MEDIA_PLAY_REQUEST)
+        // We can think of more specific requests (e.g. MEDIA_PAUSE_REQUEST, MEDIA_PLAY_REQUEST) as just shorthand
+        // translations of e.g. { type: 'mediaupdaterequest', key: nameOfState, value: valueOfState }. In this case,
+        // 'mediapauserequest'|'mediaplayrequest' can be thought of as shorthand/convenience actions for something like:
+        // { type: 'mediaupdaterequest', detail: { key: 'mediaPaused', value: true | false } }
```
heff, line 87: I'm not seeing the value in this yet myself. While I do see complexity of then understanding that a `mediaplayrequest` gets translated into a paused update before then being translated back into a media.play function call. If we're gonna introduce those logic and mental jumps there's gotta be a really good reason for it.

src/js/mediaStore.js
```diff
@@ -0,0 +1,108 @@
...
+        stateFacade[key].set(value, stateOwners);
+      }
+      // if (type === 'mediaelementchangerequest') {} // we can also e.g. get updates for the stateOwners themselves
+    },
+    getState() {
+      // return the current state, whatever it is
+      return state;
+    },
+    subscribe(callback) {
+      // whenever state has (potentially?) changed based on "state  owners events", invoke callback(latestState)
```
heff, line 99: So every state receiver would get every state change with this, right? I understand this is a pattern, and I don't necessarily assume optimization makes a noticeable difference here, but this is less optimal then the current implementation where state receivers are more selective of what state they want to be updated with, right?

heff (commented):
Some important questions, but making sense so far. 👍

src/js/media-controller.js
```diff
@@ -19,6 +19,7 @@ import {
   AttributeToStateChangeEventMap,
 } from './constants.js';
 import { MediaUIRequestHandlers, MediaUIStates, volumeSupportPromise } from './controller.js';
+import createMediaStore from './mediaStore.js';
```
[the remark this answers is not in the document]
cjpillsbury: cool makes sense to me!

src/js/media-controller.js
cjpillsbury: The last one, but I think we could consider simplifying that as well (since it's all "under the hood", not a blocker for v1.0 imo)

src/js/mediaStore.js
cjpillsbury: Referencing the design pattern name here https://en.wikipedia.org/wiki/Facade_pattern, though happy to call it whatever you like ("Names Are Hard™️")

src/js/mediaStore.js
cjpillsbury: This is just a mental model that could also make the updates "generic" if we ever wanted to "pre-transform to the shape ☝️). That said, I'm not especially concerned about this one and we can do whatever folks find easier to work with/best when weighing the tradeoffs.

src/js/mediaStore.js
cjpillsbury: Yes-ish and definitely worth calling out/thinking through. Typically the updates are mitigated through a "selector" pattern that identifies the relevant state and effectively memoizes whether or not some effect will occur only if/when a "change" for the "interesting state" happens. I'm doing a toy example of this in the higher order callback used for subscribe in `<media-controller>`

src/js/mediaStore.js
cjpillsbury: an alternative approach would be to just treat these as part of the `state` itself. I can see reasons to treat them as special, but there are some benefits of allowing folks to consume the media element itself as state (say, for advanced use cases or "not yet modeled in state" use cases.

dylanjha (commented):
Is this a good summarization, for what a React developer would have to do in their application to accomplish the following:

1. Use Media Chrome for state management
2. Build the entire UI with my own custom controls

```jsx
import {
  MediaChromeProvider,
  useMediaUIDispatch,
  useMediaUISelector
} from 'media-chrome-react';

function App () {
  const dispatch = useMediaUIDispatch();
  const mediaPaused = useMediaUIDispatch((state) => state.mediaPaused);

  return (
    <MediaChromeProvider>
      <video
        slot="media"
        ref={(videoEl) =>
          dispatch({ type: 'mediaelementchangerequest', detail: videoEl })
        }
      />
      <button
        onClick={() => {
          const type = mediaPaused ? 'mediaplayrequest' : 'mediapauserequest';
          dispatch({ type });
        }}
      >{mediaPaused ? 'Play' : 'Pause'}</button>
    </MediaChromeProvider>
  );
}
```

cjpillsbury:
> Is this a good summarization, for what a React developer would have to do in their application to accomplish the following:
> 
> 1. Use Media Chrome for state management

Yup!

> 2. Build the entire UI with my own custom controls

For the scope of this POC, yes. We also have a clear story for official ("connected") React wrapper components that wrap our web components, but that was intentionally not part of the scope of this effort. If it would be helpful, I can add an example (probably just in the next.js app for now) for how that would work. The tl;dr - that case will not be an issue.

The only place we'd potentially need to work out is what you called out in our meeting w/ @heff where folks want to use the web components directly (sans React wrappers). I have some thoughts on how we could cleanly support that as well, but that is a stretch imo and I'd like to avoid including that as a pre-req for deciding to move forward.

> 
> ```js
> import {
>   MediaChromeProvider,
>   useMediaUIDispatch,
>   useMediaUISelector
> } from 'media-chrome-react';
> 
> function App () {
>   const dispatch = useMediaUIDispatch();
>   const mediaPaused = useMediaUIDispatch((state) => state.mediaPaused);
> 
>   return (
>     <MediaChromeProvider>
>       <video
>         slot="media"
>         ref={(videoEl) =>
>           dispatch({ type: 'mediaelementchangerequest', detail: videoEl })
>         }
>       />
>       <button
>         onClick={() => {
>           const type = mediaPaused ? 'mediaplayrequest' : 'mediapauserequest';
>           dispatch({ type });
>         }}
>       >{mediaPaused ? 'Play' : 'Pause'}</button>
>     </MediaChromeProvider>
>   );
> }
> ```

examples/nextjs-with-typescript/components/Video.tsx
```diff
@@ -0,0 +1,22 @@
...
+// @ts-ignore
+import { useMediaUIDispatch as useDispatch } from './media-chrome-react/hooks/useMediaUIDispatch.js';
+
+const Video = () => {
+  const dispatch = useDispatch();
+  return (
+    <video
+      ref={(videoEl) =>
+        dispatch({ type: 'mediaelementchangerequest', detail: videoEl })
+      }
```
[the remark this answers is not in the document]
cjpillsbury: > As far as I can tell, this is the only way for the React Provider to know about the media element. As such, would we expect all React developers to wire this up? Or could we simplify with maybe another hook?
> 
> ```js
> import { useMediaRef }
> 
> <video ref={useMediaRef} />
> ```

Yup! We could definitely do that and I think it makes sense since it's a "special" kind of event/action.

> 
> Also, how do we handle multiple players on the same page so that when events are dispatched they get sent to the correct media element?

As long as we don't have nested media ui state context providers, we are good here. We can also easily support more complex cases, but I'd strongly recommend not making that a hard requirement for v1.0.

examples/nextjs-with-typescript/pages/MediaChromeReactSimple.tsx
```diff
@@ -0,0 +1,92 @@
...
+      src="https://stream.mux.com/A3VXy02VoUinw01pwyomEO3bHnG4P32xzV7u1j1FSzjNg/high.mp4"
+      muted
+      crossOrigin=""
+      playsInline
+    ></video>
+  );
+};
+
+const MediaPlayButton = () => {
+  const dispatch = useDispatch();
```
cjpillsbury, line 46: Not sure if this would need to be solved in this iteration/v1.0 requirements, but we may want to (in some manner or another) account for "standalone usage" with the react wrapper components if we go in this direction. Food for thought.

examples/nextjs-with-typescript/components/media-chrome-react/stateMgmt.js
```diff
@@ -0,0 +1,74 @@
+import React, { createContext, useContext, useMemo } from 'react';
+// NOTE: This is an official package published by the React team and part of their monorepo: https://github.com/facebook/react/tree/main/packages/use-sync-external-store
+// We may be able to migrate to https://react.dev/reference/react/useSyncExternalStore instead if preferred
+import { useSyncExternalStoreWithSelector } from 'use-sync-external-store/with-selector.js';
+import createMediaStore from 'media-chrome/dist/mediaStore.js';
+
+// The context is the thing that allows for a shared... context that propagates outside of the (V)DOM/Component tree.
+// Contexts have values. Our value will be the the media ui store (aka the "guts" of Media Controller state mgmt)
+export const ReactMediaChromeContext = createContext(null);
```
heff, line 9: Could we simplify the naming to just MediaContext?

examples/nextjs-with-typescript/pages/MediaChromeReactSimple.tsx
```diff
@@ -0,0 +1,343 @@
...
+  // @ts-ignore
+  const mediaPaused = useSelector((state) => state.mediaPaused);
+  return (
+    // @ts-ignore
+    <media-play-button
+      mediapaused={mediaPaused ? '' : undefined}
+      // NOTE: The real version would actually forward on the mediaplayrequest/mediapauserequest events
```
heff, line 224: If you have to do all this wiring still, it doesn't seem like there's much value in using the web components like this. Right? Just the default styles maybe?

examples/nextjs-with-typescript/pages/MediaChromeReactSimple.tsx
```diff
@@ -0,0 +1,343 @@
...
+      />
+    </div>
+  );
+};
+
+const Video = ({ src }: { src?: string }) => {
+  const mediaRefCallback = useMediaRefCallback();
+  return (
+    <video
+      ref={mediaRefCallback}
```
heff, line 191: I'm interested to understand better how this ultimately connects to the store.

src/js/media-controller.js
```diff
@@ -62,57 +61,57 @@ class MediaController extends MediaContainer {
...
-        if (!this.media) {
-          console.warn('MediaController: No media available.');
-          return;
-        }
-
-        MediaUIRequestHandlers[key](this.media, e, this);
-      };
-      this.addEventListener(MediaUIEvents[key], this[handlerName]);
+    this._mediaStore = createMediaStore({
+      media: this.media,
```
heff, line 75: We should make sure to test what happens when there's no media, if we haven't yet.

src/js/media-controller.js
```diff
@@ -62,57 +61,57 @@ class MediaController extends MediaContainer {
...
+      // @ts-ignore
+      rootNode: this.getRootNode(),
+      fullscreenElement: this.fullscreenElement,
+      // NOTE: Names are hard. What should these be called (both the "bucket" currently called config and the props)
```
heff, line 79: I might just call this `options`

src/js/media-controller.js
```diff
@@ -62,57 +61,57 @@ class MediaController extends MediaContainer {
...
+      config: {
+        subtitles: this.hasAttribute('defaultsubtitles'),
```
heff, line 81: The name "subtitles" here is pretty vague. I think in general it'd be good to match these option names with the attr name, unless there's a good reason not to.

src/js/media-controller.js
```diff
@@ -62,57 +61,57 @@ class MediaController extends MediaContainer {
...
+        mediaStreamType: this.getAttribute('defaultstreamtype') ?? undefined,
```
heff, line 82: Same here, `mediaStreamType` is kind of misleading without `default` in it.

src/js/media-controller.js
```diff
@@ -62,57 +61,57 @@ class MediaController extends MediaContainer {
...
-
+    const simpleStateCapture = () => {
+      let prevState = {};
+      return (nextState) => {
+        Object.entries(nextState).forEach(([stateName, stateValue]) => {
+          if (prevState[stateName] === stateValue) return;
+          // For now, we can just toLowerCase() from (camelCase) propName to (lowercase/"smushedcase") attrname
+          const attrName = stateName.toLowerCase();
+          this.propagateMediaState(attrName, stateValue);
+          // TODO: I don't think we want these events to bubble? Video element states don't. (heff)
```
heff, line 103: If we're gonna version bump I think we should change this with it. User "request" events I can see bubbling, but bubbling component state changes like this doesn't feel standard.

src/js/media-controller.js
```diff
@@ -62,57 +61,57 @@ class MediaController extends MediaContainer {
...
+    Object.values(MediaUIEvents).forEach(eventName => {
+      this.addEventListener(eventName, e => {
+        // Pass events along to the media store.
+        // @ts-ignore
+        this._mediaStore.dispatch(e);
+      });
     });
-
+    const simpleStateCapture = () => {
+      let prevState = {};
```
heff, line 96: Mostly just interested, do we need this wrapper function (simpleStateCapture)? Could prevState just be in the constructor closure, followed by a handler function that uses it?

src/js/media-controller.js
```diff
@@ -143,14 +144,31 @@ class MediaController extends MediaContainer {
 
     } else if (
       attrName === Attributes.DEFAULT_SUBTITLES &&
-      newValue !== oldValue &&
-      newValue === ''
+      newValue !== oldValue
     ) {
-      toggleSubsCaps(this, true);
+      this._mediaStore.dispatch({
+        type: 'configchangerequest',
```
heff, line 150: I'm continuing to like "option" over "config", i.e. `optionchangerequest`.

src/react/components/Context.js
```diff
@@ -0,0 +1,8 @@
+import { createContext } from 'react';
+
+export const ReactMediaChromeContext = createContext(null);
+if (process.env.NODE_ENV !== 'production') {
+  ReactMediaChromeContext.displayName = 'ReactMediaChrome';
+}
+
+export default ReactMediaChromeContext;
```
heff, line 8: Call this `MediaContext` instead?

src/react/components/Provider.js
```diff
@@ -0,0 +1,21 @@
...
+
+  // In a real version of this, we may want to make the store overridable and rely on some memoization "under the hood"
+  return React.createElement(
+    Context.Provider,
+    { value: mediaUIStore },
+    children
+  );
+}
+
+export default Provider;
```
heff, line 21: Call this `MediaContextProvider` instead?

src/react/hooks/useMediaUIDispatch.js
```diff
@@ -0,0 +1,10 @@
+import { useContext } from 'react';
+import { ReactMediaChromeContext } from '../components/Context.js';
+
+export function createMediaUIDispatchHook(context = ReactMediaChromeContext) {
+  return function useDispatch() {
+    const store = useContext(context);
+    return store.dispatch;
+  };
+}
+export const useMediaUIDispatch = createMediaUIDispatchHook();
```
heff, line 10: I'm not familiar enough with dispatch/selector functions to have a strong opinion, but `MediaUI` feels odd here for naming. In "use **[blah]** Dispatch" what is blah meant to be describing?

examples/nextjs-with-typescript/components/media-chrome-react/stateMgmt.js
cjpillsbury: Yup!

examples/nextjs-with-typescript/pages/MediaChromeReactSimple.tsx
cjpillsbury: In this example usage, yeah, it's definitely all for the styling (including e.g. swapping out the SVGs based on the `mediapaused` attr). This could also be the/a react wrapper we provide from a lib. In our first iteration, it sounds like we should treat these considerations as "out of scope" though.

examples/nextjs-with-typescript/pages/MediaChromeReactSimple.tsx
cjpillsbury: The tl;dr - it's basically just a specialized version of the examples that use `dispatch()`. Happy to discuss further. Here are some relevant tidbits:
1. `ref` callback (React component feature) - https://react.dev/reference/react-dom/components/common#ref-callback
2. `useMediaRefCallback()` definition - https://github.com/muxinc/media-chrome/pull/549/files#diff-72149d92e0737edb63a3cc1ac7aa148e6cf8b9105cb2d07eb7bd8dc05728d8bfR55-R60 (creates a function that will cause a `dispatch()` of the `"mediaelementchangerequest"` "action")
3. What media store does when it receives a `"mediaelementchangerequest"` "action" - https://github.com/muxinc/media-chrome/pull/549/files#diff-a13fb0284c9e2463ff5bd20405c5a92ac40b6d35dea7320d5e34bb3085591bf0R1000-R1002 (figures out what the state should be based on the newly added/replaced media "state owner", including potentially tearing down things like event listeners)

src/js/media-controller.js
cjpillsbury: yeah it works 👌 that's effectively the initial state by default. but def good callout.

src/js/media-controller.js
cjpillsbury: awesome yeah after getting further in the impl I was actually thinking the same thing.

src/js/media-controller.js
cjpillsbury: yup we could even add it as a private prop if that's preferred.

src/js/media-controller.js
cjpillsbury: awesome I'm down

src/react/components/Provider.js
cjpillsbury: Since this is a default export, **_typically_** they are just exported as `Provider` and you can call it whatever you'd like. I believe that's the norm. Also, fwiw, all React `<Provider>`s are context providers (by definition). That said, happy to rename.

src/react/hooks/useMediaUIDispatch.js
cjpillsbury: this was primarily based on prior convos. Other, generic state mgmt libs such as Redux have a similarly generically named `useDispatch()`. We can certainly also export using the generic name, though it's plausible that folks would have to do some renaming in their module if they (not unlikely) are using both e.g. redux and our media store in the same component/module/etc. Also, I could see folks encountering `useDispatch()` in a codebase and just assuming it's e.g. the Redux version of `useDispatch()`. If we want to keep the "blah" around, this would be more us coming up with a convention that "feels right" given the domain scoping, e.g. `useMediaDispatch()` (which is what it was originally) or `useMediaUIDispatch()` (which is what it became after convos around making it more explicitly about UI media state).

src/js/mediaStore.js
```diff
@@ -0,0 +1,1241 @@
...
+/**
+ * @typedef {'on-demand'|'live'|'unknown'} StreamTypeValue
+ */
+
+/**
+ * @typedef {'unavailable'|'unsupported'} AvailabilityTypeValue
+ */
+
+/**
+ * @typedef {Partial<HTMLVideoElement> & Pick<HTMLMediaElement,'play'> & {
```
AdamJaggard, line 31: Does the usage of `Pick` here make `play` a required property of the combined type? wondering if we need something like `Partial<Pick<HTMLMediaElement,'play'>>` to make it optional.

AdamJaggard:
The store code is currently in one large file. I wondered if it made sense to extract into something like `/media-store/core.js`, `media-store/types.js`, `media-store/defaultFacade.js`, `media-store/utils.js`, etc. For ease of reviewing and separating the logically independent parts. Emphasises the idea of things like the store and the facade not being tightly coupled. Not super important, but wanted to mention before merging and potentially moving / renaming files afterwards.

src/js/mediaStore.js
```diff
@@ -0,0 +1,1241 @@
...
+ * }} StateFacade
+ */
+
+/**
+ * @typedef {{
+ *   [K in keyof StateFacade]: ReturnType<StateFacade[K]['get']>
+ * } & {
+ *   mediaPreviewTime: number;
+ *   mediaPreviewImage: string;
+ *   mediaPreviewCoords: [string,string,string,string]
```
AdamJaggard, line 149: These stood out. Just wondered what makes them distinct as to not be part of the facade but not be defined as a separate concept and instead defined as a core part of `MediaState`. They seem more like "UI things" than "media playback things" which I think makes sense. Just making sure I understand their distinctiveness.

cjpillsbury:
>

src/js/mediaStore.js
cjpillsbury: Within the team, including discussions with @heff, we've had some soft rules around minimal requirements for what we've historically called "a slotted media element" in Media Chrome (and what is here called a "media state owner"). The tacit place we've landed for now is something like:
1. for something to count as a media state owner, it must minimally have
    1. a `play()` method that must conform to the `HTMLMediaElement` public API
    2. a `paused` property that must conform to the `HTMLMediaElement` public API (I should add this to the type definition, so I'm glad you called this out.
    3. 1+ event that's used by our generic/default state facade to monitor for `paused` state changes (I should add the add/remove event handlers as required as well here). The most obvious candidates would be `play` and `pause` (See descriptions in https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement#events as well as the algorithms/steps defined in https://html.spec.whatwg.org/multipage/media.html#playing-the-media-resource
2. _**if**_ a media state owner has a given property or method that exists on the extended set of keys defined in this type def, it must conform to the `HTMLMediaElement` public API or the corresponding extended properties as defined in the `typedef` TypeScript `type`.

cjpillsbury:
> The store code is currently in one large file. I wondered if it made sense to extract into something like `/media-store/core.js`, `media-store/types.js`, `media-store/defaultFacade.js`, `media-store/utils.js`, etc. For ease of reviewing and separating the logically independent parts. Emphasises the idea of things like the store and the facade not being tightly coupled. Not super important, but wanted to mention before merging and potentially moving / renaming files afterwards.

Yup good callout. I'm down for whichever. Let's make sure to discuss with the larger team on preferences. I'd like to move the generic functions I've created to different modules regardless (since they're generic, maybe not specific to `media-store`).

cjpillsbury:
Closing this branch to avoid clutter/confusion. Forward progress (including comments/discussion/feedback) on #784

PR #554, opened 2023-04-26
fix: css variables consistency updates

gkatsev:
BREAKING CHANGE: Change `--media-live-indicator-color` to `--media-live-button-indicator-color`, `--media-time-buffered-color` to `--media-time-range-buffered-color`, `--media-background-position` to `--media-poster-image-background-position`, and `--media-background-size` to `--media-poster-image-background-size`.

### DIFF
--- a/examples/vanilla/components.html
+++ b/examples/vanilla/components.html
@@ -48,7 +48,7 @@
         --media-font: initial;
         --media-font-family: initial;
         --media-font-size: initial;
-        --media-font-weight: inital;
+        --media-font-weight: initial;
         --media-icon-color: oklch(90% 0.4 140);
         --media-live-button-icon-color: oklch(70% 0.4 140);
         --media-live-button-indicator-color: oklch(50% 0.4 140);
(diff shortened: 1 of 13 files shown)


examples/vanilla/components.html
```diff
@@ -0,0 +1,287 @@
+<!DOCTYPE html>
```
cjpillsbury, line 1: praise: 😍

luwes (approved):
LGTM 👍

src/js/media-text-display.js
```diff
@@ -56,7 +56,6 @@ template.innerHTML = /*html*/`
  *
  * @cssproperty --media-control-display - `display` property of control.
  * @cssproperty --media-control-background - `background` of control.
- * @cssproperty --media-control-hover-background - `background` of control hover state.
```
luwes, line 59: oh since you deleted here, it's needed in media-time-display

luwes (changes requested):

src/js/media-text-display.js
gkatsev: ah, good catch

gkatsev: Updated

cjpillsbury (approved):

PR #555, opened 2023-04-28
docs: add attributes short descriptions

luwes:


src/js/media-preview-thumbnail.js
```diff
@@ -82,7 +76,6 @@ class MediaPreviewThumbnail extends window.HTMLElement {
   attributeChangedCallback(attrName, oldValue, newValue) {
     if (
       [
-        Attributes.TIME,
```
luwes, line 85: removed `Attributes.TIME` because it didn't do anything

src/js/media-container.js
```diff
@@ -209,6 +209,14 @@ function getBreakpoints(breakpoints, rect) {
 /**
  * @extends {HTMLElement}
  *
+ * @attr {boolean} audio
+ * @attr {string} autohide
+ * @attr {string} breakpoints
+ * @attr {boolean} gesturesdisabled
+ * @attr {boolean} keyboardcontrol
+ * @attr {boolean} noautohide
+ * @attr {boolean} userinactive
```
luwes, line 218: I figured media-controller and media-container we'll probably do manually to add more explanation

src/js/media-captions-button.js
```diff
@@ -37,8 +37,8 @@ const updateAriaChecked = (el) => {
 };
 
 /**
- * @attr {string} mediasubtitleslist
- * @attr {boolean} mediasubtitlesshowing
+ * @attr {string} mediasubtitleslist - (read-only) A list of all subtitles.
+ * @attr {boolean} mediasubtitlesshowing - (read-only) A list of the showing subtitles.
```
gkatsev, line 41: suggestion(non-blocking): It may be worth being explicit that it's both subtitles and captions.
```suggestion
 * @attr {string} mediasubtitleslist - (read-only) A list of all subtitles and captions.
 * @attr {boolean} mediasubtitlesshowing - (read-only) A list of the showing subtitles and captions.
```

src/js/media-preview-thumbnail.js
gkatsev: Currently, it isn't possible to force the preview thumbnail to show a particular thumbnail. Maybe `Attributes.TIME` was supposed to do that? The comment at the top of the file seems to imply that.
I think such a feature would be useful, but this isn't a blocker or req for v1.

Could you update the comment at the top of the file (or just delete it)

gkatsev (approved):
Thanks for circling back to this!

PR #558, opened 2023-04-29
docs: component reference docs

luwes:
example: https://media-chrome-docs-git-fork-luwes-reference-docs-mux.vercel.app/en/components/media-airplay-button

todo: 
- [x] CSS parts


docs/src/pages/en/components/media-captions-selectmenu.md
```diff
@@ -84,19 +84,3 @@ Check out this example for usage, but please try and use better colors as this c
...
 <media-captions-selectmenu id="mc-sm-1"></media-captions-selectmenu>
 ```
-
-### Exposed Parts
-
-| Name | Description |
-|------|-------------|
-| `button` | The default [media-captions-button](./media-captions-button) that's in the shadow DOM |
-| `listbox` | The default listbox that's in the shadow DOM |
-| `listitem` | A part that targets each listitem of the listbox |
```
luwes, line 94: reminder to add CSS parts back in via JS docs!!!

cjpillsbury:
@luwes I think we can merge this, but we've been talking about trying to differentiate between the "media state" attributes/properties vs. the component-specific attributes. Looking at your code, I think we should be able to do this programmatically pretty easily by grouping/splitting based on the attribute name being a value in `MediaUIAttributes`. Thoughts?

gkatsev:
love that we're able to get those tables generated!

gkatsev (approved):

PR #560, opened 2023-05-02
feat: Drop all deprecated or redundant components. Update tests, exam…

cjpillsbury:
…ples, docs, etc. based on changes.


examples/vanilla/control-elements/media-container.html
cjpillsbury, line 1: NOTE: If we do decide to "split up" `<media-controller>` & `<media-container>`, we can always re-add a version of this (currently this isn't even exposed on the examples index.html).

test/unit/media-container.spec.js
```diff
@@ -1,51 +1,57 @@
 import { spy } from 'sinon';
 import { fixture, assert, aTimeout } from '@open-wc/testing';
-import '../../src/js/index.js';
+import MediaContainer from '../../src/js/media-container.js';
 
 describe('<media-container>', () => {
+  // NOTE: Although MediaContainer isn't designed for independent use, registering it
+  // here to test behavior independent of <media-controller>
+  before(async () => {
```
cjpillsbury, line 8: NOTE: this is actually closer to a "unit" test, since it's actually using the MediaContainer class def directly, whereas before it was still using `MediaController`.

src/js/experimental/media-title-element.js
cjpillsbury, line 1: NOTE: If we're not confident we want this, we should just remove it now. It's simple enough to re-add if we decide it's a worthwhile official + distinct component for Media Chrome

luwes (approved):
♻️ 👍

test/unit/media-container.spec.js
gkatsev: Are these tests useful? Should they just be moved to media-controller tests and media-container.spec is just removed?

Also, is it possible to undefine an element? Because, otherwise, this test makes lasting global changes.

cjpillsbury: > Are these tests useful? Should they just be moved to media-controller tests and media-container.spec is just removed?

I think yes if we're interested in testing responsibilities/separation of concerns, more "unit-level" tests, especially if we may move to some kind of split between controller and container usage. If we only want integration-level behavior-based testing, then arguably no, though I'd lean away from that as a hard rule.

cjpillsbury: > Also, is it possible to undefine an element? Because, otherwise, this test makes lasting global changes.

Since we discussed out of band, adding here for visibility: I did a quick test to confirm - at least with our current test runner setup, it looks like each test module is "self contained", so this shouldn't spill over into the global state of other modules. Simple test was adding `assert.notExists(window.customElements.get('media-container'));` to test cases across other test modules and confirming they still passed.

PR #561, opened 2023-05-02
docs: redirect [lang]/ to the [lang]/get-started

gkatsev:
For example, visiting https://media-chrome-docs-git-fork-gkatsev-redirect-lang-to-0b369f-mux.vercel.app/en redirects to https://media-chrome-docs-git-fork-gkatsev-redirect-lang-to-0b369f-mux.vercel.app/en/get-started

Uses meta direct because we don't currently have SSR enabled.

### DIFF
--- a/docs/src/pages/[lang]/index.astro
+++ b/docs/src/pages/[lang]/index.astro
@@ -1,0 +1,12 @@
+---
+import { KNOWN_LANGUAGE_CODES } from '../../config';
+export function getStaticPaths() {
+  return KNOWN_LANGUAGE_CODES.map(lang => ({params: { lang }}));
+}
+
+const { lang } = Astro.params;
+// get-started is the "default" page right now via convention
+const htmlString = `<meta http-equiv="refresh" content="0; URL=/${lang}/get-started" />`;
+---
+<Fragment set:html={htmlString} />
+


docs/src/pages/[lang]/index.astro
```diff
@@ -0,0 +1,12 @@
+---
+import { KNOWN_LANGUAGE_CODES } from '../../config';
+export function getStaticPaths() {
+  return KNOWN_LANGUAGE_CODES.map(lang => ({params: { lang }}));
+}
+
+const { lang } = Astro.params;
+// get-started is the "default" page right now via convention
+const htmlString = `<meta http-equiv="refresh" content="0; URL=/${lang}/get-started" />`;
```
Comment (gkatsev) [inline docs/src/pages/[lang]/index.astro:9]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#http-equiv

docs/src/pages/[lang]/index.astro
```diff
@@ -0,0 +1,12 @@
...
+---
+<Fragment set:html={htmlString} />
```
Comment (luwes) [inline docs/src/pages/[lang]/index.astro:11]: never seen this before...
can this be written without `<Fragment />`?

Comment (gkatsev) [reply docs/src/pages/[lang]/index.astro:11]: Got it from https://docs.astro.build/en/core-concepts/astro-syntax/#fragments--multiple-elements.

I tried it without fragment, but then `{lang}` is escaped because it's part of a string.

As mentioned on slack, I wonder if the docs should be at `media-chrome.org/docs` and have that be the re-direct, because `media-chrome.org/[lang]` could potentially mean a different language of the landing page.

Also, ideally, we'd use Astro.redirect but that's SSR only, which is a larger effort (would need the vercel plugin and what not, and at that point, we should consider upgrading to astro 2.0 as well).

Comment (luwes) [reply docs/src/pages/[lang]/index.astro:11]: this works under the front matter,

```jsx
<meta http-equiv="refresh" content={`0; URL=/${lang}/get-started`} />
```

Comment (gkatsev) [reply docs/src/pages/[lang]/index.astro:11]: cool, let's do that then. I'll close this PR and open a ticket, so we don't forget.

PR #562, opened 2023-05-02
WIP watch manifest from docs

gkatsev:


scripts/cem-plugin-css-vars.js
```diff
@@ -52,13 +52,15 @@ export function generateCssVars() {
...
+      console.log(manifest);
+
       // Add the css properties from the super class if they are not set yet.
       for (let cls of getClasses(manifest)) {
 
         if (cls.superclass?.name) {
           const SuperClass = getClasses(manifest).find(({ name }) => cls.superclass.name === name);
 
-          for (let cssProp of SuperClass.cssProperties ?? []) {
+          for (let cssProp of SuperClass?.cssProperties ?? []) {
```
gkatsev, line 63: without it, `cem analyze --watch` fails on a refresh but `manifest` itself is pretty empty, so, the output is busted.

PR #566, opened 2023-05-03
feat: Add props and types to seek forwards and backwards elements

AdamJaggard:
Additional props and types for existing attributes and some small bits of clean up. I moved the duplicated utils into a separate file to make them easier to maintain and test in the future.

### DIFF
--- a/src/js/media-seek-backward-button.js
+++ b/src/js/media-seek-backward-button.js
@@ -1,7 +1,7 @@
 import MediaChromeButton from './media-chrome-button.js';
 import { window, document } from './utils/server-safe-globals.js';
 import { MediaUIEvents, MediaUIAttributes } from './constants.js';
-import { updateAriaLabel, updateSeekIconValue } from './utils/seek';
+import { updateAriaLabel, updateSeekIconValue } from './utils/seek.js';
 
 export const Attributes = {
   SEEK_OFFSET: 'seekoffset',
@@ -44,7 +44,7 @@ class MediaSeekBackwardButton extends MediaChromeButton {
       this.setAttribute(Attributes.SEEK_OFFSET, DEFAULT_SEEK_OFFSET);
     }
     updateAriaLabel(this);
-    updateSeekIconValue(this);
+    updateSeekIconValue(this, 'backward');
     super.connectedCallback();
   }
 
@@ -53,7 +53,7 @@ class MediaSeekBackwardButton extends MediaChromeButton {
       if (newValue == undefined) {
         this.setAttribute(Attributes.SEEK_OFFSET, DEFAULT_SEEK_OFFSET);
       }
-      updateSeekIconValue(this);
+      updateSeekIconValue(this, 'backward');
       updateAriaLabel(this);
     }
 
--- a/src/js/media-seek-backward-button.js
+++ b/src/js/media-seek-backward-button.js
@@ -76,7 +76,7 @@ class MediaSeekBackwardButton extends MediaChromeButton {
       this.removeAttribute(Attributes.SEEK_OFFSET);
       return;
     }
-    this.setAttribute(Attributes.SEEK_OFFSET, `${value}`);
+    this.setAttribute(Attributes.SEEK_OFFSET, `${+value}`);
   }
 
   // Props derived from Media UI Attributes
--- a/src/js/media-seek-backward-button.js
+++ b/src/js/media-seek-backward-button.js
@@ -7,7 +7,7 @@ export const Attributes = {
   SEEK_OFFSET: 'seekoffset',
 };
 
-const DEFAULT_SEEK_OFFSET = '30';
+const DEFAULT_SEEK_OFFSET = 30;
 
 const backwardIcon = `<svg aria-hidden="true" viewBox="0 0 20 24"><defs><style>.text{font-size:8px;font-family:Arial-BoldMT, Arial;font-weight:700;}</style></defs><text class="text value" transform="translate(2.18 19.87)">${DEFAULT_SEEK_OFFSET}</text><path d="M10 6V3L4.37 7 10 10.94V8a5.54 5.54 0 0 1 1.9 10.48v2.12A7.5 7.5 0 0 0 10 6Z"/></svg>`;
 
@@ -40,9 +40,6 @@ class MediaSeekBackwardButton extends MediaChromeButton {
   }
 
   connectedCallback() {
-    if (!this.hasAttribute(Attributes.SEEK_OFFSET)) {
-      this.setAttribute(Attributes.SEEK_OFFSET, DEFAULT_SEEK_OFFSET);
-    }
     updateAriaLabel(this);
     updateSeekIconValue(this, 'backward');
     super.connectedCallback();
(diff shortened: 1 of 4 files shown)


src/js/media-seek-backward-button.js
```diff
@@ -69,13 +53,47 @@ class MediaSeekBackwardButton extends MediaChromeButton {
...
+   * @type {number} Seek amount in seconds
+   */
+  get seekOffset() {
+    return +this.getAttribute(Attributes.SEEK_OFFSET);
+  }
+
+  set seekOffset(value) {
+    // avoid triggeting a set if no change
+    if (value === this.seekOffset) return;
+    this.setAttribute(Attributes.SEEK_OFFSET, value.toString(10));
```
luwes, line 75: same here with null / undefined and string numbers

src/js/media-seek-backward-button.js
```diff
@@ -69,13 +53,47 @@ class MediaSeekBackwardButton extends MediaChromeButton {
...
+  /**
+   * @type {number} Seek amount in seconds
+   */
+  get seekOffset() {
+    return +this.getAttribute(Attributes.SEEK_OFFSET);
+  }
+
+  set seekOffset(value) {
+    // avoid triggeting a set if no change
+    if (value === this.seekOffset) return;
```
luwes, line 74: maybe use `==` to still return if `'43' == 43`

src/js/media-seek-backward-button.js
AdamJaggard: These are easy to account for but more a matter of how we would like the API to work e.g.

- Explicit, use the the type we define or expect an error
- Implicit, we coerce many value types into the type we want

I have a slight preference for the former but I'm not that passionate about it 😅. Is being more implicit in what we accept closer to how others would expect it to work? if the type is number is it expected that '50' works? or should the type be `number | string` to explicitly state that we support it.

cjpillsbury: We don't have a consistent stance on this, but we probably should. While I normally have your leanings, @AdamJaggard, here are my thoughts specifically for Media Chrome:

1. I think we can have fairly strict TypeScript definitions, such that, when folks **_are_** using it, the errors will inform them on what kind of values are expected. 
2. I also think we want to be a bit more permissive "under the hood", in the direction that @luwes is suggesting, in what is allowed in JS. This will keep our implementation closer to the way native `HTMLElement`s actually behave, and keeping Media Chrome consistent with native elements is a design principle.

This is effectively the relationship between the official TypeScript definitions of elements in lib.dom.d.ts vs. the actual implementation/behavior of the elements. By way of example, the definition of `HTMLMediaElement` in lib.dom.d.ts includes:

```ts
/** Gets or sets the volume level for audio portions of the media element. */
volume: number;
```
And here's how an `HTMLVideoElement` actually behaves in the browser JS runtime:
```js
videoEl.volume = '0.5';
videoEl.volume; // 0.5
videoEl.volume = true;
videoEl.volume; // 1
videoEl.volume = 'foo'; // caught TypeError: Failed to set the 'volume' property on 'HTMLMediaElement': The provided double value is non-finite.
```

Given that we also cast values _**back**_ to strings (as attributes), if we wanted the widest support, we'd need to make sure we cast to a number before applying the value to `setAttribute()`.

There are other reference points that could lead to slightly different conclusions, though I think those would be the wrong direction. For example, an `HTMLInputElement` of `type="range"` has its `min`/`max/`step` properties typed & casted as a `string`, which leads to much worse ergonomics imo.

This isn't saying we _**have**_ to do things this way, but I do think it best meets the various background considerations specifically applicable to Media Chrome. If there's lingering disagreement here, I think we should defer to Adam to allow him to continue making fast forward progress and take this up out of band (ideally soon & async, but we can wait until our office hours if needed).

src/js/media-seek-backward-button.js
```diff
@@ -55,12 +40,11 @@ class MediaSeekBackwardButton extends MediaChromeButton {
   }
 
   connectedCallback() {
-    // NOTE: currently don't support changing the seek value, so only need to set this once on initialization.
     if (!this.hasAttribute(Attributes.SEEK_OFFSET)) {
       this.setAttribute(Attributes.SEEK_OFFSET, DEFAULT_SEEK_OFFSET);
     }
     updateAriaLabel(this);
-    updateSeekIconValue(this);
+    updateSeekIconValue(this, 'backward');
```
cjpillsbury, line 47: thought(non-blocking): 👍 I'm cool with moving in this direction, but just to call it out - The prior implementation was angling towards eventual internationalization of all of our labeling/human-facing language. Since we don't know when we will get to that or what the underlying architecture is going to look like (at least in the details), I don't have any problems here.

src/js/media-seek-backward-button.js
gkatsev: Having stricter type defs while being more permissive in the js sounds good to me.

How did we want to handle falsey values? Remove the attribute? Or only do it for undefined/null? Interestingly, setting volume to `undefined` throws an error, but setting it to `null` sets it to 0.

src/js/media-seek-backward-button.js
```diff
@@ -1,17 +1,15 @@
...
-import { verbs } from './labels/labels.js';
-import { getSlotted, updateIconText } from './utils/element-utils.js';
+import { updateAriaLabel, updateSeekIconValue } from './utils/seek.js';
 
 export const Attributes = {
-  SEEK_OFFSET: 'seekoffset'
+  SEEK_OFFSET: 'seekoffset',
 };
 
 const DEFAULT_SEEK_OFFSET = '30';
```
cjpillsbury, line 10: question(non-blocking): With all the attention to types (TS and JS), should this be a number?

cjpillsbury (approved):

src/js/media-seek-backward-button.js
cjpillsbury: @gkatsev that's because:
```js
+null // 0
+undefined // NaN
```

src/js/media-seek-backward-button.js
gkatsev: I changed it to a string at some point because TSC complained because it expected `setAttribute` to be given a string and all the usages of this variable are in setAttribute.

src/js/media-seek-backward-button.js
AdamJaggard: Have updated these to handle null and undefined together. String numbers already work fine as string.toString() already works as expected.

src/js/media-seek-backward-button.js
AdamJaggard: For "correctness" sake I'd prefer it to be a number but I'm going to leave it as-is for simplicity and less casting everywhere.

src/js/media-seek-forward-button.js
```diff
@@ -69,13 +53,51 @@ class MediaSeekForwardButton extends MediaChromeButton {
...
+    const attrVal = this.getAttribute(MediaUIAttributes.MEDIA_CURRENT_TIME);
+    return attrVal != null ? +attrVal : undefined;
+  }
+
+  set mediaCurrentTime(time) {
+    if (time == null) {
+      this.removeAttribute(MediaUIAttributes.MEDIA_CURRENT_TIME);
+      return;
+    }
+    this.setAttribute(MediaUIAttributes.MEDIA_CURRENT_TIME, time.toString(10));
```
luwes:
```suggestion
    this.setAttribute(MediaUIAttributes.MEDIA_CURRENT_TIME, `${+time}`);
```

can we also use `${+time}` here to keep that consistent a bit ?

AdamJaggard: I pushed this change up right as you where commenting 😛

PR #568, opened 2023-05-03
docs: update time-range

luwes:
Test url: https://media-chrome-docs-git-fork-luwes-docs-time-range-mux.vercel.app/en/components/media-time-range


docs/src/pages/en/components/media-time-range.mdx
```diff
@@ -0,0 +1,165 @@
...
+  </media-control-bar>
+</media-controller>`}
+/>
+
+
+
+## Preview thumbnails
+
+Adding the `<track default label="thumbnails" kind="metadata" src="">` 
+with a valid VTT file as `src` will enable the preview thumbnails on hover functionality.
```
gkatsev, line 40:
```suggestion
Adding a metadata text track labelled "thumbnails" (`<track default label="thumbnails" kind="metadata" src="">`) 
with a valid VTT file as `src` will enable the preview thumbnails on hover functionality.
```

Also, did we want to describe the format here? In another guide? Preview Thumbnails component?

docs/src/pages/en/components/media-time-range.mdx
```diff
@@ -0,0 +1,165 @@
...
+  </media-control-bar>
+</media-controller>`}
+/>
+
+
+Discover all the available [CSS variables](#css-variables).
+
+### Using attributes
+
+The `<media-time-range>` will be updated with **media state** attributes any time
```
gkatsev, line 149: we should be consistent in what we're calling these and use those everywhere. The section is called media ui attributes, so, should we call it media ui attributes here instead of media state attributes?

luwes: good call, I'll change it to media UI attributes

gkatsev (approved):

PR #571, opened 2023-05-03
docs: update airplay and cast button docs

gkatsev:


luwes (approved):
LGTM.

2 nits:

- I would put the 2 top paragraphs together, it shows a big spacing now for 2 sentences.
- I'd make the first sandpack a bit higher so there's less confusion the bottom right buttons are not part of Media Chrome

SCR-20230504-dkd

gkatsev:
Ah, I didn't see it locally because it was a bit squished, so the player was smaller. I was trying to minimize the empty space after the code on the left. But I agree that it's better to have those buttons not overlap.

PR #573, opened 2023-05-04
docs: add generated styling reference

luwes:
`selectmenu` is left out as a group because it didn't generate CSS vars for now.

Test url: https://media-chrome-docs-git-fork-luwes-css-vars-reference-mux.vercel.app/en/reference/styling

### DIFF
--- a/docs/src/pages/en/styling-reference.astro
+++ b/docs/src/pages/en/styling-reference.astro
@@ -3,7 +3,11 @@ import { Markdown } from '@astropub/md';
 import MainLayout from '../../layouts/MainLayout.astro';
 import manifest from '../../../../dist/custom-elements.json';
 
-const { frontmatter = {}, headings = [], url } = Astro.props as Props;
+const frontmatter = {
+  title: 'Styling Reference'
+};
+
+const headings = [];
 const components = await Astro.glob('./components/*');
 
 function hasLink(tagName) {
@@ -51,14 +55,14 @@ function capitalizeFirstLetter(str) {
 }
 
 function backticksToMdnLinks(str) {
-  return str.replace(/`([^`]+)`/, '<a href="https://developer.mozilla.org/en-US/docs/Web/CSS/$1" target="_blank">$1</a>');
+  return str.replace(
+    /`([^`]+)`/g,
+    '<a href="https://developer.mozilla.org/en-US/docs/Web/CSS/$1" target="_blank"><code>$1</code></a>'
+  );
 }
 
 ---
-
 <MainLayout frontmatter={frontmatter} headings={headings}>
-  <h1>Styling Reference</h1>
-
   {groups.map(group => (
     <Markdown of={`## ${capitalizeFirstLetter(group)}s`} />
 
--- a/docs/src/pages/en/styling-reference.astro
+++ b/docs/src/pages/en/styling-reference.astro
@@ -15,12 +15,13 @@ const groups = [
   'range',
   'display',
   'indicator',
-  'other',
 ];
 
 for (let group of groups) {
   headings.push({ depth: 2, slug: `${group}s`, text: `${capitalizeFirstLetter(group)}s` });
 }
+
+headings.push({ depth: 2, slug: `others`, text: `Others` });
 
 let allCssProps = {};
 for (let cls of getClasses(manifest)) {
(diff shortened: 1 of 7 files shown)


cjpillsbury:
> `selectmenu` is left out as a group because it didn't generate CSS vars for now.

Awesome I think that makes sense because they're still experimental.

docs/src/pages/en/styling-reference.astro
```diff
@@ -0,0 +1,107 @@
...
+function hasLink(tagName) {
+  return components.some(c => c.file.includes(tagName));
+}
+
+const groups = [
+  'button',
+  'range',
+  'display',
+  'indicator',
+  'other',
```
cjpillsbury, line 18: I think this bucket will confuse folks, since not every CSS variable will apply to every "other" component. May want to treat this group differently with how you render, even if it's just a clear up-front caveat in text?

docs/src/pages/en/styling-reference.astro
```diff
@@ -0,0 +1,107 @@
...
+                return allCssProps[key].groups.has(group);
+              }
+              return groups.every(group => !allCssProps[key].groups.has(group));
+            })
+            .map((key) => {
+              const cssProp = allCssProps[key];
+              return <tr>
+                  <td class="nowrap"><code>{cssProp.name}</code></td>
+                  <td style="max-width: 20rem;">{cssProp.default && <code>{cssProp.default}</code>}</td>
+                  <td>{cssProp.description}</td>
```
cjpillsbury, line 99: suggestion(non-blocking): Since this is just rendering the text, the ` ticks for the code are showing up. Would be 😎 if we could account for that somehow (either render as `<code>` or maybe strip out if that's an easier first pass?)

docs/src/pages/en/styling-reference.astro
```diff
@@ -0,0 +1,107 @@
...
+
+    <ul>
+      {getClasses(manifest)
+        .filter(cls => {
+          if (group !== 'other') {
+            return cls.tagName?.endsWith(group);
+          }
+          return cls.tagName && groups.every(group => !cls.tagName?.endsWith(group));
+        })
+        .map(cls => (
```
cjpillsbury, line 68: suggestion(non-blocking): I think it would be nice to ensure that the "base class" components show up at the end (maybe with a simple description of "Base class" or similar?). Based on our naming convention, we can probably get away with matching against `name.startsWith('media-chrome-')`?

docs/src/pages/en/styling-reference.astro
gkatsev: might just need to be wrapped in `<Markdown>`?

Also, the backticks did a number on your comment 😛

docs/src/pages/en/styling-reference.astro
luwes: good call, will change this

docs/src/pages/en/styling-reference.astro
luwes: made these Mdn doc links that open in a new window

docs/src/pages/en/styling-reference.astro
```diff
@@ -0,0 +1,138 @@
...
+
+function backticksToMdnLinks(str) {
+  return str.replace(
+    /`([^`]+)`/g,
+    '<a href="https://developer.mozilla.org/en-US/docs/Web/CSS/$1" target="_blank"><code>$1</code></a>'
+  );
+}
+
+---
+<MainLayout frontmatter={frontmatter} headings={headings}>
```
dylanjha, line 65: Maybe can use a short intro paragraph at the top to give a little context about these variables?

cjpillsbury: > Maybe can use a short intro paragraph at the top to give a little context about these variables?

@dylanjha I think we should hold off on any "front matter" for this PR unless you're confident this should be a standalone page (and not e.g. appended to the styling guide).

dylanjha: It is it's own page, no?

https://media-chrome-docs-git-fork-luwes-css-vars-reference-mux.vercel.app/en/styling-reference

I think it makes sense as it's own page or was there another thought?

luwes: yes, @cjpillsbury is doing the styling guide and we're thinking to put that first and then below this reference.

but separate seems okay too, I'm fine with both

cjpillsbury: also fine with both! @dylanjha if you think it should be a separate page, then I'd agree with adding some "front matter" overview

dylanjha: I vote own page cause it's pretty long

luwes: updated this with a little description at the top. 

I added a "Reference" sidebar section that we could add to with other things. 

I kept it simple for now but maybe this could be a sidebar section that is closed by default and it can open up with a plus sign similar to how Mux docs does it or how Astro docs do it with a tabs UI thing.

https://media-chrome-docs-git-fork-luwes-css-vars-reference-mux.vercel.app/en/reference/styling

cjpillsbury:
@luwes looks like we're still showing `media-chrome-listbox`, `media-captions-listbox`, and `media-playback-rate-listbox`, which are still experimental.

cjpillsbury (approved):
👍

PR #579, opened 2023-05-08
docs: mute button and volume range docs

gkatsev:
Mute button preview: https://media-chrome-docs-git-fork-gkatsev-mute-button-mux.vercel.app/en/components/media-mute-button
Volume range preview: https://media-chrome-docs-git-fork-gkatsev-mute-button-mux.vercel.app/en/components/media-volume-range


cjpillsbury (approved):
suggestion(non-blocking): Remove the play button from the sandpack examples since it's unnecessary for the demo (we don't have play button for captions either).

gkatsev:
Yeah, I was thinking of the play button should be there to get a better sense of the volume as you're playing but there's probably no need for it as you can see the button update. I'll remove it to keep the example more minimal.

PR #582, opened 2023-05-09
fix: Let media-control-bar size (height) range components like it doe…

cjpillsbury:
…s with other components.


luwes (approved):

src/js/media-control-bar.js
```diff
@@ -20,10 +20,17 @@ template.innerHTML = /*html*/`
 
     media-time-range,
     ::slotted(media-time-range),
-    ::slotted(media-progress-range),
```
luwes, line 23: 🙈

PR #587, opened 2023-05-10
feat: Airplay, Captions, and Cast button props

AdamJaggard:
Adds missing props for the Airplay, Captions, and Cast buttons.

With captions specifically, we have to decide on the prop type for getting and setting a subtitles list. This is currently a simple array of strings. This matches how we store and retrieve the data from the attribute and also makes interacting with the prop similar to how a dev would interact with the attribute directly.

The alternative is to accept `TextTrack`'s (or objects that are a subset with only the data we need). Is this preferable? I can see this being simpler for the developer assuming they have `TextTrack`'s at hand.

### DIFF
--- a/src/js/media-airplay-button.js
+++ b/src/js/media-airplay-button.js
@@ -53,7 +53,7 @@ class MediaAirplayButton extends MediaChromeButton {
     // avoid triggering a set if no change
     if (this.mediaAirplayUnavailable === value) return;
 
-    if (value === undefined) {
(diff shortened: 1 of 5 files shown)


cjpillsbury (approved):
fine for now but it seems like we're leaning towards an array of track-like objects. Feel free to update in this PR or merge and resolve as a followup.

gkatsev (commented):
Small non-blocker

PR #590, opened 2023-05-10
docs: move media-controller to a component docs page

gkatsev:


docs/src/config.ts
```diff
@@ -71,6 +70,7 @@ export const SIDEBAR: Sidebar = {
       { text: 'Fullscreen button', link: 'en/components/media-fullscreen-button' },
       { text: 'Live indicator / button', link: 'en/components/media-live-button' },
       { text: 'Loading indicator', link: 'en/components/media-loading-indicator' },
+      { text: 'Media controller', link: 'en/components/media-controller' },
```
gkatsev, line 73: I thought it was still worth maintaining the alphabetical sorting.

cjpillsbury: agree if we're planning on covering some of the guide-like stuff separately. Eventually if/when we get collapsable and sub-sections, I think we'd want something more like categorical organizing (similar to @luwes's organization of the style reference and similar to e.g. https://mui.com/material-ui/react-autocomplete/ sections + sub-sections)

cjpillsbury (approved):
👌 looks like we're missing container parts and slots in the reference/jsdocs, but that can be a followup

PR #591, opened 2023-05-10
fix: don't auto-size ranges in media-control-bar if they're all alone.

cjpillsbury:
Fixes regression from previous update to auto-size (height) ranges in control bar like other components. This makes sure we don't do that if there are only media chrome ranges as descendants.


luwes (approved):
looks good, might need a follow up. try a CSS only solution somehow

cjpillsbury:
> looks good, might need a follow up. try a CSS only solution somehow

Agree. Would def prefer a CSS only solution. Should revisit if/when we get a chance to see if we can remove the explicit height from `media-chrome-range`.

PR #592, opened 2023-05-10
fix: use default values in new prop getters for seek btns. Use props …

cjpillsbury:
…instead of attrs for values under the hood.


AdamJaggard (approved):
looks good!

PR #595, opened 2023-05-11
feat: Duration, Fullscreen, and Gesture props

AdamJaggard:
Add missing props for Duration, Fullscreen, and Gesture components.


cjpillsbury (approved):
LGTM!

PR #597, opened 2023-05-11
feat: Pip button, Play button, and Playback Rates props

AdamJaggard:
Adds missing props.

Also fixes the playback rate click handler which seemed to not work on page load where `_rates` was undefined and would throw an error. Falls back to DEFAULT_RATES correctly now.

### DIFF
--- a/src/js/media-playback-rate-button.js
+++ b/src/js/media-playback-rate-button.js
@@ -18,26 +18,6 @@ slotTemplate.innerHTML = `
 `;
 
 /**
- * This will:
- * 1. parse the space-separated attribute string (standard for representing lists as HTML/CSS values) into an array (of strings)
- * The current regex allows for commas to be present between numbers to preserve legacy behavior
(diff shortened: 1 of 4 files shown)


src/js/media-playback-rate-button.js
```diff
@@ -70,12 +79,42 @@ class MediaPlaybackRateButton extends MediaChromeButton {
     super.attributeChangedCallback(attrName, oldValue, newValue);
   }
 
+  /**
+   * @type {Array<number>} The possible playback rates
+   */
+  get rates() {
```
cjpillsbury, line 85: issue(blocker): Needs to be refactored to use `AttributeTokenList` (type `DOMTokenList`)

cjpillsbury (changes requested):
refactor for rates, otherwise 👌. Feel free to split out `media-playback-rate-button` into a separate PR if you wanna get the other components merged sooner.

cjpillsbury (approved):
🎉

PR #598, opened 2023-05-11
feat: Poster Image, Preview Thumbnail, and Time Display props

AdamJaggard:
Adds missing props for Poster Image, Preview Thumbnail, and Time Display components.

### DIFF
--- a/src/js/media-preview-thumbnail.js
+++ b/src/js/media-preview-thumbnail.js
@@ -62,8 +62,8 @@ class MediaPreviewThumbnail extends window.HTMLElement {
       MediaStateReceiverAttributes.MEDIA_CONTROLLER
     );
     if (mediaControllerId) {
-      // @ts-ignore
       this.#mediaController =
+        // @ts-ignore
         this.getRootNode()?.getElementById(mediaControllerId);
       this.#mediaController?.associateElement?.(this);
     }
--- a/src/js/media-preview-thumbnail.js
+++ b/src/js/media-preview-thumbnail.js
@@ -3,7 +3,11 @@ import {
   MediaUIAttributes,
   MediaStateReceiverAttributes,
 } from './constants.js';
-import { getOrInsertCSSRule } from './utils/element-utils.js';
+import {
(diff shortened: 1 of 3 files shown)


src/js/media-preview-thumbnail.js
```diff
@@ -89,19 +97,45 @@ class MediaPreviewThumbnail extends window.HTMLElement {
...
+  get mediaPreviewImage() {
+    return getStringAttr(this, MediaUIAttributes.MEDIA_PREVIEW_IMAGE);
+  }
+
+  set mediaPreviewImage(value) {
+    setStringAttr(this, MediaUIAttributes.MEDIA_PREVIEW_IMAGE, value);
+  }
+
+  /**
+   * @type {Array<number> | undefined} Fixed length array [x, y, width, height] or undefined
```
cjpillsbury, line 112: nitpick(non-blocking): This should probably be `[number, number, number, number] | undefined` since it's a tuple

cjpillsbury (approved):
🏄‍♂️

src/js/media-preview-thumbnail.js
```diff
@@ -56,7 +63,8 @@ class MediaPreviewThumbnail extends window.HTMLElement {
     );
     if (mediaControllerId) {
       // @ts-ignore
-      this.#mediaController = this.getRootNode()?.getElementById(mediaControllerId);
+      this.#mediaController =
+        this.getRootNode()?.getElementById(mediaControllerId);
```
cjpillsbury, line 67: nitpick(blocking): whoops lost your @ts-ignore by the formatting! either move back to one line or move `//@ts-ignore`

src/js/media-preview-thumbnail.js
AdamJaggard: I initially did this and typescript didn't like the return type because I could be technically returning an array of any length. Although I guess I could add a type comment to the return to force it?

luwes (approved):
just needs that TS fix

PR #599, opened 2023-05-11
feat: Media Time Range and Volume Range props

AdamJaggard:
Adds missing props. Fixes attr documentation for mediapreviewtime in Time Range.


cjpillsbury (approved):
LGTM!

luwes (approved):

PR #604, opened 2023-05-11
fix: show chrome on pointermove for custom video

### LINKED ISSUE #451: <custom-video> show/hide controls edge case
There's a bug with the controls hover show/hide logic not working properly if a `<custom-video>` has set `pointer-events: none`
Repro here https://media-chrome.mux.dev/examples/media-elements/vimeo.html

There's no way to not set `pointer-events: none` for the Vimeo player (w/ no controls) because the Vimeo player has logic inside the iframe saying if the controls=0 param is set, set `pointer-events: none` on the body of the iframe basically hijacking all pointer events. 

It comes down to if pointer move events are fired on `<media-controller>`, should we also make active and schedule inactive.
removing this line 
https://github.com/muxinc/media-chrome/blob/4a3720efec9e3d90afffc2da3162fc0640dce117/src/js/media-container.js#L431-L431

edit: seems I last updated that line 
https://github.com/muxinc/media-chrome/commit/2c81cbc15800bf6986e3d0d6e67468bc42daf6e1

luwes:
fixes #451 related #298

### DIFF
--- a/src/js/media-container.js
+++ b/src/js/media-container.js
@@ -498,8 +498,6 @@ class MediaContainer extends window.HTMLElement {
       // so, only run pointermove for mouse
       if (e.pointerType !== 'mouse') return;
 
-      if (e.target === this) return;
-
       setActive();
       // Stay visible if hovered over control bar
       window.clearTimeout(this._inactiveTimeout);


cjpillsbury (approved):
I can't recall why we gated this in the first place, but I'm cool with this change as long as it's been thoroughly smoke tested.

PR #609, opened 2023-05-12
docs: darkmode for sandpack

gkatsev:


docs/src/components/Sandpack.jsx
```diff
@@ -18,6 +36,52 @@ export default function ComponentSandpack({
   active = Active.HTML,
   ...props
 }) {
+
+  const [theme, setTheme] = useState(light);
+  // const [theme, setTheme] = useState(() => {
+  //   if (import.meta.env.SSR) {
+  //     return themes.githubLight;
```
gkatsev, line 43: to make it work properly on refreshes we need this useState method that I borrowed from the ThemeToggleButton.

However, if on load, what we return *here* doesn't match the expected theme, the theme doesn't load correctly.
For example, if we're in dark-mode, but we return githubLight here, it doesn't load properly.

However, I'm not sure how to know what to return here. For the WIP, I have the `useState(light)` above, which always loads in the light mode at first. But sandpack will properly update the theme when we change the theme after loading.

gkatsev: Worked around it by using useEffect(() => {}, []) to run once to set the theme.

luwes (approved):
LGTM!

I'll have a go at getting the colors aligned with Github dark, shame that's not a theme yet

PR #618, opened 2023-05-12
fix: casting

luwes:


src/js/controller.js
```diff
@@ -445,8 +445,8 @@ export const MediaUIRequestHandlers = {
     if (streamType == StreamTypes.LIVE && autoSeekToLive) {
       MediaUIRequestHandlers['MEDIA_SEEK_TO_LIVE_REQUEST'](media);
     }
-
-    media.play().catch(() => {});
+    // Not all media's play() returns a Promise, e.g. castable-video
+    media.play()?.catch(() => {});
```
cjpillsbury, line 449: ah we should probably fix that but this is def the right solution for now. Nice find!

cjpillsbury (approved):
😎

PR #623, opened 2023-05-15
fix: Update official themes to use lowercase attrs consistently.

cjpillsbury:
- Fixes lingering `kebab-case` attrs in theme definitions.
- Renames theme-specific attributes as `lowercase` instead of `kebab0case` for consistency (e.g. `seek-forward-offset` is now `seekforwardoffset`)

### DIFF
--- a/src/js/themes/microvideo.js
+++ b/src/js/themes/microvideo.js
@@ -31,7 +31,7 @@ template.innerHTML = /*html*/`
     color: var(--_primary-color);
   }
 
-  [disabled] {
+  [disabled]:not(media-live-button) {
     opacity: 60%;
     cursor: not-allowed;
   }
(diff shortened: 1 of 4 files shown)


src/js/themes/microvideo.js
```diff
@@ -31,7 +31,7 @@ template.innerHTML = /*html*/`
     color: var(--_primary-color);
   }
 
-  [disabled] {
+  [disabled]:not(media-live-button) {
```
cjpillsbury, line 34: NOTE: `media-live-button` is `disabled` whenever not playing at the live edge and will automatically update its default cursor to `not-allowed`, so excluding it from this styling so it doesn't have partial opacity under these conditions.

src/js/themes/minimal.js
```diff
@@ -70,7 +70,7 @@ template.innerHTML = /*html*/`
       color: inherit;
     }
 
-    [disabled] {
+    [disabled]:not(media-live-button) {
```
cjpillsbury, line 73: NOTE: `media-live-button` is `disabled` whenever not playing at the live edge and will automatically update its default cursor to `not-allowed`, so excluding it from this styling so it doesn't have partial opacity under these conditions.

luwes (approved):
nice finds!

Issue #627, opened 2023-05-17
Bug: MediaFullscreenButton state is broken on window resize, can't exit fullscreen


mmvsk:
There is an issue with the `media-fullscreen-button`: when the player goes fullscreen, then the window is resized (past a media query breakpoint), the `media-fullscreen-button` is re-created, without the proper `media-is-fullscreen` attribute. Therefore the button becomes an "enter fullscreen" button, and pressing it has not effect.

This bug cencerns all browsers on Android (because in iOS the native controls are used), and it happens when you switch to fullscreen, then change the device orientation: it becomes impossible to exit the fullscreen mode (except with the back button, that is now automatically hidden). It's easy to reproduce on desktop using Chrome DevTools, device mode, entering fullscreen and resizing the page through media query breakpoints (300px, 700px, ...). Tip: use `f` to exit fullscreen ;-)

To test the button is re-recreated:

```typescript
// before orientation change: the same button is preserved between/after fullscreen change
const buttonBeforeOrientationChange = player.shadowRoot?.querySelector("media-theme")?.shadowRoot?.querySelector("media-controller media-control-bar media-fullscreen-button");

// --- RESIZE THE WINDOW

// this button is created after the orientation change, and is missing the `media-is-fullscreen` attribute
const newButtonAfterOrientationChange = player.shadowRoot?.querySelector("media-theme")?.shadowRoot?.querySelector("media-controller media-control-bar media-fullscreen-button");

// --- THE FIRST AND SECOND ONE ARE NOT THE SAME
```

The first button still has the attribute, but is not visible, and the second one is visible but has no attribute.

This issue is related to https://github.com/muxinc/elements/issues/692.

PR #628, opened 2023-05-17
fix: MediaFullscreenButton state is broken on window resize (#627)

mmvsk:
Fix issue #627 by checking in the constructor of `media-fullscreen-button` if its parent `media-controller` is fullscreen, and if it is, setting the `media-is-fullscreen` attribute of the button itself.

This is not an ideal solution, as a proper one should either prevent the re-creation of the button when the window is resized, or relying only on the `media-is-fullscreen` attribute of the parent `media-controller` and not of the button itself.

However, as the bug makes it hard to exit fullscreen on Android, this fix is better than nothing.

### DIFF
--- a/src/js/media-fullscreen-button.js
+++ b/src/js/media-fullscreen-button.js
@@ -57,6 +57,32 @@ const updateAriaLabel = (el) => {
   el.setAttribute('aria-label', label);
 };
 
+/** @param {HTMLElement} el */
+const isMediaControllerFullscreen = (el) => {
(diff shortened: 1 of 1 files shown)


luwes:
thanks for contributing! 

I did some more investigating and the root cause was that the media state was incorrect when there was no fullscreen event.target available. The new fullscreen button element did correctly receive state when registering but the state itself was incorrect. the fix was getting the state from something like document.fullscreenElement.

closing your PR in favor of https://github.com/muxinc/media-chrome/pull/634

PR #629, opened 2023-05-17
feat: add icon slot to mute button

AdamJaggard:
Adds a new combined icon slot. Useful for icon animations where a single SVG is preferred over switching between multiple SVG's. Adds some docs but this could use more examples in the future.

Also modifies CSS selectors to hide the slot elements themselves instead of the content inside the slots. This is simpler and I think achieves the same goal?


docs/src/pages/en/components/media-mute-button.mdx
```diff
@@ -61,6 +62,18 @@ Here's an example of how you can replace the volume levels with the words that c
 </media-controller>`}
 />
 
+Alternatively, if you would like to represent all of the volume levels using a single element you could use the `icon` slot instead.
```
cjpillsbury, line 65: suggestion: Let's add a fully working sandpack example for this with animation.

cjpillsbury:
suggestion: Let's add an example to our `examples` called something like `animated-icons` and uses `advanced` or similar as a baseline where we can smoke test and incrementally add animated icons as they're implemented per button.

examples/vanilla/animated-icons.html
```diff
@@ -0,0 +1,111 @@
...
+                  <path fill-rule="evenodd" clip-rule="evenodd" d="M15.2677 9.30323C15.463 9.49849 15.7796 9.49849 15.9749 9.30323C16.1701 9.10796 16.1701 8.79138 15.9749 8.59612L14.2071 6.82841L15.9749 5.06066C16.1702 4.8654 16.1702 4.54882 15.9749 4.35355C15.7796 4.15829 15.4631 4.15829 15.2678 4.35355L13.5 6.1213L11.7322 4.35348C11.537 4.15822 11.2204 4.15822 11.0251 4.35348C10.8298 4.54874 10.8298 4.86532 11.0251 5.06058L12.7929 6.82841L11.0251 8.59619C10.8299 8.79146 10.8299 9.10804 11.0251 9.3033C11.2204 9.49856 11.537 9.49856 11.7323 9.3033L13.5 7.53552L15.2677 9.30323Z" fill="white"/>
+                  </svg>
```
cjpillsbury, line 91: I think your structure got a bit mangled here. Also, you should be able to remove the explicit fill color here (See https://v1.media-chrome.org/en/styling#using-custom-icons and subsequent section)

docs/src/pages/en/components/media-mute-button.mdx
AdamJaggard: have pushed this up

examples/vanilla/animated-icons.html
AdamJaggard: oh yeah I missed these two fills

src/js/media-mute-button.js
```diff
@@ -20,33 +22,30 @@ const slotTemplate = document.createElement('template');
...
-  :host([${MediaUIAttributes.MEDIA_VOLUME_LEVEL}=high]) slot:not([name=high]) > *, 
-  :host([${MediaUIAttributes.MEDIA_VOLUME_LEVEL}=high]) ::slotted(:not([slot=high])) {
+  :host(:not([${MEDIA_VOLUME_LEVEL}])) slot:not([name=high]):not([name=icon]), 
+  :host([${MEDIA_VOLUME_LEVEL}=high]) slot:not([name=high]):not([name=icon]) {
     display: none !important;
   }
 
-  :host([${MediaUIAttributes.MEDIA_VOLUME_LEVEL}=off]) slot:not([name=off]) > *, 
-  :host([${MediaUIAttributes.MEDIA_VOLUME_LEVEL}=off]) ::slotted(:not([slot=off])) {
+  :host([${MEDIA_VOLUME_LEVEL}=off]) slot:not([name=off]):not([name=icon]) {
```
cjpillsbury, line 30: praise: These are so much cleaner!

cjpillsbury (approved):
Approved with a couple of (non-blocking) suggestions. This is gonna be 😎!

PR #632, opened 2023-05-19
chore: unminified dist + simplify local serve

luwes:
- removes all minified bundles from the dist folder 
- simplifies the needed code for local serving the examples by using the NPM package https://github.com/vercel/serve


package.json
```diff
@@ -24,17 +24,17 @@
     "build:types": "tsc",
     "build:esm": "esbuild src/js/*.js src/js/*/*.js src/js/*/*/*.js --target=es2019 --format=esm --outdir=dist",
     "build:cjs": "esbuild src/js/*.js src/js/*/*.js src/js/*/*/*.js --target=es2019 --format=cjs --outdir=dist/cjs",
-    "build:iife": "esbuild src/js/index.js --bundle --target=es2019 --format=iife --outdir=dist/iife --minify --sourcemap --global-name=MediaChrome && yarn size",
```
gkatsev, line 27: I think the iife file makes sense to get auto-minified.

package.json
```diff
@@ -24,17 +24,17 @@
...
+    "build:iife": "esbuild src/js/index.js --bundle --target=es2019 --format=iife --outdir=dist/iife --global-name=MediaChrome",
     "build:react": "node ./scripts/react/build.js",
     "build": "run-s build:types build:esm build:cjs build:iife build:react",
     "watch": "run-p watch:types watch:esbuild",
     "watch:types": "yarn build:types -w",
-    "watch:esbuild": "node ./scripts/esbuild.js",
-    "dev": "yarn watch",
+    "watch:esbuild": "yarn build:esm --watch=forever",
+    "serve": "npx serve . --config vercel.json --cors",
```
gkatsev, line 33: can serve be added as a dev dependency? Then we won't even need to use `npx` here.

PR #633, opened 2023-05-19
fix: iOS fullscreen state bug

### LINKED ISSUE #593: How to tell if an iOS device is playing media chrome in Full Screen
Hi there 👋 

I noticed that the `media-is-fullscreen` attribute does not get applied to the `<media-controller>` element on iOS. It does on Desktop.

Is there a way we can programatically know whether or not the video is in fullscreen mode (that works on iOS also)?

At the moment, we read "media chrome state" from the attributes on the `<media-controller>`. The fullscreen value is incorrect on iOS, so therefore our state is incorrect.

Is there another way to read fullscreen state?

I even tried adding an event listener manually (just for iOS) but it doesn't fire when you toggle full screen 😢 

```ts
document.addEventListener("webkitfullscreenchange", function() {
  ...
});
```

### iOS - no media-is-fullscreen attribute

_*Tested using Browserstack_

Screenshot 2023-05-10 at 1 45 39 pm

### Desktop - media-is-fullscreen attribute is present

Screenshot 2023-05-10 at 1 46 15 pm

Thanks in advance

luwes:
fixes #593

### DIFF
--- a/src/js/controller.js
+++ b/src/js/controller.js
@@ -300,14 +300,11 @@ export const MediaUIStates = {
   },
   MEDIA_IS_FULLSCREEN: {
     get: function (controller, e) {
-      // iOS has a specialized fullscreen API on the video element.
+      // iOS has a specialized fullscreen API on the video element, so fall back to that just in case.
       // https://developer.apple.com/documentation/webkitjs/htmlvideoelement/1630493-webkitdisplayingfullscreen
-      const isFullscreenOnIOS = e?.target?.webkitDisplayingFullscreen;
-      if (isFullscreenOnIOS) return true;
-
+      const isSomeElementFullscreen = !!(document[fullscreenApi.element] ?? e?.target?. webkitDisplayingFullscreen);
       // Safari doesn't support ShadowRoot.fullscreenElement and document.fullscreenElement
       // could be several ancestors up the tree. Use event.target instead.
-      const isSomeElementFullscreen = !!document[fullscreenApi.element];
       const fullscreenEl = isSomeElementFullscreen && e?.target;
       return containsComposedNode(controller.fullscreenElement, fullscreenEl);
     },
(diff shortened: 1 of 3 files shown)


src/js/controller.js
```diff
@@ -300,14 +300,20 @@ export const MediaUIStates = {
...
+      // iOS has a specialized fullscreen API on the video element.
+      // https://developer.apple.com/documentation/webkitjs/htmlvideoelement/1630493-webkitdisplayingfullscreen
+      const isFullscreenOnIOS = e?.target?.webkitDisplayingFullscreen;
+      if (isFullscreenOnIOS) return true;
+
       // Safari doesn't support ShadowRoot.fullscreenElement and document.fullscreenElement
       // could be several ancestors up the tree. Use event.target instead.
       const isSomeElementFullscreen = !!document[fullscreenApi.element];
       const fullscreenEl = isSomeElementFullscreen && e?.target;
       return containsComposedNode(controller.fullscreenElement, fullscreenEl);
```
cjpillsbury, line 312: issue(blocking): we still want to confirm `containsComposedNode()` on `e.target` here (e.g. multiple videos on a page).  We should be able to normalize this to the same logic, e.g.

```suggestion
      // iOS has a specialized fullscreen API on the video element, so fall back to that just in case.
      // https://developer.apple.com/documentation/webkitjs/htmlvideoelement/1630493-webkitdisplayingfullscreen
      const isSomeElementFullscreen = !!(document[fullscreenApi.element] ?? e?.target?. webkitDisplayingFullscreen);
      // Safari doesn't support ShadowRoot.fullscreenElement and document.fullscreenElement
      // could be several ancestors up the tree. Use event.target instead.
      const fullscreenEl = isSomeElementFullscreen && e?.target;
      return containsComposedNode(controller.fullscreenElement, fullscreenEl);
```

cjpillsbury (approved):
👌

PR #636, opened 2023-05-19
Docs/0.x 1.x migration guide

cjpillsbury:
Demo docs page: https://media-chrome-docs-git-fork-cjpillsbury-docs-0x-1x-mi-d37eb4-mux.vercel.app/en/migration/from-0.x-to-1.0

### DIFF
--- a/docs/src/config.ts
+++ b/docs/src/config.ts
@@ -57,9 +57,6 @@ export const SIDEBAR: Sidebar = {
       { text: 'Keyboard shortcuts', link: 'en/keyboard-shortcuts' },
       { text: 'Audio player', link: 'en/audio-player' },
     ],
-    Reference: [
-      { text: 'Styling', link: 'en/reference/styling' },
-    ],
     'Media Elements': [
       { text: 'Cloudflare Video', link: 'en/media-elements/cloudflare-video' },
       { text: 'DASH Video', link: 'en/media-elements/dash-video' },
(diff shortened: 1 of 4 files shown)


docs/src/config.ts
```diff
@@ -101,14 +98,16 @@ export const SIDEBAR: Sidebar = {
...
     ],
     Advanced: [
       { text: 'Stream Type', link: 'en/stream-type' },
       { text: 'Design Principles', link: 'en/design-principles' },
       { text: 'Architecture', link: 'en/architecture' },
     ],
-    "Learn More": [
+    "Resources": [
+      { text: 'Styling Reference', link: 'en/reference/styling' },
+      { text: 'Migrating to 1.0', link: 'en/migration/from-0.x-to-1.0' },
```
luwes, line 110: nice!

luwes (approved):
looking good!

docs/src/pages/en/migration/from-0.x-to-1.0.md
```diff
@@ -0,0 +1,334 @@
...
+
+```html
+<media-controller>
+  <!-- ... other components -->
+</media-controller>
+```
+
+## Working with themes
+
+All of the prior callotus about component attribute names, CSS variables, and the like, will need to be updated on any corresponding themes usage. In addition, just like our other components, all of our "official" themes that come with Media Chrome have been updated to have `lowercase` attribute names instead of `kebab-case`. For example,
```
luwes, line 284: 😄

docs/src/pages/en/migration/from-0.x-to-1.0.md
```diff
@@ -0,0 +1,344 @@
...
+### Breaking changes
+
+**General**
+- All attributes have been changed from `kebab-case` to `lowercase` (aka "`smushedcase`")
+- Several CSS variables have been renamed for consistency, more consistently applied, and more consistently defaulted
+- `AttributeTokenList` (e.g. `hotkeys`) removed the `keys()` method
+
+**Captions & Subtitles**
+- Captions & Subtitles list attributes are now just subtitles attributes
+- Enabling captions/subtitles by default belongs to the `<media-controller>` instead of the `<media-captions-button>`
```
gkatsev, line 18: would be nice to show the new attribute here so that they don't need to look it up separately.

gkatsev: ah, I guess it exists below. Would be nice to have these link there.

gkatsev (commented):
i guess my only suggestion is that it would be good if the overview section links down below because as I was reading it I was thinking that I'd need to look up elsewhere to get more details.

PR #639, opened 2023-05-22
fix: Make sure we monitor slotchange on slots for media state receive…

cjpillsbury:
…rs (for e.g. themes).

### DIFF
--- a/src/js/media-controller.js
+++ b/src/js/media-controller.js
@@ -664,24 +664,19 @@ const monitorForMediaStateReceivers = (
   };
 
   // Storing refs & handlers here so we can remove on unsubscribe
-  const slotAndChangeHandlers = [];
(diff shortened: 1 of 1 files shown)


src/js/media-controller.js
```diff
@@ -673,6 +693,8 @@ const monitorForMediaStateReceivers = (
   const unsubscribe = () => {
     // Unregister any Media State Receiver descendants (including ourselves)
     traverseForMediaStateReceivers(rootNode, unregisterMediaStateReceiver);
+    // Make sure we remove all slotchange event listeners
+    slotAndChangeHandlers.forEach(([slotEl, eventHandler]) => slotEl.removeEventListener(eventHandler));
```
AdamJaggard, line 697: I think `removeEventListener` is missing the event name as a first argument here

PR #642, opened 2023-05-23
chore: switch to .mjs and .cjs file extensions

gkatsev:
BREAKING CHANGE: esm dist files now have the mjs file extension and commonjs has the cjs file extension.


src/js/index.mjs
```diff
@@ -0,0 +1,54 @@
...
+import MediaPipButton from './media-pip-button.mjs';
+import MediaPlayButton from './media-play-button.mjs';
+import MediaPlaybackRateButton from './media-playback-rate-button.mjs';
+import MediaPosterImage from './media-poster-image.mjs';
+import MediaSeekBackwardButton from './media-seek-backward-button.mjs';
+import MediaPreviewTimeDisplay from './media-preview-time-display.mjs';
+import MediaPreviewThumbnail from './media-preview-thumbnail.mjs';
+import MediaTimeRange from './media-time-range.mjs';
+import MediaLoadingIndicator from './media-loading-indicator.mjs';
+import MediaVolumeRange from './media-volume-range.mjs';
```
luwes, line 27: the source files should really keep the `.js` extension. 
we can compile to anything but this is unneeded IMO.

luwes (dismissed):

docs/src/components/Sandpack.jsx
```diff
@@ -112,8 +112,8 @@ export default function ComponentSandpack({
     './styles.css',
     ...Object.keys(dependencies),
     ...Object.keys(files).reduce((importPaths, fileAbsPath) => {
-      // Only automatically import .css or .js files for now
-      if (fileAbsPath.endsWith('.css') | fileAbsPath.endsWith('.js')) {
+      // Only automatically import .css, .js, or .mjs files for now
+      if (fileAbsPath.endsWith('.css') | fileAbsPath.endsWith('.js') | fileAbsPath.endsWith('.mjs')) {
```
luwes, line 116: discussed internally, if possible we should avoid renaming all ESM dist files to have the `.mjs` extension to not have breaking changes. using named exports internally is a good workaround.


https://github.com/muxinc/media-chrome/compare/main...luwes:dist?expand=1
https://esbuild.github.io/content-types/#default-interop

it’s recommend for library authors by esbuild

> If you are a library author: When writing new code, you should strongly consider avoiding the default export entirely. It has unfortunately been tainted with compatibility problems and using it will likely cause problems for your users at some point.

gkatsev:
Replaced by https://github.com/muxinc/media-chrome/pull/650

PR #643, opened 2023-05-24
feat: Icon slots for buttons with multiple slots

AdamJaggard:
Adds an optional `icon` slot to buttons with multiple available slots e.g. Captions, Cast, Fullscreen, PiP and Play.

Updates docs with basic examples of usage. There's also a new `animated-icons.html` in the examples folder but this also currently has very basic examples to allow us to check the slots work as expected. We can expand this in the future with fancier animated demos.


luwes (approved):
Looks good! 

nit: in the docs the alternative way of using 1 icon could possibly have a subtitle

PR #651, opened 2023-05-26
chore: remove size script

luwes:
makes it easier to watch the iife build w/ `yarn build:iife --watch`


package.json
```diff
@@ -48,7 +48,7 @@
     "build:esm": "esbuild src/js/*.js src/js/*/*.js src/js/*/*/*.js --target=es2019 --format=esm --outdir=dist",
     "build:cjs": "esbuild src/js/*.js src/js/*/*.js src/js/*/*/*.js --target=es2019 --format=cjs --outdir=dist/cjs",
     "postbuild:cjs": "echo '{\"type\": \"commonjs\"}' > ./dist/cjs/package.json",
-    "build:iife": "esbuild src/js/index.js --bundle --target=es2019 --format=iife --outdir=dist/iife --minify --sourcemap --global-name=MediaChrome && yarn size",
```
gkatsev, line 51: do we want to remove it altogether? It could be moved to a `postbuild:iife` which will allow for `yarn build:iife --watch`

gkatsev (approved):

package.json
luwes: oh, yes, if that works sounds good

PR #653, opened 2023-05-26
docs: move docs to docs/, set up url redirects

gkatsev:
this will enable `media-chrome.org/docs` to redirect to docs and will redirect en/ to en/get-started and everything to docs/en/[...path] so existing links will continue working.


luwes (approved):
👍

luwes (approved):

PR #658, opened 2023-05-31
feat: add `backdrop-filter` to range track & bg

luwes:
related issue https://github.com/muxinc/media-chrome/discussions/646


cjpillsbury (approved):
LGTM. Looks like we have a distinction between "track" and "bar" implicitly. We may want to make sure we're consistent with that language.

PR #660, opened 2023-06-02
feat: media-icon experiment

luwes:
Related discussion https://github.com/muxinc/media-chrome/discussions/470

test url: https://media-chrome-git-fork-luwes-media-icon-mux.vercel.app/examples/vanilla/control-elements/media-icon.html

a quick experiment on how there could be icon packs as a module import and consumable with a new `<media-icon>` custom element. 

the resolver callback lets you request the single icon however you want. 

I've implemented the default icon set as a JSON map but it could also be an url request for each icon or a querySelector in a template element holding each SVG element.

it does add some complexity, I'm not fully convinced it's worth it. 
maybe it's best left to the theme developer.

### DIFF
--- a/examples/vanilla/control-elements/media-icon.html
+++ b/examples/vanilla/control-elements/media-icon.html
@@ -1,0 +1,23 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta name="viewport" content="width=device-width" />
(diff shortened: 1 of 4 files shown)


luwes:
closing for the time being

PR #664, opened 2023-06-06
feat: Add preview chapter state support (WIP).

cjpillsbury:
Adding for reference and making available branch for baseline of future efforts. Callouts:
* Need to complete ARIA impl for media-chapter-display (may want to rename media-preview-chapter-display)
* There are some assumptions on the preview time around thumbnails. We should be able to find and propagate chapter data, thumbnail data, both, or neither, depending on what's available.
* There is not yet any chapter marker support for the timeline presentation in media-time-range. We may want to hold off on this until we support state propagation via props instead of attrs.


luwes:
this PR was merged in #755 🎉

PR #665, opened 2023-06-06
Feat/props state propagation

cjpillsbury:
This PR:
1. Adds support for propagating state via `props` if/when available
2. Detecting whether or not a descendant is a "media state receiver" based on the presence of corresponding properties
3. Prefers using media state props over attrs (detected via `observedAttributes` or `mediachromeattributes`) when both are present
4. Adds tests for these defined behaviors/relationships
5. A couple of bug fixes discovered while implementing

### DIFF
--- a/src/js/constants.js
+++ b/src/js/constants.js
@@ -61,6 +61,37 @@ export const MediaStateReceiverAttributes = {
   MEDIA_CONTROLLER: 'mediacontroller',
 };
 
+export const MediaUIProps = {
+  MEDIA_AIRPLAY_UNAVAILABLE: 'mediaAirplayUnavailable',
+  MEDIA_FULLSCREEN_UNAVAILABLE: 'mediaFullscreenUnavailable',
+  MEDIA_PIP_UNAVAILABLE: 'mediaPipUnavailable',
+  MEDIA_CAST_UNAVAILABLE: 'mediaCastUnavailable',
+  MEDIA_PAUSED: 'mediaPaused',
+  MEDIA_HAS_PLAYED: 'mediaHasPlayed',
+  MEDIA_ENDED: 'mediaEnded',
+  MEDIA_MUTED: 'mediaMuted',
+  MEDIA_VOLUME_LEVEL: 'mediaVolumeLevel',
+  MEDIA_VOLUME: 'mediaVolume',
+  MEDIA_VOLUME_UNAVAILABLE: 'mediaVolumeUnavailable',
+  MEDIA_IS_PIP: 'mediaIsPip',
+  MEDIA_IS_CASTING: 'mediaIsCasting',
+  MEDIA_SUBTITLES_LIST: 'mediaSubtitlesList',
+  MEDIA_SUBTITLES_SHOWING: 'mediaSubtitlesShowing',
+  MEDIA_IS_FULLSCREEN: 'mediaIsFullscreen',
+  MEDIA_PLAYBACK_RATE: 'mediaPlaybackRate',
+  MEDIA_CURRENT_TIME: 'mediaCurrentTime',
+  MEDIA_DURATION: 'mediaDuration',
+  MEDIA_SEEKABLE: 'mediaSeekable',
+  MEDIA_PREVIEW_TIME: 'mediaPreviewTime',
+  MEDIA_PREVIEW_IMAGE: 'mediaPreviewImage',
+  MEDIA_PREVIEW_COORDS: 'mediaPreviewCoords',
+  MEDIA_LOADING: 'mediaLoading',
+  MEDIA_BUFFERED: 'mediaBuffered',
+  MEDIA_STREAM_TYPE: 'mediaStreamType',
+  MEDIA_TARGET_LIVE_WINDOW: 'mediaTargetLiveWindow',
+  MEDIA_TIME_IS_LIVE: 'mediaTimeIsLive',
+}
+
 export const MediaUIAttributes = {
   MEDIA_AIRPLAY_UNAVAILABLE: 'mediaairplayunavailable',
   MEDIA_FULLSCREEN_UNAVAILABLE: 'mediafullscreenunavailable',
--- a/src/js/constants.js
+++ b/src/js/constants.js
@@ -23,39 +23,6 @@ export const MediaUIEvents = {
   UNREGISTER_MEDIA_STATE_RECEIVER: 'unregistermediastatereceiver',
 };
 
-export const MediaStateChangeEvents = {
-  MEDIA_AIRPLAY_UNAVAILABLE: 'mediaairplayunavailablechange',
-  MEDIA_FULLSCREEN_UNAVAILABLE: 'mediafullscreenunavailablechange',
-  MEDIA_PIP_UNAVAILABLE: 'mediapipunavailablechange',
-  MEDIA_CAST_UNAVAILABLE: 'mediacastunavailablechange',
-  MEDIA_PAUSED: 'mediapausedchange',
-  MEDIA_HAS_PLAYED: 'mediahasplayedchange',
-  MEDIA_ENDED: 'mediaendedchange',
-  MEDIA_MUTED: 'mediamutedchange',
-  MEDIA_VOLUME_LEVEL: 'mediavolumelevelchange',
-  MEDIA_VOLUME: 'mediavolumechange',
-  MEDIA_VOLUME_UNAVAILABLE: 'mediavolumeunavailablechange',
-  MEDIA_IS_PIP: 'mediaispipchange',
-  MEDIA_IS_CASTING: 'mediaiscastingchange',
-  MEDIA_SUBTITLES_LIST: 'mediasubtitleslistchange',
-  MEDIA_SUBTITLES_SHOWING: 'mediasubtitlesshowingchange',
-  MEDIA_IS_FULLSCREEN: 'mediaisfullscreenchange',
-  MEDIA_PLAYBACK_RATE: 'mediaplaybackratechange',
-  MEDIA_CURRENT_TIME: 'mediacurrenttimechange',
-  MEDIA_DURATION: 'mediadurationchange',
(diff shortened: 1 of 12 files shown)


src/js/constants.js
```diff
@@ -61,6 +61,37 @@ export const MediaStateReceiverAttributes = {
   MEDIA_CONTROLLER: 'mediacontroller',
 };
 
+export const MediaUIProps = {
```
cjpillsbury, line 64: Should we consider deriving the various other dictionary objects from this for easier maintainability? i.e. make`MediaUIAttributes` and `MediaStateChangeEvents` just a `reduce()` translation of this?

src/js/controller.js
```diff
@@ -175,7 +173,7 @@ export const MediaUIStates = {
 
       // Account for cases where metadata from slotted media has an "empty" seekable (CJP)
       if (!start && !end) return undefined;
-      return [Number(start.toFixed(3)), Number(end.toFixed(3))].join(':');
+      return [Number(start.toFixed(3)), Number(end.toFixed(3))];
```
cjpillsbury, line 176: was going back and forth on this: Should we truncate the values here or keep the precision for "real" values and only truncate on serialization? I leaned in this direction, only because our media state receiver component state is stored via attrs anyway (e.g. `set mediaSeekable` -> setAttribute('mediaseekable', serializedVal)` to store the value), so unless/until we move to private vars for state "source of truth", this felt like the right approach.

src/js/media-time-range.js
```diff
@@ -351,7 +351,7 @@ class MediaTimeRange extends MediaChromeRange {
     return getNumericAttr(this, MediaUIAttributes.MEDIA_PLAYBACK_RATE) ?? 1;
   }
 
-  set mediaPlaybackrate(value) {
+  set mediaPlaybackRate(value) {
```
luwes, line 354: good catch 👍

src/js/constants.js
luwes: I like that 👍

src/js/media-controller.js
```diff
@@ -496,8 +499,29 @@ const getMediaUIAttributesFrom = (child) => {
   );
 };
 
+const hasMediaUIProps = (mediaStateReceiverCandidate) => {
+  if (
+    mediaStateReceiverCandidate.nodeName?.includes('-')
+    && !!window.customElements.get(mediaStateReceiverCandidate.nodeName?.toLowerCase())
+    && !(mediaStateReceiverCandidate instanceof window.customElements.get(mediaStateReceiverCandidate.nodeName.toLowerCase()))
+  ) {
+    window.customElements.upgrade(mediaStateReceiverCandidate);
```
luwes, line 508: my s1 here is that we shouldn't upgrade every media state receiver to accept state.

this is a solution for that 
https://github.com/muxinc/media-chrome/blob/ef110153af0d0d6f8fb5b9fb11b4e8e0a3a9607d/src/js/media-theme-element.js#L103-L111

or keep using attributes for propagating state  😅 

---

just checking the code base, we do another unneeded upgrade here 
https://github.com/muxinc/media-chrome/blob/7930c204943985fd568fde377243ea1949456ec8/src/js/media-controller.js#L485-L485
we can get that with`customElements.get('m-c').observedAttributes`

luwes (dismissed):
we shouldn't upgrade eagerly I think

src/js/controller.js
luwes: I agree, 3 decimals precision seems sufficient

src/js/media-controller.js
cjpillsbury: Per out of band convo, we're planning on at least discussing these orthogonal architectural changes before merging this PR.

luwes (approved):
LGTM, follow up with the upgradeProperties logic

src/js/controller.js
```diff
@@ -448,7 +451,7 @@ export const MediaUIStates = {
   MEDIA_SUBTITLES_LIST: {
     get: function (controller) {
       // TODO: Move to non attr specific values
-      return stringifyTextTrackList(getSubtitleTracks(controller)) || undefined;
+      return getSubtitleTracks(controller).map(({ kind, label, language }) => ({ kind, label, language })) || undefined;
```
heff, line 454: Lets return an empty array here, or null. Or help me understand why returning `undefined` is correct. I'm not sure why we'd ever do that over something more distinguishable from "property does not exist".

src/js/media-controller.js
```diff
@@ -239,19 +240,20 @@ class MediaController extends MediaContainer {
...
     // Should hasPlayed refer to the media element or the controller?
     // i.e. the poster might re-show if not handled by the poster el
-    this.propagateMediaState(MediaUIAttributes.MEDIA_PAUSED, true);
+    this.propagateMediaState(MediaUIProps.MEDIA_PAUSED, true);
   }
 
   propagateMediaState(stateName, state) {
-    const previousState = this.getAttribute(stateName);
+    const attrName = stateName.toLowerCase();
+    const previousState = this.getAttribute(attrName);
```
heff, line 248: This still relies on the controller el attrs holding state? For some reason I thought we'd moved away from that with the controller updates.

src/js/media-controller.js
```diff
@@ -239,19 +240,20 @@ class MediaController extends MediaContainer {
...
 
     propagateMediaState(this.mediaStateReceivers, stateName, state);
 
-    if (previousState === this.getAttribute(stateName)) return;
+    if (previousState === this.getAttribute(attrName)) return;
 
     // TODO: I don't think we want these events to bubble? Video element states don't. (heff)
```
heff, line 254: Want to go ahead and address this TODO for me? 😬 Feels pretty unnecessary to be bubbling all these events. (no pressure)

src/js/media-controller.js
heff, line 508: Not sure if this line will survive as it is, but is it safe to assume the custom element here is defined yet? (e.g. the script has been loaded and been processed) This seems to assume so.

src/js/media-controller.js
```diff
@@ -496,8 +499,29 @@ const getMediaUIAttributesFrom = (child) => {
...
 
+const hasMediaUIProps = (mediaStateReceiverCandidate) => {
+  if (
+    mediaStateReceiverCandidate.nodeName?.includes('-')
+    && !!window.customElements.get(mediaStateReceiverCandidate.nodeName?.toLowerCase())
+    && !(mediaStateReceiverCandidate instanceof window.customElements.get(mediaStateReceiverCandidate.nodeName.toLowerCase()))
+  ) {
+    window.customElements.upgrade(mediaStateReceiverCandidate);
+  }
+  return MEDIA_UI_PROP_NAMES.some(propName => propName in mediaStateReceiverCandidate);
```
heff, line 510: Seeing this implemented, I'm feeling like it might start pushing on the safeness of the `media` prefix. No changes, just thinking about it.

heff (commented):
👍 In general. I want to chat about return values before setting this API in stone.

src/js/constants.js
cjpillsbury: we can apply ad hoc rules to do things like omit, but I'd rather treat them as exceptions in the codebase. Should be easier to maintain.

src/js/controller.js
cjpillsbury: Broadly, I'd lean towards not changing our decisions on data modeling in this PR, but I'd also lean towards keeping the data simple for now. @AdamJaggard and I discussed data shape modeling for this during his component props add and he settled on an easier to work with API ("tuples"). When we first introduced seekable, we decided to keep it simple/consistent with the way it works for `HTMLMediaElement` unless/until it makes sense to support more complex cases, which would likely tie into some kind of media-ui-extension.

src/js/media-controller.js
cjpillsbury: Yeah for sure, this came up in previous discussions. As long as we're diligent about it wrt attr/prop names in "built-ins" for HTML (an extension of https://www.w3.org/TR/design-principles/#attribute-reuse) and expect these conflicts to be the exception and not the rule for custom elements, we should be able to introduce some kind of "ignore" logic in our media state receiver detection/propagation in the future.

src/js/media-controller.js
cjpillsbury: `upgrade()` will "force" a define for a custom element that's registered (which we identify via the `customElements.get()`, above).

src/js/media-controller.js
cjpillsbury: No, the media-controller is still the "source of truth" for checking state changes. That can change when we implement the separate media state model/store abstraction.

src/js/controller.js
heff: Makes sense 👍

src/js/media-controller.js
heff: The conclusion was no, this could be breaking, so too much for this PR.

src/js/media-controller.js
heff: Makes sense 👍

src/js/media-controller.js
```diff
@@ -532,17 +533,22 @@ const setAttr = async (child, attrName, attrValue) => {
...
+    return setStringAttr(child, attrName, attrValue);
   }
-  if (Number.isNaN(attrValue)) {
+  // Treat empty arrays as "nothing" values
+  if (Array.isArray(attrValue) && !attrValue.length) {
     return child.removeAttribute(attrName);
   }
 
+  // For "special" values with custom serializers or all other values
   const val = CustomAttrSerializer[attrName]?.(attrValue) ?? attrValue;
```
heff, line 552: Should the check for a custom serializer not come before all the type checking?

heff (approved):
One question but looks good 👍

src/js/media-controller.js
cjpillsbury: with the current impl, no, but your intuition here makes a lot of sense and maybe recommends some refactors in our projections. Gonna keep as is for now to not hold up the PR.

PR #669, opened 2023-06-14
improv: add notoggle attribute to media-time-display

dylanjha:


dylanjha:
tested out and this is working on basic.html example 👍🏼

cjpillsbury (changes requested):
nitpick(blocking): make sure you add a corresponding `noToggle` getter/setter.
suggestion(non-blocking): we should remove probably also remove the `cursor` and `hover` styles when `notoggle` is true. See the constructor for the relevant styles (can simply update the default/fallback value of `--media-control-hover-background` if that's easier). Feel free to hmu if you wanna chat through any of this.

src/js/media-time-display.js
```diff
@@ -86,7 +87,7 @@ class MediaTimeDisplay extends MediaTextDisplay {
   #slot;
 
   static get observedAttributes() {
-    return [...super.observedAttributes, ...CombinedAttributes, 'disabled'];
+    return [...super.observedAttributes, ...CombinedAttributes, 'disabled', 'notoggle'];
```
cjpillsbury, line 90: nitpick(blocking): add `NO_TOGGLE: 'notoggle'` to `Attributes` object instead of declaring it inline here.

src/js/media-time-display.js
```diff
@@ -141,6 +142,9 @@ class MediaTimeDisplay extends MediaTextDisplay {
   }
 
   toggleTimeDisplay() {
+    if (this.hasAttribute('notoggle')) {
```
cjpillsbury, line 145: nitpick(blocking): use aforementioned `Attributes.NO_TOGGLE` here for consistency

dylanjha: should we use the prop instead?

cjpillsbury (approved):
Looks 😎 🎉.

src/js/media-time-display.js
```diff
@@ -106,6 +108,9 @@ class MediaTimeDisplay extends MediaTextDisplay {
       'background',
       'var(--media-control-hover-background, rgba(50 50 70 / .7))'
     );
+    const { style: noToggle } = getOrInsertCSSRule(this.shadowRoot, ':host([notoggle])');
+    noToggle.setProperty('cursor', 'default');
+    noToggle.setProperty('background', 'var(--media-text-background, var(--media-control-background, var(--media-secondary-color, rgb(20 20 30 / .7))))');
```
luwes, line 113: could this line be removed if hover selector was changed to `:host(:hover:not([notoggle]))` or equivalent?
save some bytes

cjpillsbury: ah that's a good idea. I think both styles can probably be applied using the same logic, actually.

Issue #656, opened 2023-05-30
Inconsistent MediaTimeRange preview and submitted value 


HermanBilous:
Steps to reproduce:
1. Go to https://media-chrome.mux.dev/examples/vanilla/advanced.html
2. Click on duration display to enable positive time
3. Try seeking to 1:00, 1:29 etc.

Expected result:
Current time after click is exactly the same as on preview.

Actual result:
Current time after seeking is sometimes 1s off compared to preview.

Alternative steps to reproduce:
1. Go to https://media-chrome.mux.dev/examples/vanilla/advanced.html
2. Open chrome devtools, in "Inspect" tab find media-controller element.
3. Click and hold on any position in seek bar, while observing mediacurrenttime and mediapreviewtime attributes in media-controller element.

Expected result:
mediacurrenttime and mediapreviewtime attrs are the same

Actual result:
there is a slight difference between the values

Browser - Google Chrome Version 113.0.5672.126 (Official Build) (x86_64)
System - MacOS Big Sur version 11.6.8 (20G730)

Seems like the issue is related to the fact that preview is handled by the mouse position relative to seek bar size, while the actual value is received from an input type="range" element.

heff:
Interesting. Can confirm I see the issue on my side. It's a good bet it has something to do with calculating the preview time from an element size that includes more than just the range area.

HermanBilous:
Could someone look into doing a release with this fix? @heff @luwes ?

Thanks!

luwes:
thanks for reminding. just release v1.1.1 with the fix!

PR #670, opened 2023-06-16
fix: time range preview and value inconsistency

HermanBilous:
Fixes https://github.com/muxinc/media-chrome/issues/656

As it turned out, input[type=range] makes half the width of thumb a "dead zone" where the value is either max or min. If we account for that when calculating preview values, we get the "right" result in preview. On a video of 3 minutes, the difference is < 0.01s between preview and input value.


heff:
Awesome! Thanks @HermanBilous. @luwes any thoughts on the implementation here since you've worked on the bar here more recently?

luwes (approved):
thanks for the contribution! 

this makes sense to me, also verified by these answers 
https://stackoverflow.com/questions/48880523/how-to-precisely-find-thumb-position-of-input-type-range

did a quick test and worked as intended.

PR #671, opened 2023-06-18
fix: use globalThis i/o window (declarative shadow DOM)

luwes:
I'm making some progress with a generic solution for SSR. 
Got it working in a plain Node app and Cloudflare worker, working on Next.js now... (added https://wesc-nextjs.vercel.app/)
https://github.com/luwes/wesc

The worker is pretty cool, it can take any URL with Media Chrome components and add DSD to it 
https://wesc.luwes.workers.dev/?url=https%3A%2F%2Fmedia-chrome.mux.dev%2Fexamples%2Fvanilla%2Fadvanced.html
https://wesc.luwes.workers.dev/?url=https://www.media-chrome.org/

There is a limit on CPU time for cloudflare workers so it might time out but I haven't seen it yet.
SSR'ing 3 Media Chrome players took like 35ms locally.

---

one of the things that came up is that some frameworks like Next.js recommend server checks with `typeof window === 'undefined'` https://stackoverflow.com/questions/49411796/how-do-i-detect-whether-i-am-on-server-on-client-in-next-js

so we don't want to be shimming `window` but `globalThis` with a DOM shim otherwise we are making the server / build look like a client environment.

### DIFF
--- a/docs/src/pages/docs/en/keyboard-shortcuts.md
+++ b/docs/src/pages/docs/en/keyboard-shortcuts.md
@@ -27,7 +27,7 @@ Example (`keysused` attribute):
 Example (`keysUsed` property):
 
 ```js
-class MyInteractiveElement extends window.HTMLElement {
+class MyInteractiveElement extends globalThis.HTMLElement {
   get keysUsed() {
     return ['Enter', ' '];
   }
--- a/scripts/react/build.js
+++ b/scripts/react/build.js
@@ -243,15 +243,13 @@ const setupGlobalsAsync = async () => {
   const customElementNames = await import(
     path.join(projectRoot, 'dist', 'utils', 'server-safe-globals.js')
   ).then((exports) => {
-    globalThis.window = exports.Window;
-    globalThis.document = exports.Document;
-    globalThis.window.document = globalThis.document;
-    window.customElementNames = [];
-    window.customElements.define = (name, _classRef) =>
-      window.customElementNames.push(name);
+    Object.assign(globalThis, exports.globalThis);
+    globalThis.customElementNames = [];
+    globalThis.customElements.define = (name, _classRef) =>
+      globalThis.customElementNames.push(name);
     // NOTE: The current implementation relies on the fact that `customElementNames` will be mutated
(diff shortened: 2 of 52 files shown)


luwes:
Got a Next.js example running! https://wesc-nextjs.vercel.app/

In the new Next.js app router it's really minimal what you need to add because it supports async components

- DOM shim: https://github.com/luwes/wesc/blob/main/examples/nextjs/app/page.js#L1
- WeSC container: https://github.com/luwes/wesc/blob/main/examples/nextjs/components/player.js#L26

heff:
> so we don't want to be shimming window but globalThis with a DOM shim otherwise we are making the server / build look like a client environment.

Have you actually run into this problem? We're not setting window globally, just locally to the project, so any `typeof window` check shouldn't be affected by our local usage of a variable named `window` right?

`globalThis` is already "server safe", so it feels backwards to then create a "server safe shim" for it. It also doesn't have a consistent API across platforms, so how do you _shim_ it? We're intentionally shimming the `window` API, not globalThis so 
components expecting `window` functions don't blow up on the server.

luwes:
@heff I haven't run into any issues but I think it would break that `typeof window === 'undefined'` check though.

I know the window export variable is local to Media Chrome but the DOM API's have to be shimmed on a global object right, either window or globalThis.

In the window scenario this will happen.

```js
// This will be true if JSDOM or Linkedom shimmed window with equivalent API's.
const isShimmed = Object.keys(windowShim)
  .every(key => key in window);

// If we keep using window the `window` in the ternary else has to be shimmed with equivalent API's for SSG / SSR to work
export const Window = isServer && !isShimmed ? windowShim : window;

// => typeof window === 'undefined' won't be true anymore 
```

heff:
> but the DOM API's have to be shimmed on a global object right, either window or globalThis.

No? We've shimmed all the DOM APIs we've needed [on windowShim](https://github.com/muxinc/media-chrome/blob/main/src/js/utils/server-safe-globals.js#L15). As long as we continue to import our internal version of `window` whenever a window API is needed, no shim has to go on a global object.

I'm not clear on if what you're doing here requires a bunch of additional window APIs though. I think I need you to walk me through this feature in general to get caught up.

cjpillsbury (approved):
The future is now.

PR #672, opened 2023-06-21
Add typesVersions in package.json

### LINKED ISSUE #680: What is the correct import for React?
Hello,

For some reason in my next.js project, importing from `media-chrome/dist/react` as per the docs does not work, but the following does, despite being flagged as incorrect by typescript:

```
import {
  MediaController,
  MediaControlBar,
  MediaTimeRange,
  MediaTimeDisplay,
  MediaVolumeRange,
  MediaPlayButton,
  MediaMuteButton,
} from 'media-chrome/react'
```

Why does this work and the doc-specified import path not, and which should I be doing? 

Thanks!

Simon

jonabc:
fix #680 

Thanks for adding exports declarations for EMS and CJS!  I was testing [the v1.x version of this package in the sanity-plugin-mux-input plugin](https://github.com/sanity-io/sanity-plugin-mux-input/pull/285) and TypeScript was having problems with `import { ... } from 'media-chrome/react'` which stemmed from problems with type resolution.

The solution I came across was adding in these `typesVersions` declarations which routes type resolution for react vs non-react imports to the expected files under `dist` to match what's declared under the `exports` field.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,12 @@
   },
   "main": "dist/index.js",
   "type": "module",
+  "typesVersions": {
+    "*": {
+      "react": ["dist/react/index.d.ts"],
+      "*": ["dist/index.d.ts"]
+    }
+  },
   "exports": {
     ".": {
       "import": "./dist/index.js",


jonabc:
@cjpillsbury have you had a chance to take a look at this yet?  No rush, I'm just trying to figure out where it's at.

cjpillsbury:
@jonabc hey just started doing some smoke testing this morning. Unfortunately, it's looking like we're going to need something a bit more complicated for the types. tl;dr - we'd want to fix this case without breaking our default type declaration file mappings for *all* valid dist module uses. E.g. - If you take a look at https://cdn.jsdelivr.net/npm/media-chrome@1.1.1/dist/ (for convenience of browsing the dist), you'll notice that we have per-module declaration files available. We also have different module "flavors" of these (e.g. ESM, CJS) for both React and base web components. I haven't dug deeply, but, given that we've introduced the more modern `exports` in our package (https://github.com/muxinc/media-chrome/blob/main/package.json#L13-L35), I _**think**_/hope we can take advantage of TypeScript's support of that (See, e.g.: https://www.typescriptlang.org/docs/handbook/esm-node.html#packagejson-exports-imports-and-self-referencing). I probably won't have time to experiment with this just yet, but if you were interested, we always love some community love. Feel free to follow up with any additional questions/comments/concerns!

cjpillsbury:
> but there is a validation rule on sanity studio plugins that requires moduleResolution to be set as node 😢

well that's unfortunate but also good to know with any potential solution.

> had multiple incompatible changes that prevents something else from working 🤷

yeah this is the thread we're always needling/tightrope we're always walking. My quick smoke test was to try to target any of the non-default dist paths and see if typescript worked before vs. after your change, using our in-repo [Next.JS + TypeScript example](https://github.com/muxinc/media-chrome/tree/main/examples/nextjs-with-typescript) to do some very basic tests. Assuming I can peel away some time, I am definitely happy to help figure out/test a path forward

luwes (approved):
LGTM. tested this with https://arethetypeswrong.github.io/

we might want to remove the `"*": ["dist/index.d.ts"]` because it's done automatically

luwes:
SCR-20230725-cwf

SCR-20230725-ecj

PR #674, opened 2023-06-23
fix: Make computation for breakpoints in ResizeObserver async.

cjpillsbury:
Ensures we do not end up with loop limit exceeded RTEs when breakpoints computations take too long.


luwes (approved):
LGTM!

PR #675, opened 2023-06-27
fix: React createElement function signature

luwes:
not a biggie it seems but children are spread in the React.createElement function signature
https://react.dev/reference/react/createElement

### DIFF
--- a/scripts/react/build.js
+++ b/scripts/react/build.js
@@ -33,8 +33,8 @@ const toReactComponentStr = (config) => {
   const { elementName } = config;
   const ReactComponentName = toPascalCase(elementName);
   return `/** @type { import("react").HTMLElement } */
-const ${ReactComponentName} = React.forwardRef(({ children, ...props }, ref) => {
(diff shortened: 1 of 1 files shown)


AdamJaggard (commented):
random question: I'm assuming the answer is yes 😅, do we guarantee that the `children` variable is always spreadable? only because in "normal" react it can be lots of different types, I'm guessing we know what it is for our use case making this fine

luwes:
I'm not 100% sure but I think that's what the JSX should return. 
it's for these wrappers

```js
/** @type { import("react").HTMLElement } */
const MediaCastButton = React.forwardRef(({ children, ...props }, ref) => {
  return React.createElement('media-cast-button', toNativeProps({ ...props, ref }), children);
});

export { MediaCastButton };
```

these are the output of the current react script, I think this is roughly translated to JSX similar to

```jsx
<MediaCastButton>
  <>
    <Child />
    <Child />
  </>
</MediaCastButton>
```

cjpillsbury (approved):

PR #683, opened 2023-07-05
Rendition select

luwes:
Related #387

test urls: 

- https://media-chrome-git-fork-luwes-rendition-select-mux.vercel.app/examples/vanilla/control-elements/media-rendition-listbox.html
- https://media-chrome-git-fork-luwes-rendition-select-mux.vercel.app/examples/vanilla/control-elements/media-rendition-selectmenu.html

### DIFF
--- a/src/js/constants.js
+++ b/src/js/constants.js
@@ -59,7 +59,7 @@ export const MediaUIProps = {
   MEDIA_TARGET_LIVE_WINDOW: 'mediaTargetLiveWindow',
   MEDIA_TIME_IS_LIVE: 'mediaTimeIsLive',
   MEDIA_RENDITION_LIST: 'mediaRenditionList',
-  MEDIA_RENDITION_ENABLED: 'mediaRenditionEnabled',
+  MEDIA_RENDITIONS_ENABLED: 'mediaRenditionsEnabled',
   MEDIA_RENDITION_ACTIVE: 'mediaRenditionActive',
 };
 
--- a/src/js/constants.js
+++ b/src/js/constants.js
@@ -18,6 +18,7 @@ export const MediaUIEvents = {
   MEDIA_SHOW_SUBTITLES_REQUEST: 'mediashowsubtitlesrequest',
   MEDIA_DISABLE_SUBTITLES_REQUEST: 'mediadisablesubtitlesrequest',
   MEDIA_PLAYBACK_RATE_REQUEST: 'mediaplaybackraterequest',
+  MEDIA_RENDITION_REQUEST: 'mediarenditionrequest',
   MEDIA_SEEK_TO_LIVE_REQUEST: 'mediaseektoliverequest',
   REGISTER_MEDIA_STATE_RECEIVER: 'registermediastatereceiver',
   UNREGISTER_MEDIA_STATE_RECEIVER: 'unregistermediastatereceiver',
@@ -57,6 +58,8 @@ export const MediaUIProps = {
   MEDIA_STREAM_TYPE: 'mediaStreamType',
   MEDIA_TARGET_LIVE_WINDOW: 'mediaTargetLiveWindow',
   MEDIA_TIME_IS_LIVE: 'mediaTimeIsLive',
+  MEDIA_RENDITION_LIST: 'mediaRenditionList',
+  MEDIA_RENDITION_ACTIVE: 'mediaRenditionActive',
 };
 
 const MediaUIPropsEntries = /** @type {[keyof MediaUIProps, string][]} */ (
--- a/src/js/constants.js
+++ b/src/js/constants.js
@@ -59,8 +59,7 @@ export const MediaUIProps = {
   MEDIA_TARGET_LIVE_WINDOW: 'mediaTargetLiveWindow',
   MEDIA_TIME_IS_LIVE: 'mediaTimeIsLive',
   MEDIA_RENDITION_LIST: 'mediaRenditionList',
-  MEDIA_RENDITIONS_ENABLED: 'mediaRenditionsEnabled',
-  MEDIA_RENDITION_ACTIVE: 'mediaRenditionActive',
+  MEDIA_RENDITION_SELECTED: 'mediaRenditionSelected',
 };
 
 const MediaUIPropsEntries = /** @type {[keyof MediaUIProps, string][]} */ (
--- a/src/js/constants.js
+++ b/src/js/constants.js
@@ -59,6 +59,7 @@ export const MediaUIProps = {
   MEDIA_TARGET_LIVE_WINDOW: 'mediaTargetLiveWindow',
   MEDIA_TIME_IS_LIVE: 'mediaTimeIsLive',
   MEDIA_RENDITION_LIST: 'mediaRenditionList',
+  MEDIA_RENDITION_ENABLED: 'mediaRenditionEnabled',
   MEDIA_RENDITION_ACTIVE: 'mediaRenditionActive',
 };
 
--- a/src/js/controller.js
+++ b/src/js/controller.js
@@ -483,7 +483,7 @@ export const MediaUIStates = {
   MEDIA_RENDITION_SELECTED: {
     get: function (controller) {
       const { media } = controller;
-      return media?.videoRenditions?.[media.videoRenditions?.selectedIndex];
+      return media.videoRenditions?.selectedIndex;
     },
     mediaEvents: ['emptied'],
     videoRenditionsEvents: ['addrendition', 'removerendition', 'change'],
--- a/src/js/controller.js
+++ b/src/js/controller.js
@@ -483,7 +483,7 @@ export const MediaUIStates = {
   MEDIA_RENDITION_SELECTED: {
     get: function (controller) {
       const { media } = controller;
-      return media.videoRenditions?.selectedIndex;
+      return media?.videoRenditions?.[media.videoRenditions?.selectedIndex];
     },
     mediaEvents: ['emptied'],
     videoRenditionsEvents: ['addrendition', 'removerendition', 'change'],
--- a/src/js/controller.js
+++ b/src/js/controller.js
@@ -475,11 +475,11 @@ export const MediaUIStates = {
   MEDIA_RENDITION_LIST: {
     get: function (controller) {
       const { media } = controller;
-      const selected = media.videoTracks?.[media.videoTracks?.selectedIndex ?? 0];
+      const selected = media?.videoTracks?.[media.videoTracks?.selectedIndex ?? 0];
 
       if (!selected) return [];
 
-      return [...selected.renditions ?? []].map(({ id, height }) => ({ id, height }));
+      return [...selected.renditions ?? []];
     },
     mediaEvents: ['loadstart'],
     renditionListEvents: ['addrendition', 'removerendition'],
@@ -487,7 +487,7 @@ export const MediaUIStates = {
   MEDIA_RENDITION_ENABLED: {
     get: function (controller) {
       const { media } = controller;
-      const selected = media.videoTracks?.[media.videoTracks?.selectedIndex ?? 0];
+      const selected = media?.videoTracks?.[media.videoTracks?.selectedIndex ?? 0];
 
       if (!selected) return [];
 
@@ -496,9 +496,7 @@ export const MediaUIStates = {
       // If all renditions are enabled it means `auto` is selected.
       if (renditions.every(r => r.enabled)) return [];
 
-      return renditions
-        .filter(r => r.enabled)
-        .map(({ id, height }) => ({ id, height }));
+      return renditions.filter(r => r.enabled);
     },
     mediaEvents: ['loadstart'],
     renditionListEvents: ['addrendition', 'removerendition', 'change'],
@@ -506,14 +504,12 @@ export const MediaUIStates = {
   MEDIA_RENDITION_ACTIVE: {
     get: function (controller) {
       const { media } = controller;
-      const selected = media.videoTracks?.[media.videoTracks?.selectedIndex ?? 0];
+      const selected = media?.videoTracks?.[media.videoTracks?.selectedIndex ?? 0];
 
       if (!selected) return [];
 
       const renditions = [...selected.renditions ?? []];
-      return renditions
-        .filter(r => r.active)
-        .map(({ id, height }) => ({ id, height }));
+      return renditions.filter(r => r.active);
     },
     mediaEvents: ['loadstart'],
     renditionListEvents: ['renditionchange'],
@@ -822,7 +818,7 @@ export const MediaUIRequestHandlers = {
     const selected = media.videoTracks?.[media.videoTracks?.selectedIndex ?? 0];
 
     for (const rendition of selected?.renditions ?? []) {
-      rendition.enabled = renditionId == 'auto' || rendition.id == renditionId;
+      rendition.enabled = renditionId === '' || rendition.id == renditionId;
     }
   }
 };
--- a/src/js/controller.js
+++ b/src/js/controller.js
@@ -789,13 +789,8 @@ export const MediaUIRequestHandlers = {
     if (!media?.videoRenditions) return;
 
     const renditionId = event.detail;
-    const index = [...media.videoRenditions]
-      .findIndex(r => r.id == renditionId);
-
-    console.log(index, media.videoRenditions.selectedIndex);
+    const index = [...media.videoRenditions].findIndex(r => r.id == renditionId);
 
     media.videoRenditions.selectedIndex = index;
-
-    console.log(index, media.videoRenditions.selectedIndex);
   }
 };
--- a/src/js/controller.js
+++ b/src/js/controller.js
@@ -483,7 +483,7 @@ export const MediaUIStates = {
   MEDIA_RENDITION_SELECTED: {
     get: function (controller) {
       const { media } = controller;
-      return media?.videoRenditions?.[media.videoRenditions?.selectedIndex];
+      return media?.videoRenditions?.[media.videoRenditions?.selectedIndex]?.id;
     },
     mediaEvents: ['emptied'],
     videoRenditionsEvents: ['addrendition', 'removerendition', 'change'],
@@ -794,6 +794,8 @@ export const MediaUIRequestHandlers = {
     const renditionId = event.detail;
     const index = [...media.videoRenditions].findIndex(r => r.id == renditionId);
 
-    media.videoRenditions.selectedIndex = index;
+    if (media.videoRenditions.selectedIndex != index) {
+      media.videoRenditions.selectedIndex = index;
+    }
   }
 };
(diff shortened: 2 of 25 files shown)


src/js/controller.js
```diff
@@ -769,4 +811,14 @@ export const MediaUIRequestHandlers = {
...
     media.currentTime = seekable.end(seekable.length - 1);
   },
+  MEDIA_RENDITION_REQUEST: (media, event) => {
+    if (!media) return;
+
+    const renditionId = event.detail;
+    const selected = media.videoTracks?.[media.videoTracks?.selectedIndex ?? 0];
+
+    for (const rendition of selected?.renditions ?? []) {
+      rendition.enabled = renditionId === '' || rendition.id == renditionId;
```
heff, line 821: Requesting a rendition of _empty string_ feels odd. Could we make "auto" a real value?

src/js/experimental/media-rendition-button.js
```diff
@@ -0,0 +1,77 @@
...
+ * @attr {string} mediarenditionactive - (read-only) Set to the current active rendition.
+ *
+ * @cssproperty [--media-rendition-button-display = inline-flex] - `display` property of button.
+ */
+class MediaRenditionButton extends MediaChromeButton {
+  static get observedAttributes() {
+    return [
+      ...super.observedAttributes,
+      MediaUIAttributes.MEDIA_RENDITION_ENABLED,
+      MediaUIAttributes.MEDIA_RENDITION_ACTIVE,
```
heff, line 48: This is kind of showing my point from above, that the UI doesn't care about `MEDIA_RENDITION_LIST` as it is.

src/js/experimental/media-rendition-listbox.js
```diff
@@ -0,0 +1,153 @@
...
+
+    super.disconnectedCallback();
+  }
+
+  get mediaRenditionList() {
+    return this.#renditionList;
+  }
+
+  set mediaRenditionList(list) {
+    this.removeAttribute(MediaUIAttributes.MEDIA_RENDITION_LIST);
```
heff, line 76: Why remove the attr here?

src/js/experimental/media-rendition-listbox.js
```diff
@@ -0,0 +1,153 @@
...
+
+    this.#renditionList = list;
+    this.#render();
+  }
+
+  get mediaRenditionEnabled() {
+    if (this.value) {
+      return this.mediaRenditionList.filter(({ id }) => id == this.value);
```
heff, line 84: This can only ever return one item right? So no need to return an array?

src/js/experimental/media-rendition-listbox.js
```diff
@@ -0,0 +1,153 @@
...
+    }
+    return undefined;
```
heff, line 86: I'd prefer null here. I forget where @cjpillsbury and I landed on arrays specifically.

src/js/experimental/media-rendition-listbox.js
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  set mediaRenditionEnabled(list) {
+    this.removeAttribute(MediaUIAttributes.MEDIA_RENDITION_ENABLED);
```
heff, line 90: Same question as above on removing the attr.

src/js/experimental/media-chrome-listbox.js
```diff
@@ -5,42 +5,39 @@ const template = document.createElement('template');
...
-  media-chrome-listitem[aria-selected="true"] {
+  ::slotted(media-chrome-option[aria-selected="true"]),
+  media-chrome-option[aria-selected="true"] {
     background-color: var(--media-listbox-selected-background, rgb(122 122 184 / .8));
   }
 
-  ::slotted(media-chrome-listitem:hover),
-  media-chrome-listitem:hover {
+  ::slotted(media-chrome-option:hover),
+  media-chrome-option:hover {
```
heff, line 35: Right now hovering on a selected item triggers the hover state, but it shouldn't. We should ignore this when selected.

src/js/controller.js
heff: If this is disabling all other renditions at the same time, so only one can be enabled at a time, then I'm really not sure why we're working with arrays for _enabled_.

src/js/media-container.js
```diff
@@ -351,9 +351,13 @@ class MediaContainer extends window.HTMLElement {
   }
 
   static get observedAttributes() {
-    return [Attributes.AUTOHIDE, Attributes.GESTURES_DISABLED].concat(
-      MEDIA_UI_ATTRIBUTE_NAMES
-    );
+    return [Attributes.AUTOHIDE, Attributes.GESTURES_DISABLED]
+      .concat(MEDIA_UI_ATTRIBUTE_NAMES)
+      .filter(name => ![
```
heff, line 356: Why? (could at least use a comment)

src/js/media-controller.js
```diff
@@ -219,39 +225,71 @@ class MediaController extends MediaContainer {
...
+    this.#handleVideoTrack({ type: 'removeall' });
+
     // Reset to paused state
     // TODO: Can we just reset all state here?
     // Should hasPlayed refer to the media element or the controller?
     // i.e. the poster might re-show if not handled by the poster el
     this.propagateMediaState(MediaUIProps.MEDIA_PAUSED, true);
   }
 
+  #handleVideoTrack = ({ type = '', track = null } = {}) => {
```
heff, line 259: This is complicated. It'd be nice if we could pull this into the above logic more, though I see how it's different because there needs to be a listener on each individual track.

src/js/media-controller.js
```diff
@@ -625,6 +666,18 @@ const propagateMediaState = (els, stateName, val) => {
   });
 };
 
+const getStateValue = (els, stateName) => {
```
heff, line 669: Why was this needed?

heff (commented):
Looks like it solves the need functionally. 👍 

The rendition data structures feel like they could use another pass. There seems to be competing designs happening between media-ui-extensions and what's being implemented here, like returning arrays that can only ever have one value and not actually using RENDITION_LIST anywhere.

src/js/constants.js
```diff
@@ -57,6 +58,9 @@ export const MediaUIProps = {
   MEDIA_STREAM_TYPE: 'mediaStreamType',
   MEDIA_TARGET_LIVE_WINDOW: 'mediaTargetLiveWindow',
   MEDIA_TIME_IS_LIVE: 'mediaTimeIsLive',
+  MEDIA_RENDITION_LIST: 'mediaRenditionList',
+  MEDIA_RENDITION_ENABLED: 'mediaRenditionEnabled',
+  MEDIA_RENDITION_ACTIVE: 'mediaRenditionActive',
```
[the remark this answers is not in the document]
luwes: yes, when the rendition is enabled / disabled it's not immediately playing because a stream switch can take a bit.
when the rendition starts playing then it becomes active...

src/js/controller.js
```diff
@@ -470,13 +470,55 @@ export const MediaUIStates = {
...
+      const selected = media?.videoTracks?.[media.videoTracks?.selectedIndex ?? 0];
+
+      if (!selected) return [];
+
+      return [...selected.renditions ?? []];
+    },
+    mediaEvents: ['loadstart'],
+    renditionListEvents: ['addrendition', 'removerendition'],
+  },
+  MEDIA_RENDITION_ENABLED: {
```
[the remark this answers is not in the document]
luwes: good call, will change to "renditions".

The rendition API supports enabling just a few renditions. Maybe only 270p and 360p for whatever reason. 
This would allow that. 

It's also one of the reasons why the `selectedIndex` property for renditions in media ui ext doesn't make sense.

src/js/controller.js
```diff
@@ -470,13 +470,55 @@ export const MediaUIStates = {
...
+
+      // If all renditions are enabled it means `auto` is selected.
+      if (renditions.every(r => r.enabled)) return [];
+
+      return renditions.filter(r => r.enabled);
+    },
+    mediaEvents: ['loadstart'],
+    renditionListEvents: ['addrendition', 'removerendition', 'change'],
+  },
+  MEDIA_RENDITION_ACTIVE: {
```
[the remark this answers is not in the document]
luwes: it's using arrays in the same fashion as `MEDIA_SUBTITLES_LIST` and `MEDIA_SUBTITLES_SHOWING`.
it makes it easier to use the stringify and parse utilities for all of them.

src/js/controller.js
luwes: the renditions API supports enabling 1, multiple or all renditions so an array makes sense.

via the UI we'll probably not support the multiple enable though...

src/js/experimental/media-rendition-listbox.js
luwes: this is related to @cjpillsbury feedback of making some rendition related API only via props (not reflecting to attrs) but still accept attributes.

the prop / attribute is not reflected anymore so we used this behavior for mediaController.fullscreenElement in removing the attribute if the property was used.

the alternative is if there was ever an attribute set to also keep updating the attribute if ever the property was set but would make it more complex I think.

src/js/media-controller.js
luwes: before we could easily check all the attribute values on media-controller but since we will not set some complex data attributes like mediarenditionlist on <media-controller> we'll have to get the values from one of the control media state receivers. cc @cjpillsbury

src/js/media-controller.js
luwes: yes, that's what I was hinting at in point 1 in the PR description.
I don't like it that much either.

src/js/experimental/media-rendition-button.js
luwes: `MEDIA_RENDITION_LIST` is used in rendition listbox, it still accepts the attribute but it'll prefer to use the property if it is available

src/js/experimental/media-rendition-listbox.js
luwes: this is solved with making this plural mediaRenditionsEnabled because there can be a case w/ multiple renditions enabled

src/js/experimental/media-rendition-listbox.js
luwes: I'm fine with changing. we use `undefined` in more places for props currently I think (e.g. in Mux player) but null is used a lot in DOM API's, getAttribute returns null too, className,

src/js/experimental/media-rendition-listbox.js
heff: Yeah, that makes sense. Complex values shouldn't be sprouted because you're not going to style things with them, and even simplifying the complex values would open up other complexity.

Deleting it is still form of sprouting, but I guess things like `autoplay` already do that.

src/js/experimental/media-rendition-listbox.js
heff: Yeah, let's use null here. And in general save `undefined` specifically for unimplemented APIs and functions that are not intended to ever return anything. That's my experience of how most JS APIs are in HTML.

src/js/media-controller.js
heff: Probably fine but feels a little odd to use a list of stateReceivers as the source of truth for the prior state. I wonder of we could use MediaUIStates, here or if that'd already be changed at this point. This might be a reason to move forward with state properties for media controller specifically.

src/js/experimental/media-rendition-listbox.js
luwes: yes, I can see that. do you think it's better to leave it as is? not remove the attribute.
I'm fine with both I think, it's a pretty rare case

src/js/controller.js
```diff
@@ -470,13 +470,29 @@ export const MediaUIStates = {
...
+    },
+    mediaEvents: ['emptied'],
+    videoRenditionsEvents: ['addrendition', 'removerendition'],
+  },
+  MEDIA_RENDITION_SELECTED: {
+    get: function (controller) {
+      const { media } = controller;
+      return media?.videoRenditions?.[media.videoRenditions?.selectedIndex];
+    },
+    mediaEvents: ['emptied'],
```
heff, line 488: Is there no event we should be listening to for the first load? I worry a little about missing an add rendition event, if that's what's being relied on.

src/js/controller.js
```diff
@@ -470,13 +470,29 @@ export const MediaUIStates = {
...
+      const { media } = controller;
+      return [...media?.videoRenditions ?? []];
+    },
+    mediaEvents: ['emptied'],
+    videoRenditionsEvents: ['addrendition', 'removerendition'],
+  },
+  MEDIA_RENDITION_SELECTED: {
+    get: function (controller) {
+      const { media } = controller;
+      return media?.videoRenditions?.[media.videoRenditions?.selectedIndex];
```
heff, line 486: Nice, the API turned out relatively simple, even without IDs.

I think this property should just return the index value rather than the whole rendition object. As is it's duplicate info with the list, and using just the index means it could still be a simple HTML attribute, even if the list is more complex.

src/js/controller.js
```diff
@@ -769,4 +785,12 @@ export const MediaUIRequestHandlers = {
 
     media.currentTime = seekable.end(seekable.length - 1);
   },
+  MEDIA_RENDITION_REQUEST: (media, event) => {
+    if (!media?.videoRenditions) return;
```
heff, line 789: Let's add a console warning here. i.e. "Rendition selection not supported by this media". Silent failing here could lead to some frustration.

src/js/experimental/media-rendition-button.js
```diff
@@ -0,0 +1,68 @@
...
+  <path d="M13.5 2.5h2v6h-2v-2h-11v-2h11v-2Zm4 2h4v2h-4v-2Zm-12 4h2v6h-2v-2h-3v-2h3v-2Zm4 2h12v2h-12v-2Zm1 4h2v6h-2v-2h-8v-2h8v-2Zm4 2h7v2h-7v-2Z" />
+</svg>`;
+
+const slotTemplate = document.createElement('template');
+slotTemplate.innerHTML = /*html*/`
+  <slot name="icon">${renditionIcon}</slot>
+`;
+
+/**
+ * @attr {string} mediarenditionsenabled - (read-only) Set to the enabled rendition.
```
heff, line 38: "enabled" is wrong now, right?

src/js/experimental/media-rendition-button.js
```diff
@@ -0,0 +1,68 @@
...
+  constructor() {
+    super({ slotTemplate });
+  }
+
+  get mediaRenditionSelected() {
+    return getRenditionListAttr(this, MediaUIAttributes.MEDIA_RENDITION_SELECTED)[0];
+  }
+
+  set mediaRenditionSelected(rendition) {
+    const value = rendition ? [rendition] : null;
```
heff, line 59: It doesn't need to be an array anymore, right? Or why is this an array?

src/js/experimental/media-rendition-listbox.js
```diff
@@ -0,0 +1,155 @@
...
+      container.append(this.#autoOption);
+    }
+
+    let isAuto = !this.mediaRenditionSelected;
+    if (isAuto) {
+      this.#autoOption.setAttribute('aria-selected', 'true');
+      this.#autoOption.setAttribute('tabindex', '0');
+    } else {
+      this.#autoOption.setAttribute('aria-selected', 'false');
+      this.#autoOption.setAttribute('tabindex', '-1');
```
heff, line 113: Why is auto getting a -1 tabindex here? I'm assuming we still want auto tab selectable? Probably worth a comment.

src/js/experimental/media-rendition-listbox.js
```diff
@@ -0,0 +1,155 @@
...
+
+    for (const rendition of renditionList) {
+
+      /** @type {HTMLOptionElement} */
+      const option = document.createElement('media-chrome-option');
+      option.part.add('option');
+      option.value = `${rendition.id}`;
+      option.textContent = `${Math.min(rendition.width, rendition.height)}p`;
+
+      if (rendition.enabled && !isAuto) {
```
heff, line 124: Enabled's not a thing anymore, right?

src/js/media-container.js
```diff
@@ -248,6 +248,17 @@ function getBreakpoints(breakpoints, rect) {
  * @cssprop --media-slot-display - `display` of the media slot (default none for [audio] usage).
  */
 class MediaContainer extends globalThis.HTMLElement {
+  static get observedAttributes() {
+    return [Attributes.AUTOHIDE, Attributes.GESTURES_DISABLED]
+      .concat(MEDIA_UI_ATTRIBUTE_NAMES)
+      // Filter out specific / complex data media UI attributes
+      // that shouldn't be propagated to this state receiver element.
```
heff, line 255: Does it break somehow? Is the list wrong if these aren't actually attributes?

heff (commented):
Looking a lot better and pretty close! The main design question is about the selected rendition being a list rather than just the index value.

src/js/controller.js
luwes: I can see that. 
let's add `loadstart` as well

src/js/controller.js
luwes: after implementing this I remembered why I'd prefer also the complex object here. 
because often (and I would like to use) the quality value to style the rendition / quality button. 
see screenshot

SCR-20230725-mod


by adding the width and height of the selected rendition we can add HD, 4K labels or icons easily to the button with CSS...

src/js/experimental/media-rendition-listbox.js
luwes: I removed this from this PR. I'll address it in the other one.
I think it is needed once you select an option, it's how it's done.
Set the selected option to tabindex=0 and the rest of the options to tabindex=-1

src/js/experimental/media-rendition-listbox.js
luwes: good catch! this is selected

src/js/media-container.js
luwes: these are filtered out because they can get pretty big.
but no, it would not break anything.
they just take a lot of HTML space so to speak.

src/js/controller.js
heff: To the button itself, not the options? What would that look like exactly? And can the CSS be used reliably with varying widths and heights?

luwes: good points, I also just reminded it would work better with also having the quality that is actually playing.

something like this

SCR-20230725-qod

luwes: going with rendition index only the `mediarenditionselected` attribute

src/js/media-container.js
luwes: only the attribute `MediaUIAttributes.MEDIA_RENDITION_LIST` is filtered out now.
it's set directly via a JS property w/ fallback to attribute

heff (approved):
Looks good!

Issue #678, opened 2023-06-30
Media seek forward button has wrong aria-label applied


mskorsur:
Hi everyone, 

I've just stumbled across this issue when I tried to run my test suite against version 1.1.2 of media-chrome. Media seek forward button has the wrong aria-label applied to it that says "seek back N seconds", the same one as the media seek backward button. 

From what I was able to figure out from the code is that in utils.js file there's only seek back label used and there's no check for whether it's forward or backward button. 

The issue can be observed when inspecting the forward button on the [homepage example video](https://www.media-chrome.org/)

Thanks for looking into this!

heff:
Thanks for pointing that out! The [label is in there](hhttps://github.com/muxinc/media-chrome/blob/93ff3cba432978eceb4259d5013673dda4c16c31/src/js/labels/labels.js#L27) but apparently it's not being used correctly. Could be an easy contribution if you want to give it a shot!

mskorsur:
Thanks for looking into this! Unfortunately, I don't have that much time to get familiar with your codebase and contribution guidelines, so I'll wait for the fix from your side. Much appreciated!

PR #685, opened 2023-07-11
fix: Seek forward aria label said "back"

heff:
fixes: #678

The seek util abstractions felt a little unnecessary and seem to have led to this bug so I backed them out in this, and also moved aria updating to one place, the seekOffset setter function.

The setNumericAttr checks the existing value to prevent infinite loops, which made it possible to keep this simple.


AdamJaggard (approved):
Looks good

luwes (approved):

PR #650, opened 2023-05-26
chore: hybrid package

gkatsev:

PR #687, opened 2023-07-15
fix: media-chrome/dist/react import

luwes:
related #680

allows to use `media-chrome/dist/react` as import 

this issue came up after adding support for CJS in https://github.com/muxinc/media-chrome/pull/650

Typescript types seem to still be broken for the new import path `media-chrome/react`
https://github.com/muxinc/media-chrome/pull/672

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,11 @@
       "require": "./dist/cjs/react/index.js",
       "default": "./dist/react/index.js"
     },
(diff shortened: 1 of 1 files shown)


cjpillsbury (approved):
If this is a breaking change, make sure you update to `feat:` or manually target a minor version release.

Issue #689, opened 2023-07-17
Volume & Time Range height


evoactivity:
<img width="339" alt="image" src="https://github.com/muxinc/media-chrome/assets/83799/6fb67637-257f-4eb3-bddd-331c743cf006">

<img width="465" alt="image" src="https://github.com/muxinc/media-chrome/assets/83799/10e7b00a-e5a7-450f-816c-305aaf174c73">

Seems volume and time range height is explicitly set where the other elements auto fill available space.

Is there a reason these elements have explicit height?
https://github.com/muxinc/media-chrome/blob/e86f14f06d7c1c7eb674159826c464f91be558de/src/js/media-control-bar.js#L33

heff:
@cjpillsbury looks like you might have been the last to touch this line. Any quick thoughts?

cjpillsbury:
Yeah it's a known shortcoming. The tl;dr:
1. Unlike `media-chrome-button`, which is entirely "DIY" and doesn't rely on a button "under the hood," `media-chrome-range` uses a descendant `input type="range"` for its core UI
2. In order to support click/touch interaction anywhere on the component, including the "padded areas", we have to size the inner input element to the full size of the component (especially relevant for mobile/touch device use)

@luwes is hoping to eventually refactor our implementation to only rely on `input type="range"` for the "state management" bits, but we haven't yet had time.

@evoactivity Since it looks like you're using the controls in a `media-control-bar`, I recently made it so you can at least scale reliably using our `--media-control-height` CSS var that will scale all controls consistently. For a simple example, see:
https://codesandbox.io/s/media-chrome-bottom-controls-forked-726lqc?file=/index.html

evoactivity:
The first screenshot in my original post is what happens out of the box, just by adding the live button, so merging #690 will fix that.

I've tried removing this height (and the one in media-chrome-range.js) so it fills the space automatically and made the control bar very large so there was a lot of empty space to tap. I could still control the volume slider from anywhere in the area on desktop and mobile.

luwes:
Thanks for reporting! This is resolved in https://github.com/muxinc/media-chrome/releases/tag/v1.5.0

PR #690, opened 2023-07-17
Live indicator svg only as tall as required

evoactivity:
The live svg was causing the live button to be ever so slightly taller than required and exposed #689

### DIFF
--- a/src/js/media-live-button.js
+++ b/src/js/media-live-button.js
@@ -8,7 +8,7 @@ const { MEDIA_TIME_IS_LIVE, MEDIA_PAUSED } = MediaUIAttributes;
 const { MEDIA_SEEK_TO_LIVE_REQUEST, MEDIA_PLAY_REQUEST } = MediaUIEvents;
 
 const indicatorSVG =
-  '<svg viewBox="0 0 6 12"><circle cx="3" cy="6" r="2"></circle></svg>';
+  '<svg viewBox="0 0 4 4"><circle cx="2" cy="2" r="2"></circle></svg>';
 
 const slotTemplate = document.createElement('template');
 slotTemplate.innerHTML = /*html*/`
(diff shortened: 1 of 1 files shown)


src/js/media-live-button.js
```diff
@@ -8,7 +8,7 @@ const { MEDIA_TIME_IS_LIVE, MEDIA_PAUSED } = MediaUIAttributes;
 const { MEDIA_SEEK_TO_LIVE_REQUEST, MEDIA_PLAY_REQUEST } = MediaUIEvents;
 
 const indicatorSVG =
-  '<svg viewBox="0 0 6 12"><circle cx="3" cy="6" r="2"></circle></svg>';
+  '<svg viewBox="0 0 4 4" style="width:6px; height:6px;"><circle cx="2" cy="2" r="2"></circle></svg>';
```
luwes, line 11: can this be done without the style attribute? 
might as well change everything in SVG

luwes:
@evoactivity do you have an example online?

I'm not seeing that the media-live-button is making the controlbar higher than it should be
https://github.com/muxinc/media-chrome/pull/700

evoactivity:
@luwes I found what is triggering this for me.

I have this css so the button only shows when on a live stream

```css
media-live-button {
  display: none;
}

media-controller[mediastreamtype='live'] media-live-button {
  display: block;
}
```

If I use `display: flex;` or `dispay: inline-flex;` to show it, the issue is not apparrent, so that is what I'll do for now.

Right now the spacing between the indicator and text is done with an `&nbsp;` spacer slot (and a little padding in the svg). My preference would be to add another css variable and use `gap: var(--media-live-button-gap, 6px);` instead of relying on a non breaking space.

luwes:
the main thing we wanted to prevent was have the live label wrap to the next line I believe.

it's possible with CSS but it would add more code. `&nbsp;` seemed a pretty simple and flexible solution.
you could also swap it out with a normal space if you like it to be wrapped in some cases.

evoactivity:
The element is a flex container though, so wrapping would only be an issue if it was set to block, and setting it to block right now causes the height issue.

I'll draft something up and open it up for discussion.

luwes:
thanks @evoactivity! yes, that's a good idea to move it to a discussion.
closing this PR.

PR #692, opened 2023-07-18
fix: Attr setting error in seek buttons

### LINKED ISSUE #691: Issue using seek forward and backward buttons
Hello,
seems there's an issue introduced with this [PR](https://github.com/muxinc/media-chrome/pull/685). When I try
to use seek forward and backward buttons I get the following error: `Uncaught DOMException: Failed to construct 'CustomElement': The result must not have attributes.`

From what I've been able [to investigate](https://stackoverflow.com/a/43837330), this occurs when modifying attributes in the constructor of the element, and that should be avoided. This happens with these buttons, since they set `aria-label` property in [their constructor](https://github.com/muxinc/media-chrome/blob/e86f14f06d7c1c7eb674159826c464f91be558de/src/js/media-seek-forward-button.js#L42). This should be rewritten to use connectedCallback() possibly as suggested in the stack overflow response.

Thanks for looking into this!

heff:
fixes #691

Call the attr setting in connectedCallback rather than constructor.

### DIFF
--- a/src/js/media-seek-backward-button.js
+++ b/src/js/media-seek-backward-button.js
@@ -39,7 +39,11 @@ class MediaSeekBackwardButton extends MediaChromeButton {
 
   constructor(options = {}) {
     super({ slotTemplate, ...options });
+  }
+
+  connectedCallback(){
     this.seekOffset = getNumericAttr(this, Attributes.SEEK_OFFSET, DEFAULT_SEEK_OFFSET);
+    super.connectedCallback();
   }
 
   attributeChangedCallback(attrName, _oldValue, newValue) {
--- a/src/js/media-seek-forward-button.js
+++ b/src/js/media-seek-forward-button.js
@@ -39,7 +39,11 @@ class MediaSeekForwardButton extends MediaChromeButton {
 
   constructor(options = {}) {
     super({ slotTemplate, ...options });
+  }
+
+  connectedCallback(){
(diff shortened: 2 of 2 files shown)


cjpillsbury (approved):
🏄‍♂️

src/js/media-seek-backward-button.js
```diff
@@ -39,7 +39,11 @@ class MediaSeekBackwardButton extends MediaChromeButton {
 
   constructor(options = {}) {
     super({ slotTemplate, ...options });
+  }
+
+  connectedCallback(){
     this.seekOffset = getNumericAttr(this, Attributes.SEEK_OFFSET, DEFAULT_SEEK_OFFSET);
+    super.connectedCallback();
```
luwes, line 46: it depends on use case but I think generally the super calls go before the code in the extending class.

`this.seekOffset` could potentially be changed in `super.connectedCallback()`

src/js/media-seek-backward-button.js
```diff
@@ -39,7 +39,11 @@ class MediaSeekBackwardButton extends MediaChromeButton {
 
   constructor(options = {}) {
     super({ slotTemplate, ...options });
+  }
+
+  connectedCallback(){
```
luwes, line 44: nit (non-b): a space after is generally used in the code base

```suggestion
  connectedCallback() {
```

luwes (approved):

PR #696, opened 2023-07-21
Fixes `<media-live-button/>`

evoactivity:
`setAttribute` was accidentally used in the constructor. I've moved the logic to a function similarly to other elements.

### DIFF
--- a/src/js/media-live-button.js
+++ b/src/js/media-live-button.js
@@ -45,11 +45,11 @@ slotTemplate.innerHTML = /*html*/`
 `;
 
 const updateAriaAttributes = (el) => {
-  const isPaused = el.mediaPaused || !el.mediaTimeIsLive
-  const label = isPaused ? verbs.SEEK_LIVE() : verbs.PLAYING_LIVE();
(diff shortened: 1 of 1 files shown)


src/js/media-live-button.js
```diff
@@ -44,6 +44,16 @@ slotTemplate.innerHTML = /*html*/`
   <slot name="spacer">&nbsp;</slot><slot name="text">LIVE</slot>
 `;
 
+const updateAriaAttributes = (el) => {
+  const isPaused = el.mediaPaused || !el.mediaTimeIsLive
```
heff, line 48: `isPaused` isn't the best name to represent `!el.mediaTimeIsLive` since the video is not paused in that case. `isNotLive` might be clearer. `isPausedOrNotLive` would be most clear and accurate probably.

heff (approved):
Awesome, thanks! One small comment, but approved. If you don't address the comment I'll merge in a few days.

src/js/media-live-button.js
evoactivity: I've updated the variable name :)

heff (approved):

PR #703, opened 2023-07-26
chore: Define the type of `_inactiveTimeout` so it is not inferred as `NodeJS.Timeout`

evoactivity:
`_inactiveTimeout` was incorrectly typed to `NodeJS.Timeout` which made my linter unhappy.

Either `number` or `ReturnType<typeof setTimeout>` would work but the latter feels more explicit.

### DIFF
--- a/src/js/media-container.js
+++ b/src/js/media-container.js
@@ -486,6 +486,7 @@ class MediaContainer extends globalThis.HTMLElement {
       // Setting autohide to -1 turns off autohide
       if (this.autohide < 0) return;
 
+      /** @type {ReturnType<typeof setTimeout>} */
       this._inactiveTimeout = setTimeout(() => {
         setInactive();
       }, this.autohide * 1000);


src/js/media-container.js
```diff
@@ -486,6 +486,7 @@ class MediaContainer extends globalThis.HTMLElement {
       // Setting autohide to -1 turns off autohide
       if (this.autohide < 0) return;
 
+      /** @type {ReturnType<typeof setTimeout>} */
```
luwes, line 489: I think we'd prefer to keep the code less noisy if possible.

isn't it the same signature? on my end TS is not erroring / warning.
yarn build seems to pass as well

evoactivity: In my project I have a setup where my templates are type aware, so the types for media-chrome end up being loaded, but I do not have `@types/node` installed so typescript can't find the `NodeJS` namespace and gives me a lint error.

image

image

The signatures are different, in node you get the instance of the Timeout class, in the browser it's just a number.

evoactivity: The test runner is bringing in @types/node so you won't see this error when running `tsc`. This will also be the reason it gets typed as `NodeJS.Timeout` to begin with.

image

evoactivity: Another option instead of a type comment would be to use `window.setTimeout` as that would indicate to typescript it should use the DOM typings, not Node. But I haven't tested that, and it would probably need a comment anyway to communicate why only this instance of `setTimeout` is using `window` so it's not removed by someone in the future.

luwes (approved):
👍

Issue #662, opened 2023-06-05
Time range drag doesn't move current time on iOS


flipnow:
Steps to reproduce:

1. Open https://4rv092.csb.app/ on iOS safari
2. Try to drag time range but not by input thumb

Actual result:

Dragging only changes preview time but not current time.

Expected result:

I am not entirely sure if it's expected or maybe just a quirk of iOS system but it's definitely not consistent with other platforms. On any desktop browser or even Android Chrome dragging on time range (not by thumb indicator) does move current time to correct position.

Why is this an issue:

On smaller iPhone screens if user doesn't hit precisely the input thumb drag will have no effect which can cause a lot frustration.

heff:
> dragging on time range (not by thumb indicator) does move current time to correct position

I think there's more of bug here. When you tap on the bar (not the handle) it seeks to that time, but touchstart+touchmove (drag without the handle) does not update the time. That's not right so labeling as a bug. Some digging is needed to understand why the touch events aren't working as expected here.

luwes:
Thanks for the feedback! This is resolved in https://github.com/muxinc/media-chrome/releases/tag/v1.5.0

PR #704, opened 2023-07-26
feat: separate slider UI from input range + improve mobile

luwes:
related discussion: https://github.com/muxinc/media-chrome/discussions/421

- [x] fix jitteriness of progress bar on video startup (especially noticeable in Safari)
- [x] fix not hiding controls while dragging on mobile #751
- [x] add touch move handler for mobile enabling dragging #662

--- 

this change removes the variable `min` and `max` values from the range and sets them to a fixed `0` and `1`.
it simplifies the code as you can see in this PR.

this change doesn't use the UI from input[type=range] anymore but the range is still used as state object, event listener.
this will allow new features and more flexibility, some examples:

- less have to worry about browser inconsistencies in styling the range
- a gradient as background color of the progress bar
- using transitions on the progress bar width and left positioning of the thumb
- it does away with the annoying calculating issue where you have to incorporate the thumb width of the range

### DIFF
--- a/src/js/media-chrome-range.js
+++ b/src/js/media-chrome-range.js
@@ -6,8 +6,6 @@ const template = document.createElement('template');
 template.innerHTML = /*html*/`
   <style>
     :host {
-      --thumb-height: var(--media-range-thumb-height, 10px);
-      --track-height: var(--media-range-track-height, 4px);
       --_focus-box-shadow: var(--media-focus-box-shadow, inset 0 0 0 2px rgb(27 127 204 / .9));
       --_media-range-padding: var(--media-range-padding, var(--media-control-padding, 10px));
 
@@ -47,30 +45,34 @@ template.innerHTML = /*html*/`
       min-height: min(100%, calc(var(--media-control-height, 24px) + 2 * var(--_media-range-padding)));
     }
 
-    input[type=range] {
-      ${/* Reset */''}
+    #range {
       -webkit-appearance: none; ${/* Hides the slider so that custom slider can be made */''}
+      -webkit-tap-highlight-color: transparent;
       background: transparent; ${/* Otherwise white in Chrome */''}
-
-      ${/* Fill host with the range */''}
-      min-height: 100%;
-      width: var(--media-range-track-width, 100%); ${/* Specific width is required for Firefox. */''}
-
       box-sizing: border-box;
-      padding: 0;
       margin: 0;
+      padding: 0;
+      ${/* The input range acts as a hover and hit zone for input events. */''}
+      width: var(--media-range-track-width, 100%); ${/* Firefox requires specific width. */''}
+      transform: translate(var(--media-range-track-translate-x, 0px), calc(var(--media-range-track-translate-y, 0px)));
+      display: var(--media-time-range-hover-display, block);
+      bottom: var(--media-time-range-hover-bottom, -5px);
+      height: var(--media-time-range-hover-height, max(calc(100% + 5px), 20px));
+      position: absolute;
+      cursor: pointer;
+      z-index: 1; ${/* Apply z-index to overlap buttons below. */''}
     }
 
     ${/* Special styling for WebKit/Blink */''}
     ${/* Make thumb width/height small so it has no effect on range click position. */''}
-    input[type=range]::-webkit-slider-thumb {
+    #range::-webkit-slider-thumb {
       -webkit-appearance: none;
       width: .1px;
       height: .1px;
     }
 
     ${/* The thumb is not positioned relative to the track in Firefox */''}
-    input[type=range]::-moz-range-thumb {
+    #range::-moz-range-thumb {
       background: transparent;
       border: transparent;
       width: .1px;
@@ -80,7 +82,7 @@ template.innerHTML = /*html*/`
     #background,
     #track {
       width: var(--media-range-track-width, 100%);
-      height: var(--track-height);
+      height: var(--media-range-track-height, 4px);
       border-radius: var(--media-range-track-border-radius, 1px);
       transform: translate(var(--media-range-track-translate-x, 0px), calc(var(--media-range-track-translate-y, 0px)));
       position: absolute;
@@ -130,8 +132,8 @@ template.innerHTML = /*html*/`
     }
 
     #thumb {
-      height: var(--thumb-height);
       width: var(--media-range-thumb-width, 10px);
+      height: var(--media-range-thumb-height, 10px);
       margin-left: calc(var(--media-range-thumb-width, 10px) / -2);
       border: var(--media-range-thumb-border, none);
       border-radius: var(--media-range-thumb-border-radius, 10px);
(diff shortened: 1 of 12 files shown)


src/js/media-time-range.js
```diff
@@ -535,7 +540,7 @@ class MediaTimeRange extends MediaChromeRange {
 
     // Get mouse position percent
     const rangeRect = this.range.getBoundingClientRect();
-    let mouseRatio = (evt.clientX - rangeRect.left - this.thumbWidth / 2) / (rangeRect.width - this.thumbWidth);
+    let mouseRatio = (evt.clientX - rangeRect.left) / rangeRect.width;
```
cjpillsbury, line 544: praise: I originally had a question here since this was an annoying edge case we had to account for... but instead, with your new bespoke UI, it's another improvement! 🎉

src/js/media-time-range.js
```diff
@@ -443,58 +461,45 @@ class MediaTimeRange extends MediaChromeRange {
...
         ([start, end]) => start <= currentTime && currentTime <= end
       ) ?? [];
-      relativeBufferedEnd = bufferedEnd - range.min;
+      relativeBufferedEnd = calcRangeValueFromTime(this, bufferedEnd);
     } else {
       // If we've ended, there may be some discrepancies between seekable end, duration, and current time.
       // In this case, just presume `relativeBufferedEnd` is the maximum possible value for visualization
       // purposes (CJP.)
-      relativeBufferedEnd = relativeMax;
+      relativeBufferedEnd = 1;
```
cjpillsbury, line 489: suggestion(non-blocking): I believe we can simplify this (and account for some other cases as well) if we update `calcRangeValueFromTime()` to do a `Math.min(returnValue, 1)` (and probably could do a similar cap for `Math.max(returnValue, 0)` to account for edge cases, though maybe less of a concern?)

src/js/media-time-range.js
```diff
@@ -443,58 +461,45 @@ class MediaTimeRange extends MediaChromeRange {
...
     // Find the buffered range that "contains" the current time and get its end.
-    // If none, just assume the start of the media timeline/range.min for
+    // If none, just assume the start of the media timeline for
     // visualization purposes.
     let relativeBufferedEnd;
+
     if (!this.mediaEnded) {
       const currentTime = this.mediaCurrentTime;
-      const [, bufferedEnd = range.min] = buffered.find(
+      const [, bufferedEnd = 0] = buffered.find(
```
cjpillsbury, line 481: question(blocking): Shouldn't this be `bufferedEnd = this.mediaSeekableStart]`?

cjpillsbury (changes requested):
One small blocking question and one suggestion (plus out of band convo). Otherwise, I'm pumped about this! I didn't do a fine toothed comb review of all of the CSS variables, but make sure those are consistent or call them out if not.

cjpillsbury:
oh one last thing: I haven't reasoned through everything here, but what happens to the new bespoke UI if/when we resize the time range?

src/js/media-time-range.js
luwes: I think you're right. will update!

src/js/media-time-range.js
luwes: updated!

src/js/media-container.js
```diff
@@ -542,10 +542,6 @@ class MediaContainer extends globalThis.HTMLElement {
     });
 
     this.addEventListener('pointermove', (e) => {
-      // pointermove doesn't happen with touch on taps on iOS, but does on android,
-      // so, only run pointermove for mouse
-      if (e.pointerType !== 'mouse') return;
```
luwes, line 547: this change keeps the controls showing while touch dragging

hivokas:
@luwes thanks for working on this PR! Will it fix this issue https://github.com/muxinc/elements/issues/798?

src/js/media-time-range.js
```diff
@@ -63,14 +68,22 @@ template.innerHTML = /*html*/`
...
-    :host([${MediaUIAttributes.MEDIA_PREVIEW_IMAGE}]:hover) [part~="preview-box"],
-    :host([${MediaUIAttributes.MEDIA_PREVIEW_TIME}]:hover) [part~="preview-box"] {
+    :host(:is([${MediaUIAttributes.MEDIA_PREVIEW_IMAGE}], [${MediaUIAttributes.MEDIA_PREVIEW_TIME}])[dragging]) [part~="preview-box"] {
       transition-duration: var(--media-preview-transition-duration-in, .5s);
       transition-delay: var(--media-preview-transition-delay-in, .25s);
       visibility: visible;
       opacity: 1;
     }
 
+    @media (hover: hover) {
```
luwes, line 78: on devices that don't support hover, don't add the :hover selector.
:hover gets enabled by tap on mobile and is sticky. use the explicit [dragging] attribute...

src/js/media-chrome-range.js
```diff
@@ -71,74 +39,115 @@ template.innerHTML = /*html*/`
...
+      ${/* The input range acts as a hover and hit zone for input events. */''}
+      display: var(--media-time-range-hover-display, block);
+      bottom: var(--media-time-range-hover-bottom, -7px);
+      height: var(--media-time-range-hover-height, max(100% + 7px, 25px));
+      width: 100%;
+      position: absolute;
+      cursor: pointer;
+
       -webkit-appearance: none; ${/* Hides the slider so that custom slider can be made */''}
+      -webkit-tap-highlight-color: transparent;
```
luwes, line 67: this removes a background color on touch, looks much better.

src/js/media-chrome-range.js
```diff
@@ -71,74 +39,115 @@ template.innerHTML = /*html*/`
...
-      padding: 0;
       margin: 0;
+      z-index: 1;
+    }
+
+    @media (hover: hover) {
+      #range {
+        bottom: var(--media-time-range-hover-bottom, -5px);
+        height: var(--media-time-range-hover-height, max(100% + 5px, 20px));
+      }
```
luwes, line 77: on desktop the hit area can be smaller

cjpillsbury (approved):
👍

hivokas:
@luwes thank you for your work! It works much better on iOS now!

JordanKatz:
@luwes this bug fix does not pass QA. 

The drag and drop UI works better, but the player doesnt obey the command to change the media when user lets go of the slider

iPhone video recording
https://drive.google.com/file/d/1jiqzPUudxErcqx2ZTc3qts0ZaxxvgVG4/view?usp=drive_link

actual link 
https://resueno.bushidoco.de/media/murray-kyle-remix-contest-protect-the-water

this is very painful as our investors caught the bug hurting our ability to close money. 

please investigate and let us know. 

thanks

luwes:
@JordanKatz this is not released yet in Mux player. 
 you make the Google drive link public please? thanks

JordanKatz:
@luwes updated link. thx.

PR #705, opened 2023-07-28
fix: select keyboard focus

luwes:
when opening the select list keyboard focus wouldn't jump from button to list.
removing the fade in duration of visibility fixed this.


AdamJaggard (approved):
good catch

PR #706, opened 2023-07-28
fix: -:-- time codes on player load to 0:00

luwes:
this was recently introduced with the NaN return value in getNumericAttr I think.

it looks pretty bad sometimes https://media-chrome.mux.dev/examples/vanilla/mobile.html

### DIFF
--- a/src/js/utils/time.js
+++ b/src/js/utils/time.js
@@ -80,7 +80,7 @@ export function formatTime(seconds, guide) {
   if (isNaN(seconds) || seconds === Infinity) {
     // '-' is false for all relational operators (e.g. <, >=) so this setting
     // will add the minimum number of fields specified by the guide
-    h = m = s = '-';
+    h = m = s = '0';
   }
 
   // Check if we need to show hours


heff (approved):
This is how it's always been as far as I know. I'm not sure if something upstream is making it show up as a blank time more now. I did dashes because it felt a little wrong to have zeros for duration before it's known. But I'm fine with going this direction also. 👍

luwes:
double checked because I swear it was zeros before 😅 

@cjpillsbury 's PR introduced it 
https://github.com/muxinc/media-chrome/pull/665
https://media-chrome-git-fork-cjpillsbury-feat-props-state-p-389bb5-mux.vercel.app/examples/vanilla/mobile.html

The PR right before looked like this with the 0:00
https://media-chrome-git-fork-luwes-media-icon-mux.vercel.app/examples/vanilla/mobile.html

it's true that the utility function always made NaN or Infinity a dash
I'll do a quick check what really caused this

luwes:
yup fairly sure it's caused by the NaN default value 

```js
export function getNumericAttr(el, attrName, defaultValue = Number.NaN) {
  const attrVal = el.getAttribute(attrName);
  return attrVal != null ? +attrVal : defaultValue;
}
```

PR #707, opened 2023-07-28
fix: set breakpoints on connected 

### LINKED ISSUE #697: set breakpoint attributes quicker again
https://github.com/muxinc/media-chrome/pull/682 added a delay to setting these attributes

in themes these attributes are used to trigger render and Mux player uses the breakpoints to conditionally render size based views.
as a result this delay adds a delay to render the view in Mux player and this is noticeable on player load.

some examples: 

- https://elements-demo-nextjs.vercel.app/MuxPlayer
- https://dev.playerx.io/demo/mux/

try refreshing a couple of times and you can see the jumping of control bar

luwes:
fix #697


cjpillsbury (approved):
🎉

PR #711, opened 2023-08-03
fix: Move all breakpoint computation to resizeObserver. defer quickly…

cjpillsbury:
… (0ms) for initial computation.

Manual testing on mux-player showed that first render of theme had expected breakpoints based on player size. Locally confirmed this still resolves known resize error scenario.

### DIFF
--- a/src/js/media-container.js
+++ b/src/js/media-container.js
@@ -332,7 +332,7 @@ class MediaContainer extends globalThis.HTMLElement {
         resizeCallback(entries);
         // Once we've completed, reset the pending cb flag to false
         pendingResizeCb = false;
-      }, 16); // Use 16ms to ensure at most ~1 invocation per animation frame (~60fps)
+      }, 0);
       pendingResizeCb = true;
     };
     const resizeObserver = new ResizeObserver(deferResizeCallback);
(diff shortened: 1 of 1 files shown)


src/js/media-container.js
```diff
@@ -332,8 +334,12 @@ class MediaContainer extends globalThis.HTMLElement {
         resizeCallback(entries);
         // Once we've completed, reset the pending cb flag to false
         pendingResizeCb = false;
-      }, 16); // Use 16ms to ensure at most ~1 invocation per animation frame (~60fps)
+      }, delayMS);
       pendingResizeCb = true;
+      // After first resize, delay 16ms to ensure at most ~1 invocation per animation frame (~60fps)
```
luwes, line 339: this comment is not correct as far as I understand it. 
it's more like the browser couldn't do all the observations in 1 animation frame because too much stuff changed.

Logrocket's comment is incorrect:

> The first thing you may notice is that we’ve wrapped everything in a call to requestAnimationFrame. This simply throttles how often our code can run (typically 60 times per second). This helps avoid ResizeObserver loop limit exceeded console warnings, which can happen whenever your observer callback tries to run multiple times during a single animation frame.

This one from Stackoverflow is more how it is:
> This error means that ResizeObserver was not able to deliver all observations within a single animation frame.

Here is the explanation from a Webkit:
https://webkit.org/blog/9997/resizeobserver-in-webkit/#:~:text=observation.-,How%20Observations%20are%20Delivered

---

this translates roughly to:

because a large chain of tasks (causing layouts etc) is initiated resulting from setting the breakpoint attribute defer setting the breakpoint attribute to not show the error.

technically having this timeout delay be 0s all the time should also hide the warning.
could you try that out?

luwes (approved):

PR #716, opened 2023-08-10
Listbox fixes

luwes:
- adds a header slot to listbox + example
  https://media-chrome-git-fork-luwes-listbox-fixes-mux.vercel.app/examples/vanilla/control-elements/media-rendition-selectmenu.html
- adds a border-radius CSS var for listbox
- adds default font and color styles to the selectlist
- moves the `Auto` option to the bottom
- fixes an issue using the media-chrome-listbox as standalone 
  https://media-chrome-git-fork-luwes-listbox-fixes-mux.vercel.app/examples/vanilla/control-elements/media-chrome-listbox.html

### DIFF
--- a/examples/vanilla/control-elements/media-rendition-selectmenu.html
+++ b/examples/vanilla/control-elements/media-rendition-selectmenu.html
@@ -20,6 +20,18 @@
       display: none;
     }
 
+    :not(:defined) .rendition-title {
+      display: none;
+    }
+
+    .rendition-title {
+      background: rgb(20 20 30 / .8);
+      border-bottom: 1px solid rgb(255 255 255 / .38);
+      border-radius: 2px 2px 0 0;
+      text-align: center;
+      padding-block: .4em;
+    }
+
     video {
       width: 100%;      /* prevents video to expand beyond its container */
     }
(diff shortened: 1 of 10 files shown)


examples/vanilla/control-elements/media-rendition-selectmenu.html
```diff
@@ -7,23 +7,28 @@
   <script type="module" src="../../../dist/index.js"></script>
   <script type="module" src="../../../../dist/experimental/media-rendition-selectmenu.js"></script>
   <style>
+    /* Hide custom elements that are not defined yet */
+    :not(:defined) {
```
cjpillsbury, line 11: praise: 😎

src/js/experimental/media-chrome-listbox.js
```diff
@@ -136,6 +157,8 @@ class MediaChromeListbox extends globalThis.HTMLElement {
     if (!indicator)
       indicator = this.shadowRoot.querySelector(`[name="${name}-indicator"] > svg`);
 
+    // @ts-ignore
+    indicator = indicator.cloneNode(true);
```
cjpillsbury, line 161: question(non-blocking): 
it looks like we mean two things by `indicator` here:
1. a "reference element" for the indicator (is this guaranteed to always exist? looks like yes, but just confirming)
2. the instance that will be used for a given option, which is a clone of (1) and gets some updates based on its usage.

We may want to either rename this method to clarify or split this into two different methods/functions for clarity.

src/js/experimental/media-chrome-selectmenu.js
```diff
@@ -8,6 +8,12 @@ const template = document.createElement('template');
 template.innerHTML = /*html*/`
   <style>
   :host {
+    font: var(--media-font,
```
cjpillsbury, line 11: praise: 👌

cjpillsbury (approved):
one nb question, otherwise 😎 👍

src/js/experimental/media-chrome-listbox.js
luwes: good call, removing the slot attribute was causing issues if it was done on the original element.
I'll improve this

PR #719, opened 2023-08-14
fix: monitor if breakpoints have been computed and only trigger theme…

cjpillsbury:
… update if they have.

### DIFF
--- a/src/js/media-theme-element.js
+++ b/src/js/media-theme-element.js
@@ -66,8 +66,11 @@ export class MediaThemeElement extends globalThis.HTMLElement {
     }
 
     const observer = new MutationObserver((mutationList) => {
+      // Only update if `<media-controller>` has computed breakpoints at least once.
+      if (this.mediaController?.breakpointsUncomputed) return;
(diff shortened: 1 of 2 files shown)


src/js/media-theme-element.js
```diff
@@ -67,14 +67,17 @@ export class MediaThemeElement extends globalThis.HTMLElement {
...
+        const target = /** @type {HTMLElement & { computedBreakpoints?: boolean }} */ (mutation.target);
 
         // Render on each attribute change of the `<media-theme(-x)>` element.
         if (target === this) return true;
 
         // Only check `<media-controller>`'s attributes below.
         if (target.localName !== 'media-controller') return false;
 
+        // Only update if `<media-controller>` has computed breakpoints at least once.
+        if (!target.computedBreakpoints) return false;
```
luwes, line 79: should this go above the first `if (target === this) return true;`?

luwes (approved):

luwes (approved):
👍

PR #722, opened 2023-08-16
fix: media-chrome-option outline fix and new css vars

AdamJaggard:
Adds new CSS vars for controlling hover and non hover outlines. This fixes an issue where outlines are being cut off on the left and right side of the options.

Hover outline styles are also applied to the `:focus-visible` state for consistency.

There's also a new `--media-option-focus-shadow` var which is helpful for disabling the shadow on `:focus-visible` for simply allowing the option to look the same on hover and focus relying only on the outline vars.

### DIFF
--- a/src/js/experimental/media-chrome-option.js
+++ b/src/js/experimental/media-chrome-option.js
@@ -1,7 +1,7 @@
 import { globalThis, document } from '../utils/server-safe-globals.js';
 
 const template = document.createElement('template');
-template.innerHTML = /*html*/`
+template.innerHTML = /*html*/ `
 <style>
   :host {
     cursor: pointer;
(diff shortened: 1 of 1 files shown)


src/js/experimental/media-chrome-option.js
```diff
@@ -13,17 +13,21 @@ template.innerHTML = /*html*/`
...
+    outline: var(--media-option-outline, 0);
+    outline-offset: var(--media-option-outline-offset, -1px);
   }
 
   :host(:focus-visible) {
-    box-shadow: inset 0 0 0 2px rgb(27 127 204 / .9);
-    outline: 0;
+    box-shadow: var(--media-option-focus-shadow, inset 0 0 0 2px rgb(27 127 204 / .9));
+    outline: var(--media-option-hover-outline, 0);
+    outline-offset: var(--media-option-hover-outline-offset, -1px);
```
luwes, line 23: can this be without an explicit -1px as default? 

like can it get it from `:host` by default?

AdamJaggard: do you mean add another var higher up the chain that it inherits from? (that itself is defaulted to -1px?)

luwes: yea I was thinking it should get the default automatically from 

```css
:host { outline-offset: var(--media-option-outline-offset, -1px); }
```

not sure if it needs an explicit `revert` or `inherit` or so 

could also do something like if it's not easy with CSS keywords

`outline-offset: var(--media-option-hover-outline-offset, var(--media-option-outline-offset, -1px))`

AdamJaggard: `revert` and `inherit` doesn't seem to work, assumedly because it's not by default an inheritable property. But I've made the hover states use the non-hover offset property as a fallback like in your last example.

luwes (approved):
LGTM!

PR #724, opened 2023-08-18
feat: add option-selected part for selected option elements

AdamJaggard:
Allows you to select only selected options via css `::part(selected-option)` as `::part` doesn't support combining with other selectors like `[aria-selected="true"]`.


luwes (approved):
🚢

luwes (approved):

PR #730, opened 2023-08-22
fix: account for single child in react wrapper templating.

### LINKED ISSUE #727: MediaChrome React component crash if given single child
Hello. I'm building a player with Media Chrome using the react components found in `media-chrome/dist/react`. I unfortunately get an error when giving a single child to a React component. The following crashes...

```
<MediaPlayButton>
            <svg slot="play" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24">
              ...
            </svg>
</MediaPlayButton>
```
But multiple children is ok (notice extra div under svg)...
```
<MediaPlayButton>
            <svg slot="play" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24">
              ...
            </svg>
            <div />
</MediaPlayButton>
```

The error is
```
UncaughtError: undefined TypeError: Spread syntax requires ...iterable[Symbol.iterator] to be a function
```

I'm on version **1.2.5**

Thanks!

cjpillsbury:
fixes #727

### DIFF
--- a/scripts/react/build.js
+++ b/scripts/react/build.js
@@ -34,7 +34,7 @@ const toReactComponentStr = (config) => {
   const ReactComponentName = toPascalCase(elementName);
   return `/** @type { import("react").HTMLElement } */
 const ${ReactComponentName} = React.forwardRef(({ children = [], ...props }, ref) => {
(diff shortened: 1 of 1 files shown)


luwes (approved):
👍

PR #731, opened 2023-08-30
fix: Themes not rendering if no breakpoint is active on load

AdamJaggard:
This fixes an issue where themes don't render any of the theme template **_on page load_** if the player is smaller than the smallest breakpoint defined. This affects all themes.

The default mux player theme (classic) redefined the `sm` breakpoint as 300 instead of the default which makes it unlikely to appear in normal usage. The new 2023 theme defined a larger `sm` breakpoint and so the UI stopped loading on mobile device sizes.

Although this bug is producible through themes using the breakpoint variables, it's a lower level issue around how and when we decide to render a theme template. This bug exists because when no breakpoints are active, we don't set any new properties on the media controller (like `breakpointsm`), and this in turn means the mutation observer on `media-theme-element` doesn't trigger, and that's the only mechanism we have to trigger a render (nothing else calls the render function other than these mutation callbacks).

Additionally `streamtype` is undefined the first time a theme template is processed (if it's being automatically derived), and because most themes use this variable as a top level branching point to show different types of UI (live vs vod), no elements are shown when a theme is first processed. There's likely other variables that could be used to show this behaviour but it's more common to build top level theme logic around `streamtype` and breakpoint variables so they expose it.

This fix bypasses the issue by triggering a render indirectly via a new event that fires after breakpoints are initially processed, even if no breakpoint is active and no attributes will change. This doesn't fix the underlying issue though.

### DIFF
--- a/src/js/constants.js
+++ b/src/js/constants.js
@@ -80,15 +80,16 @@ export const MediaUIAttributes =
   );
 
 export const MediaStateChangeEvents =
-  /** @type {{ [k in keyof MediaUIProps | 'USER_INACTIVE' | 'BREAKPOINTS_CHANGE']: string }} */ (
+  /** @type {{ [k in keyof MediaUIProps | 'USER_INACTIVE' | 'BREAKPOINTS_CHANGE' | 'BREAKPOINTS_COMPUTED']: string }} */ (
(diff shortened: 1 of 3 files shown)


cjpillsbury (approved):
🚀

PR #734, opened 2023-09-14
fix: improve selectmenu overflow

luwes:
related resize observer performance issue:
https://github.com/WICG/resize-observer/issues/59

test url:
https://media-chrome-git-fork-luwes-selectmenu-overflow-mux.vercel.app/examples/vanilla/control-elements/media-rendition-selectmenu.html

before:

SCR-20230914-cp9

after (w/ scrolling):

SCR-20230914-h68

### DIFF
--- a/src/js/experimental/media-chrome-selectmenu.js
+++ b/src/js/experimental/media-chrome-selectmenu.js
@@ -19,13 +19,12 @@ template.innerHTML = /*html*/`
     flex-shrink: .5;
   }
 
-  [name="listbox"]::slotted(*),
+  [name=listbox]::slotted(*),
   [part=listbox] {
     position: absolute;
     left: 0;
     bottom: 100%;
     max-height: 300px;
-    overflow: hidden auto;
     transition: var(--media-selectmenu-transition-in,
       visibility 0s, transform .15s ease-out, opacity .15s ease-out);
     transform: var(--media-listbox-transform-in, translateY(0) scale(1));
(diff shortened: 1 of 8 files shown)


AdamJaggard:
If you open the new scrolling selectmenu and then resize the player (using the browser or programatically) you end up with a mismatch again. I think it just needs re-calculating on resize while it's open.

image

luwes:
good call will update this!

src/js/experimental/media-chrome-selectmenu.js
```diff
@@ -229,42 +233,31 @@ class MediaChromeSelectMenu extends globalThis.HTMLElement {
...
-    this.#button.removeAttribute('disabled');
-    this.#button.addEventListener('click', this.#handleButtonClick);
-    this.#button.addEventListener('keydown', this.#keydownListener);
-    this.#listbox.addEventListener('keydown', this.#keydownListener);
-    this.#toggleExpanded();
-    this.#listbox.addEventListener('change', this.#handleOptionChange);
+    this.#button.toggleAttribute('disabled', false);
+    this.addEventListener('change', this.#handleOptionChange);
+    this.addEventListener('keydown', this.#keydownListener);
+    this.addEventListener('click', this.#clickHandler);
```
luwes, line 251: moving to event delegation here makes it easier and safer I think. 
if the button is removed from the selectmenu, there is no click event on the button anymore

AdamJaggard (approved):
This looks good, resizing is working well.

I like the use of a single persistent resize observer instead of constantly creating new ones and destroying old ones, should make it much easier to keep track of. Less instances to accidentally hang around.

PR #743, opened 2023-10-10
feat!: use remote playback API

luwes:
BREAKING CHANGE: uses castable-video v1

Requires https://github.com/muxinc/castable-video/pull/16

https://github.com/muxinc/media-chrome/discussions/654

# IMPORTANT
Mux player requires the new `<castable-video>` with this change!
https://github.com/muxinc/castable-video


src/js/controller.js
```diff
@@ -356,67 +356,54 @@ export const MediaUIStates = {
...
     // NOTE: only adding this if airplay is supported, in part to avoid unnecessary battery consumption per
     // Apple docs recommendations (See: https://developer.apple.com/documentation/webkitjs/adding_an_airplay_button_to_your_safari_media_controls)
-    // For a more advanced solution, we could monitor for media state receivers that "care" about airplay support and add/remove
-    // whenever these are added/removed. (CJP)
-    // NOTE: I don't think only adding this if supported helps, so making this match others.
-    // If Airplay's not supported then it it doesn't hurt to add it.
-    // If battery is really an issue we need a different approach.
-    // Also this is a terrible API, Apple. (heff)
-    get: function (controller, e) {
+    get: function (controller, availability) {
```
cjpillsbury, line 372: nitpick(blocking): I think we should avoid having one-off argument "shapes" for the second argument here. Let's minimally pass in an object of the shape `{ availiability }`. This will have the added benefit of being backwards compatible with the previous, airplay-specific event-based impl

src/js/controller.js
```diff
@@ -356,67 +356,54 @@ export const MediaUIStates = {
...
       if (!airplaySupported) return AvailabilityStates.UNSUPPORTED;
```
cjpillsbury, line 373: praise: Was originally concerned that we wouldn't be able to disambiguate between chromecast vs. airplay availability states because of the generic nature of the Remote Playback API, but then I realized that these initial supported state gates here and for the chromecast equivalent account for that concern 😎

cjpillsbury (dismissed):
Haven't smoke tested this, but overall this is looking great. Have one small ask for the signature change. Also have one larger scope creep question that doesn't need to go into this PR: If we're going to use the Remote Playback API for Airplay as well as chromecast, does that mean we can finally add a `MEDIA_IS_AIRPLAYING` state (equivalent to `MEDIA_IS_CASTING`? 🤞

luwes:
yep I believe so wrt `MEDIA_IS_AIRPLAYING` 🎉 
will update from feedback soon

src/js/controller.js
luwes: updated to receive `{ availability }`

src/js/constants.js
```diff
@@ -46,6 +46,7 @@ export const MediaUIProps = {
   MEDIA_VOLUME_UNAVAILABLE: 'mediaVolumeUnavailable',
   MEDIA_IS_PIP: 'mediaIsPip',
   MEDIA_IS_CASTING: 'mediaIsCasting',
+  MEDIA_IS_AIRPLAYING: 'mediaIsAirplaying',
```
cjpillsbury, line 49: praise: 🎉

PR #753, opened 2023-10-26
fix: selectmenu anchor right + add listbox row layout

luwes:
https://media-chrome-git-fork-luwes-playbackrate-menu-mux.vercel.app/examples/vanilla/control-elements/media-playback-rate-selectmenu.html

- [x] change default rates to `1, 1.2, 1.5, 1.7, 2`
- [x] anchor selectmenu's by right of button (because generally they are on right side of player)
- [x] make playbackrate button a minimum width for default rates so there is no jump when changing 2 characters
- [x] add new `--media-listbox-layout: row` CSS var on listbox for a flex row layout + style tweaks for that layout

### DIFF
--- a/src/js/experimental/media-chrome-selectmenu.js
+++ b/src/js/experimental/media-chrome-selectmenu.js
@@ -23,7 +23,6 @@ template.innerHTML = /*html*/`
   [name=listbox]::slotted(*),
   [part=listbox] {
     position: absolute;
-    right: 0;
     bottom: 100%;
     max-height: 300px;
     transition: var(--media-selectmenu-transition-in,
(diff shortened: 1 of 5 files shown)


src/js/experimental/media-chrome-selectmenu.js
```diff
@@ -198,6 +197,8 @@ class MediaChromeSelectMenu extends globalThis.HTMLElement {
   #hide() {
     if (this.#listboxSlot.hidden) return;
 
+    unobserveResize(getBoundsElement(this), this.#updateMenuPosition);
```
cjpillsbury, line 200: question(non-blocking): Any reason for the reorder here?

luwes: yes, if the listbox is hidden we don't want the updateMenuPosition to muddle the transition in case the element gets a resize observe event

cjpillsbury (approved):
👌 LGTM

PR #757, opened 2023-11-02
fix: calc padding CSS var issue

luwes:
this improves an issue with calculating the range width and the padding that can be set via CSS vars.

previously if the CSS var was set to `0` or another valid padding value `3px 5px` for example the `calc()` function would fail.

this new approach requires 2 padding elements but it doesn't fail against CSS vars like above


cjpillsbury (approved):
😎

PR #766, opened 2023-11-15
fix: Android toggle controls on tapping

luwes:
this change fixes an issue where tapping the player on Android would show and hide the controls right away instead of scheduling the hiding after a timeout.

also fixes a small issue where the time range animation would not start right away when the time range became visible,
the animation is disabled when it gets hidden to save CPU.

### DIFF
--- a/src/js/media-container.js
+++ b/src/js/media-container.js
@@ -265,6 +265,7 @@ class MediaContainer extends globalThis.HTMLElement {
       ].includes(name));
   }
 
+  #pointerDownTimeStamp = 0;
   breakpointsComputed = false;
 
   constructor() {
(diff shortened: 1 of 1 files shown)


src/js/media-container.js
```diff
@@ -477,98 +478,116 @@ class MediaContainer extends globalThis.HTMLElement {
...
+  }
 
-    const setActive = () => {
-      if (!this.hasAttribute(Attributes.USER_INACTIVE)) return;
+  #handlePointerMove(event) {
+    if (event.pointerType !== 'mouse') {
+      // On mobile we toggle the controls on a tap which is handled in pointerup,
+      // but Android fires pointermove events even when the user is just tapping.
+      // Prevent calling setActive() on tap because it will mess with the toggle logic.
+      const MAX_TAP_DURATION = 200;
```
cjpillsbury, line 521: per out of band convo:
```suggestion
      const MAX_TAP_DURATION = 250;
```

cjpillsbury (approved):
👍

PR #769, opened 2023-11-15
fix: time range bug fixes

luwes:
- [x] start animation on timerange visible
- [x] account for dragging a CSS transformed (rotate, scale) range on mobile

### DIFF
--- a/src/js/media-time-range.js
+++ b/src/js/media-time-range.js
@@ -263,6 +263,7 @@ class MediaTimeRange extends MediaChromeRange {
     this.range.setAttribute('aria-label', nouns.SEEK());
     this.#toggleRangeAnimation();
 
+    // NOTE: Adding an event listener to an ancestor here.
     this.#rootNode = this.getRootNode();
     this.#rootNode?.addEventListener('transitionstart', this);
   }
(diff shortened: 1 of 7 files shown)


src/js/media-time-range.js
```diff
@@ -254,18 +256,23 @@ class MediaTimeRange extends MediaChromeRange {
...
     this.#animation = new RangeAnimation(this.range, this.#updateRange, 60);
-    this.addEventListener('transitionstart', this);
   }
 
   connectedCallback() {
     super.connectedCallback();
     this.range.setAttribute('aria-label', nouns.SEEK());
     this.#toggleRangeAnimation();
+
+    this.#rootNode = this.getRootNode();
```
cjpillsbury, line 266: I don't want to block this PR based on adding an event listener to an ancestor, but can you put a comment here so we can keep track of this in our audit of potential GC/dangling references? Something like:
```js
// NOTE: Adding an event listener to an ancestor here.
```

cjpillsbury (approved):
Haven't done a ton of smoke testing here, but overall LGTM. Had one ask to help with our eventual memory management audit/eval.

PR #770, opened 2023-11-16
feat: Add user prefs for subtitles+cc lang, including local storage.

cjpillsbury:
Preview: https://media-chrome-git-fork-cjpillsbury-feat-media-lang-su-7022a5-mux.vercel.app/examples/vanilla/control-elements/media-captions-selectmenu.html


src/js/utils/captions.js
```diff
@@ -271,16 +272,38 @@ export const toggleSubsCaps = (el, force) => {
...
+      subTrackObjs = parseTextTracksStr(
         el.getAttribute(MediaUIAttributes.MEDIA_SUBTITLES_LIST) ?? ''
-      ) ?? [];
-    if (subTrackStr) {
+      );
+    }
+
+    let subTrack = subTrackObjs[0];
+    if (!noSubtitlesPref) {
+      const subtitlesPref = globalThis.localStorage.getItem('media-chrome-pref-subtitles');
```
heff, line 284: Since this is for languages let's make the pref name more specific, like `media-chrome-pref-subtitles-lang`. I can imagine other subtitle related prefs.

Should the pref be across all tracks? `media-chrome-pref-texttrack-lang`

src/js/utils/captions.js
```diff
@@ -250,10 +250,11 @@ export const areSubsOn = (el) => {
  *
  * This was originally in media-captions-button.
  *
- * @param {HTMLElement} el - An HTMLElement that has caption related attributes on it.
+ * @param {HTMLElement & { mediaSubtitlesList?: any[] }} el - An HTMLElement that has caption related attributes on it.
  * @param {boolean} [force] - An optional boolean that will force captions to the given state. True for on and false for off
+ * @param {boolean} [noSubtitlesPref = false] - An optional boolean that will force captions to the given state. True for on and false for off
  */
-export const toggleSubsCaps = (el, force) => {
+export const toggleSubsCaps = (el, force, noSubtitlesPref = false) => {
```
heff, line 257: This might be a different topic of old wounds but remind me why this can't just be handled by the controller, and has to be in the UI components?

src/js/experimental/media-captions-listbox.js
```diff
@@ -135,6 +136,17 @@ class MediaCaptionsListbox extends MediaChromeListbox {
         detail: this.value,
       }
     );
+
+    const preferredLanguage = parseTextTrackStr(this.value)?.language;
+    if (preferredLanguage) {
+      const mediaController = this.hasAttribute(MediaStateReceiverAttributes.MEDIA_CONTROLLER)
+        // @ts-ignore
+        ? closestElementSelector(this, `#${this.getAttribute(MediaStateReceiverAttributes.MEDIA_CONTROLLER)}`)
+        : closestComposedNode(this, 'media-controller');
```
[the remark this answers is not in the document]
heff: An alternative at the same level might be that the controller sets the user pref as a state?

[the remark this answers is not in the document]
cjpillsbury: per out of band convo, will try to simplify this via the introduction of a new `mediatogglesubtitlesrequest` (per naming convention in our current usage and EU standards, "subtitles" includes both subs & cc)

src/js/utils/captions.js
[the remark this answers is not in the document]
cjpillsbury: I went with the negative phrasing so the "standard usage" for this method wouldn't require a refactor. I could remove the default if that feels bad, since `undefined` is falsy?

cjpillsbury: Per out of band convo, this was because of prior hesitancy to introduce a "toggle" subs/cc request event. We'll be introducing it now to avoid these acrobatics.

src/js/utils/captions.js
cjpillsbury: Are you thinking `texttrack` because you envision going forward that this would also apply to e.g. `chapters` or `metadata` text tracks with a defined language? Based on our prior convos around "app"/"player" language (e.g. dialogs, tooltips, A11Y language, etc.), I can see arguments for treating these cases more like those cases. On the other hand, these would likely be "media specific (aka strongly tied to a particular media `src`), which roughly resembles subtitles/cc. My S1 is to treat it like our (yet to be implemented) "app"/"player" language settings (likely tied to the `lang` attribute/prop), but I'm down to discuss since it's def less clear cut than the others. If we stick with my S1, I'd lean towards `media-chrome-pref-subtitles-lang` (or similar).

Or, was that a slip and you didn't actually intend to swap out `subtitles` for `texttrack`?

src/js/utils/captions.js
cjpillsbury: No longer an issue with the refactor that adds the toggle request event.

src/js/experimental/media-captions-listbox.js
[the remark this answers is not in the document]
cjpillsbury: No longer an issue with the refactor for the toggle request event.

src/js/utils/captions.js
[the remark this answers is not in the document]
cjpillsbury: No longer an issue with the refactor that adds the toggle request event.

src/js/experimental/media-captions-listbox.js
```diff
@@ -1,16 +1,33 @@
...
   <path d="M22.83 5.68a2.58 2.58 0 0 0-2.3-2.5c-3.62-.24-11.44-.24-15.06 0a2.58 2.58 0 0 0-2.3 2.5c-.23 4.21-.23 8.43 0 12.64a2.58 2.58 0 0 0 2.3 2.5c3.62.24 11.44.24 15.06 0a2.58 2.58 0 0 0 2.3-2.5c.23-4.21.23-8.43 0-12.64Zm-11.39 9.45a3.07 3.07 0 0 1-1.91.57 3.06 3.06 0 0 1-2.34-1 3.75 3.75 0 0 1-.92-2.67 3.92 3.92 0 0 1 .92-2.77 3.18 3.18 0 0 1 2.43-1 2.94 2.94 0 0 1 2.13.78c.364.359.62.813.74 1.31l-1.43.35a1.49 1.49 0 0 0-1.51-1.17 1.61 1.61 0 0 0-1.29.58 2.79 2.79 0 0 0-.5 1.89 3 3 0 0 0 .49 1.93 1.61 1.61 0 0 0 1.27.58 1.48 1.48 0 0 0 1-.37 2.1 2.1 0 0 0 .59-1.14l1.4.44a3.23 3.23 0 0 1-1.07 1.69Zm7.22 0a3.07 3.07 0 0 1-1.91.57 3.06 3.06 0 0 1-2.34-1 3.75 3.75 0 0 1-.92-2.67 3.88 3.88 0 0 1 .93-2.77 3.14 3.14 0 0 1 2.42-1 3 3 0 0 1 2.16.82 2.8 2.8 0 0 1 .73 1.31l-1.43.35a1.49 1.49 0 0 0-1.51-1.21 1.61 1.61 0 0 0-1.29.58A2.79 2.79 0 0 0 15 12a3 3 0 0 0 .49 1.93 1.61 1.61 0 0 0 1.27.58 1.44 1.44 0 0 0 1-.37 2.1 2.1 0 0 0 .6-1.15l1.4.44a3.17 3.17 0 0 1-1.1 1.7Z"/>
 </svg>`;
 
 const slotTemplate = document.createElement('template');
-slotTemplate.innerHTML = /*html*/`
+slotTemplate.innerHTML = /*html*/ `
```
luwes, line 30: could this space be removed? 
Sublime text removes the color highlighting if there's a space :(

src/js/experimental/media-captions-listbox.js
```diff
@@ -39,16 +56,22 @@ class MediaCaptionsListbox extends MediaChromeListbox {
...
+      oldValue !== newValue
+    ) {
       this.value = newValue;
-
     } else if (attrName === 'aria-multiselectable') {
       // diallow aria-multiselectable
       this.removeAttribute('aria-multiselectable');
-      console.warn("Captions List doesn't currently support multiple selections. You can enable multiple items via the media.textTrack API.");
+      console.warn(
+        "Captions List doesn't currently support multiple selections. You can enable multiple items via the media.textTrack API."
```
luwes, line 73: `media.textTracks` or just `TextTrack`

src/js/experimental/media-captions-listbox.js
```diff
@@ -135,6 +185,27 @@ class MediaCaptionsListbox extends MediaChromeListbox {
...
+    if (preferredLanguage) {
+      const mediaController = this.hasAttribute(
+        MediaStateReceiverAttributes.MEDIA_CONTROLLER
+      )
+        ? // @ts-ignore
+          closestElementSelector(
+            this,
+            `#${this.getAttribute(
+              MediaStateReceiverAttributes.MEDIA_CONTROLLER
+            )}`
```
luwes, line 199: I would keep the id linking contained in the same shadow DOM I think.
that's how native HTML also behaves afaik. thoughts?

src/js/experimental/media-captions-listbox.js
cjpillsbury: of course! if needed I can re-implement without the silly ide/prettier reformatting.

src/js/experimental/media-captions-listbox.js
cjpillsbury: Cool I didn't change that text but happy to update 😄 (change here is just another auto-reformatting)

src/js/experimental/media-captions-listbox.js
cjpillsbury: ah this is actually another thing that I think can be completely removed after my refactor to support toggle requests, but I'll confirm. If not, we are in a tricky situation with well-defined nested component cases (e.g. in this case, `media-captions-selectmenu` having a `media-captions-listbox` in its shadow DOM. We could, ofc, pass the value down instead? Either way, I think this should be irrelevant and not a thing we need to hash out right now(🤞)

cjpillsbury: no longer relevant bc no longer needed 🎉

examples/vanilla/control-elements/media-captions-selectmenu.html
```diff
@@ -71,11 +73,12 @@ <h2>Disabled</h2>
       muted
       crossorigin
     >
-      <track label="English" kind="captions" srclang="en" src="../vtt/elephantsdream/captions.en.vtt"></track>
+      <track label="German" kind="captions" srclang="de" src="../vtt/elephantsdream/captions.de.vtt"></track>
```
cjpillsbury, line 76: Note: Changed this order to ensure we select navigator lang by default (instead of simply the first one).

This only applies if the local storage pref is disabled or if nothing has been stored to the local storage for the user pref yet.

src/js/experimental/media-captions-listbox.js
```diff
@@ -122,10 +122,35 @@ class MediaCaptionsListbox extends MediaChromeListbox {
...
+            detail: showingSubs,
+          }
+        )
+      );
+    }
 
-    if (!this.value) return;
+    // NOTE: The same applies here, where we only need to request a state change
+    // if `onChange` is a result of a state change local to the component (via user
+    // interaction or direct programmatic state change of `this.value`) (CJP)
```
luwes, line 152: again, the comment is making it more complex than reading the code IMO

src/js/experimental/media-captions-listbox.js
```diff
@@ -122,10 +122,35 @@ class MediaCaptionsListbox extends MediaChromeListbox {
...
+    const showingSubs = this.mediaSubtitlesShowing;
+    const showingSubsStr = this.getAttribute(
+      MediaUIAttributes.MEDIA_SUBTITLES_SHOWING
+    );
+    // NOTE: Since 'change' events and thus the onChange method will be fired/invoked
+    // from both user interaction and programmatic value changes, *including those
+    // propagated from the media controller itself*, we can check if the value (which
+    // will change from user interactions/direct programmatic changes) is actually different
+    // the mediaSubtitlesShowing value (which won't have been updated yet if the change was
+    // "local to the component") to determine if we need to request state changes. (CJP)
```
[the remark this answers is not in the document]
cjpillsbury: Yeah it did a full round trip twice in these cases. Since the places with comments got missed by our core team, I want to include some comment in these places. Given your concerns, I'll make it brief:
`// Don't make request if this was the result of a media state state change`

src/js/utils/captions.js
heff: Not a slip, just exploring if once a user sets a lang pref in subtitles if it's be a positive user experience to default other tracks to that lang when possible. I think there's something there, but no need to solve here. media-chrome-pref-subtitles-lang 👍

src/js/experimental/media-captions-listbox.js
```diff
@@ -122,10 +122,29 @@ class MediaCaptionsListbox extends MediaChromeListbox {
...
 
   #onChange() {
-    // turn off currently selected tracks
-    toggleSubsCaps(this, false);
+    const showingSubs = this.mediaSubtitlesShowing;
+    const showingSubsStr = this.getAttribute(
+      MediaUIAttributes.MEDIA_SUBTITLES_SHOWING
+    );
+
+    // Don't make request if this was the result of a media state change (CJP)
```
luwes, line 130: ty 👍

src/js/media-captions-button.js
```diff
@@ -18,7 +17,7 @@ const ccIconOff = `<svg aria-hidden="true" viewBox="0 0 26 24">
 </svg>`;
 
 const slotTemplate = document.createElement('template');
-slotTemplate.innerHTML = /*html*/`
+slotTemplate.innerHTML = /*html*/ `
```
luwes, line 20: another added space here

src/js/controller.js
```diff
@@ -793,16 +793,70 @@ export const MediaUIRequestHandlers = {
...
+          subTrack = preferredAvailableSubs[0];
+        }
+      }
+      const { language, label, kind } = subTrack;
+      updateTracksModeTo(
+        TextTrackModes.SHOWING,
+        tracks,
+        [{ language, label, kind }]
+      );
+    }
```
luwes, line 858: I would suspect the `force` var still to be used somewhere here.
thinking through this...

luwes (approved):
LGTM

src/js/utils/element-utils.js
```diff
@@ -28,6 +28,14 @@ export const closestComposedNode = (childNode, selector) => {
   return closestComposedNode(childNode.getRootNode().host, selector);
 };
 
+export const closestElementSelector = (childNode, selector) => {
```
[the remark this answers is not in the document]
cjpillsbury: good catch. removed.

src/js/media-captions-button.js
cjpillsbury: boooooo! stupid ides. Fixing.

src/js/controller.js
cjpillsbury: should be accounted for on https://github.com/muxinc/media-chrome/pull/770/files#diff-f47c60f82e0b188cc72db120e2829c4473753d20186183b4c2e1492169d61321R826 right?

cjpillsbury: oh no you're right. We don't currently have a use case in our codebase but we should account for that. Updating.

cjpillsbury: also just re-thought through this and my former self did account for that... def required some reasoning. The tl;dr:

1. The early bail cases for `forced` say "if we're already in the forced state we want, we don't have to do anything"
2. Then, we're only in cases where we used `forced` to cause the same state change we already would have without specifying `forced`.
3. So, whether we stipulated `forced` or not, we should be disabling subtitle tracks that are currently showing or showing a subtitle track because none are showing.

PR #775, opened 2023-12-01
fix: video element memory leak + optimizations

### LINKED ISSUE #750: Detached HTMLVideoElements leaks when used with react
Steps to reproduce:
1. Modify examples/nextjs-with-typescript/pages/index.tsx by adding a button that mounts and unmounts <MediaController /> e.g:
```typescript
const [mounted, setMounted] = useState(false);
return (
....
      <button onClick={() => setMounted(prev => !prev)}>
          {mounted ? "Unmount": "Mount"}
      </button>
      {mounted && (<MediaController />)}
....
```
2. Run example and open it in chrome.
3. Click on button a few times to mount and unmount controller
4. Create memory snapshot with devtools memory tab. Search for "video" in class filter

Actual result:
Observe a list that looks like this
image

Note: number of Detached HTMLVideoElements will be equal to number of times "unmount" is clicked + 1.

Expected result:
Only one Detached HTMLVideoElement is present in snapshots. Clicking on button won't change the number of Detached HTMLVideoElements.

luwes:
fixes #750


luwes:
ha! I put the track elements back in and the memory leak is back... digging

```html
            <track
              label="thumbnails"
              default
              kind="metadata"
              src="https://image.mux.com/DS00Spx1CV902MCtPj5WknGlR102V5HFkDe/storyboard.vtt"
            />
            <track
              label="English"
              kind="captions"
              srcLang="en"
              src="./vtt/en-cc.vtt"
            />
```

src/js/media-container.js
```diff
@@ -484,11 +454,31 @@ class MediaContainer extends globalThis.HTMLElement {
...
-    });
+    globalThis.window?.addEventListener('mouseup', this);
+  }
+
+  disconnectedCallback() {
+    // When disconnected from the DOM, remove root node and media event listeners
+    // to prevent memory leaks and unneeded invisble UI updates.
+    if (this.media) {
+      this.mediaUnsetCallback(this.media);
+    }
```
luwes, line 465: this fixed the 1st memory leak caused by the global event listener or the event listener to the root node related to fullscreen change.

src/js/media-controller.js
```diff
@@ -152,18 +154,37 @@ class MediaController extends MediaContainer {
...
+
+    this.enableHotkeys();
+  }
+
+  disconnectedCallback() {
+    // mediaUnsetCallback() is called in super.disconnectedCallback();
+    super.disconnectedCallback();
+    this.disableHotkeys();
+    // Disable captions on disconnect to prevent a memory leak if they stay enabled.
+    toggleSubsCaps(this, false);
```
luwes, line 181: this fixed a 2nd memory leak, I don't know the specifics but when subtitles stay enabled the detached video element doesn't get GC'd

cjpillsbury: @luwes in flight PR #770 removes the `toggleSubsCaps()` utility method. Can you refactor to something like this instead?
https://github.com/muxinc/media-chrome/pull/770/files/681fe4d8cb5fc323b663f52de67a77fa2351b7fa#diff-2b43f6e6b30d1c84504e425bd4fcaa330513d0389aed6f8127ecef8bbae96629R138-R147

src/js/media-controller.js
```diff
@@ -190,7 +211,7 @@ class MediaController extends MediaContainer {
       });
 
       rootEvents?.forEach((eventName) => {
-        this.getRootNode().addEventListener(eventName, handler);
+        this.#rootNode?.addEventListener(eventName, handler);
```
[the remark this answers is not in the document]
luwes: pretty sure it will yes, the #rootNode gets updated on each connectedCallback

PR #779, opened 2023-12-05
Feat/default duration

cjpillsbury:
Adds support for `defaultduration` to apply before media is loaded.

**NOTE:** The `<media-time-range>` can still be interacted with before the media has loaded, but that was already true.

### DIFF
--- a/examples/vanilla/advanced.html
+++ b/examples/vanilla/advanced.html
@@ -31,13 +31,14 @@
   <body>
     <main>
       <h1>Media Chrome Advanced Video Usage Example</h1>
-      <media-controller defaultsubtitles>
+      <media-controller defaultsubtitles defaultduration="134">
         <video
           slot="media"
           src="https://stream.mux.com/A3VXy02VoUinw01pwyomEO3bHnG4P32xzV7u1j1FSzjNg/high.mp4"
           muted
           crossorigin
           playsinline
+          preload="none"
         >
           <track
             label="thumbnails"
--- a/src/js/controller.js
+++ b/src/js/controller.js
@@ -150,18 +150,24 @@ export const MediaUIStates = {
   MEDIA_DURATION: {
     get: function (controller) {
       const { media } = controller;
-      const defaultDuration = controller.hasAttribute('defaultduration')
-        ? +controller.getAttribute('defaultduration')
-        : NaN;
 
-      // TODO: What if duration is infinity/live? (heff)
-      if (!media || !Number.isFinite(media.duration)) {
-        return defaultDuration;
+      // If `defaultduration` is set and we don't yet have a usable `duration`
+      // available, use the default duration.
+      if (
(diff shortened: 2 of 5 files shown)


examples/vanilla/advanced.html
```diff
@@ -31,13 +31,14 @@
   <body>
     <main>
       <h1>Media Chrome Advanced Video Usage Example</h1>
-      <media-controller defaultsubtitles>
+      <media-controller defaultsubtitles defaultduration="134">
```
cjpillsbury, line 34: **NOTE:** we can revert this if preferred, though I think it might be a good example for `advanced`.

src/js/controller.js
```diff
@@ -150,10 +150,18 @@ export const MediaUIStates = {
...
+        : NaN;
 
       // TODO: What if duration is infinity/live? (heff)
       if (!media || !Number.isFinite(media.duration)) {
-        return NaN;
+        return defaultDuration;
+      }
+
+      // Apply defaultDuration if it's set and the media is not yet ready
+      if (!(media.readyState || Number.isNaN(defaultDuration))) {
```
luwes, line 163: why not rely on media.duration? media.duration >= 0 and not NaN 
could backfire with custom media elements that have a limited media API

cjpillsbury: Interesting thought! I think `readyState` better represents the actual conditions, but only relying on `duration` sounds like it would be good simply because it reduces the barrier of entry for custom slotted media elements' API. _**technically**_ 0 duration media is valid, but I think we can ignore that wild scenario unless/until it comes up. Sold.

luwes: right `0` would be wild :) was thinking current time

luwes (approved):
LGTM

src/js/controller.js
```diff
@@ -151,9 +151,23 @@ export const MediaUIStates = {
...
       const { media } = controller;
 
+      // If `defaultduration` is set and we don't yet have a usable `duration`
+      // available, use the default duration.
+      if (
+        controller.hasAttribute('defaultduration') &&
+        (!media ||
+          !media.duration ||
+          Number.isNaN(media.duration) ||
+          !Number.isFinite(media.duration))
```
heff, line 161: If the source is live (Infinite), would this screw up some things?

src/js/controller.js
```diff
@@ -151,9 +151,23 @@ export const MediaUIStates = {
...
+      ) {
+        return +controller.getAttribute('defaultduration');
+      }
+
+      // If `defaultduration` is unset, we still want to propagate `NaN`
+      // for some cases to ensure appropriate media state receiver updates.
       // TODO: What if duration is infinity/live? (heff)
```
heff, line 168: Oh, ha. Still a future us problem? Or can we remove this note?

heff (approved):
A few questions but lgtm.

PR #781, opened 2023-12-07
fix(pip-unsupported-apple-pwa): Detect Apple+Safari+PWA for no PiP su…

cjpillsbury:
…pport. Add PWA example.

**To Test**
- go here https://media-chrome-git-fork-cjpillsbury-fix-pip-unsupporte-fcc966-mux.vercel.app/examples/vanilla/pwa/
- For iOS
    - PiP should be available if you have not installed as a PWA
    - PiP should not be available if you have installed as a PWA
- For macOS + Chrome PiP should be available for both PWA and in browser
- For Android + Chrome PiP should be available for both PWA and in browser


luwes (approved):
sweet! LGTM. tested on iOS and Mac Chrome

PR #782, opened 2023-12-11
fix: prevent unset callback if there is no media

luwes:
related https://github.com/muxinc/elements/issues/849


AdamJaggard (approved):
This looks sensible!

PR #784, opened 2024-01-02
Feat/media store rearchitecture

cjpillsbury:


examples/nextjs-with-typescript/pages/MediaChromeReactSimple.tsx
```diff
@@ -0,0 +1,396 @@
...
+import { MediaUIEvents } from '../../../dist/constants';
+
+const PlayButton = () => {
+  const dispatch = useDispatch();
+  const mediaPaused = useSelector((state) => state.mediaPaused);
+  return (
+    <button
+      style={{ cursor: 'pointer' }}
+      onClick={() => {
+        const type = mediaPaused ? 'mediaplayrequest' : 'mediapauserequest';
```
luwes, line 24: since these event names will be more public and frequently used just thought of building a new shelter type building for my bicycle here :) 

we good with the naming? the `request` part arguably could be left off, just a thought.
it could also be namespaced potentially `MediaRequestEvents.PLAY` or `MediaActions.PLAY`...

examples/vanilla/themes/minimal-theme.html
```diff
@@ -7,7 +7,7 @@
       rel="stylesheet"
       href="https://cdn.jsdelivr.net/npm/water.css@2/out/water.css"
     />
-    <script src="https://cdn.jsdelivr.net/npm/@mux/mux-video@0"></script>
+    <script type="module" src="https://cdn.jsdelivr.net/npm/@mux/mux-video"></script>
```
luwes, line 10: 👍 😅 that was a blocking script, reminds me of our Mux player docs 
https://docs.mux.com/guides/mux-player-web#example-html-element-implementation
we should really put a defer on that

src/js/constants.js
```diff
@@ -152,11 +152,24 @@ export const PointerTypes = {
   TOUCH: 'touch',
 };
 
+/**
+ * @type {{
+ *   UNAVAILABLE: 'unavailable';
+ *   UNSUPPORTED: 'unsupported';
+ * }}
+ */
```
luwes, line 160: TS didn't derive this?
that's a shame

src/js/media-controller.js
```diff
@@ -65,60 +70,96 @@ class MediaController extends MediaContainer {
...
+        defaultStreamType: /** @type {import('./mediaStore.js').StreamTypeValue} */ (this.getAttribute(Attributes.DEFAULT_STREAM_TYPE)) ?? undefined,
+        liveEdgeOffset: this.hasAttribute(Attributes.LIVE_EDGE_OFFSET) ? +this.getAttribute(Attributes.LIVE_EDGE_OFFSET) : undefined,
+        // NOTE: This wasn't updated if it was changed later. Should it be? (CJP)
+        noVolumePref: this.hasAttribute(Attributes.NO_VOLUME_PREF),
+        noSubtitlesLangPref: this.hasAttribute(Attributes.NO_SUBTITLES_LANG_PREF),
+      },
+      // NOTE: When using a default store, we're going to assume that we should stop monitoring the state owners whenever
+      // there are no subscribers. This allows us to do proper teardown/GC when the media-controller is no longer in use.
+      /** @TODO Re-fix me! (current solution may be buggy) (CJP) */
+      monitorStateOwnersOnlyWithSubscriptions: true,
```
luwes, line 140: should this looked at before merging?

src/js/media-controller.js
```diff
@@ -65,60 +70,96 @@ class MediaController extends MediaContainer {
...
+    }
+    this.#mediaStore = value;
+
+    if (!this.#mediaStore && !this.hasAttribute(Attributes.NO_DEFAULT_STORE)) {
+      this.#setupDefaultStore();
+      return;
+    }
+
+    this.#mediaStoreUnsubscribe = this.#mediaStore?.subscribe(this.#mediaStateCallback);
+  }
```
luwes, line 161: nice!

src/js/media-controller.js
```diff
@@ -128,6 +169,41 @@ class MediaController extends MediaContainer {
       this.removeAttribute(Attributes.FULLSCREEN_ELEMENT);
     }
     this.#fullscreenElement = element;
+    // Use the getter in case the fullscreen element was reset to "`this`"
+    this.#mediaStore.dispatch({ type: 'fullscreenelementchangerequest', detail: this.fullscreenElement });
+  }
+
+  connectedCallback() {
+
+    super.connectedCallback?.();
```
luwes, line 178:
```suggestion
    super.connectedCallback?.();

```

src/js/media-controller.js
```diff
@@ -128,6 +169,41 @@ class MediaController extends MediaContainer {
       this.removeAttribute(Attributes.FULLSCREEN_ELEMENT);
     }
     this.#fullscreenElement = element;
+    // Use the getter in case the fullscreen element was reset to "`this`"
+    this.#mediaStore.dispatch({ type: 'fullscreenelementchangerequest', detail: this.fullscreenElement });
```
luwes, line 173: this is a long one 😅 `fullscreenelementchangerequest` 
I would remove request here, `request` is implied I think in each action / dispatch. it could fail...

src/js/media-controller.js
```diff
@@ -406,16 +341,11 @@ class MediaController extends MediaContainer {
...
-    Object.keys(MediaUIStates).forEach((stateConstName)=>{
-      const stateDetails = MediaUIStates[stateConstName];
-
-      propagateMediaState(
-        [el],
-        MediaUIProps[stateConstName],
-        stateDetails.get(this)
-      );
-    });
+    if (this.#mediaStore) {
```
luwes, line 344:
```suggestion

    if (this.#mediaStore) {
```

a little white space helps with code readability

src/js/media-controller.js
```diff
@@ -552,14 +470,7 @@ class MediaController extends MediaContainer {
...
       case 'ArrowRight':
-        currentTimeStr = this.getAttribute(
-          MediaUIAttributes.MEDIA_CURRENT_TIME
-        );
-        currentTime =
-          currentTimeStr && !Number.isNaN(+currentTimeStr)
-            ? +currentTimeStr
-            : DEFAULT_TIME;
-        detail = Math.max(currentTime + seekOffset, 0);
+        detail = Math.max((this.mediaStore.getState().mediaCurrentTime ?? 0) + DEFAULT_SEEK_OFFSET, 0);
```
luwes, line 473: cool! 😎

examples/nextjs-with-typescript/pages/MediaChromeReactSimple.tsx
cjpillsbury: tl;dr preamble - per our prior convo, let's make sure we keep any of the react-specific stuff out of scope for this, since none of it needs to be set in stone for our initial merge. If preferred, I can drop all of the examples here from the branch/PR.

Now, to your actual question:
I like this question/these ideas. I'd be down with a rename. Keep in mind that these are (at least currently) used by both the "actual" eventing system of media controller/Media Chrome web components and the media store (as a simple pass through), so if we rename that may introduce complexity or require a renaming of the events themselves. The constants lookup table may be a good resolution here and is fairly common in, e.g. the redux world for actions. @heff thoughts.

src/js/constants.js
cjpillsbury: Yeah, it's a bit more "loosey goosey" between the .js and the default inferred object typings. If this were actual typescript, the solution is pretty terse:
```ts
export const AvailabilityStates = {
  UNAVAILABLE: 'unavailable',
  UNSUPPORTED: 'unsupported',
} as const; // the as const does all the work of e.g. inferring keys as literal types.
```

src/js/media-controller.js
cjpillsbury: Should be fixed but yes I should revisit 💯

src/js/media-controller.js
cjpillsbury: This one's a bit more flexible compared to the prior event rename suggestion. In our current architecture, we wouldn't typically expect a developer to dispatch these events, even in the web components world. i.e. this is more "under the hood" eventing/dispatching. On the plus side, the naming convention is consistent and explicit/clear on what it does with its current name. On the minus side, it's a big boi. My primary concern with `'fullscreenelementchange'` is it sounds like it is an event announcing that `fullscreenElement` **_has_** changed (like other "change" events), when it's actually an event asking to **_cause_** a change. @heff thoughts?

src/js/constants.js
cjpillsbury: Yeah between this being .js and the default behavior for inferred types on complex objects, it ends up being a bit more "loosey goosey". If this were written in TypeScript, the solution would be much more terse/simple 💔, e.g.:
```ts
const AvailabilityStates = {
    UNAVAILABLE: 'unavailable',
    UNSUPPORTED: 'unsupported',
} as const; // the `as const` does all the work of e.g. inferring keys as literal types.
```

luwes (dismissed):
overall looks great to me! 
media store is indeed a huge file, would be good to split it up somewhat.

blocking: 
- rename to media-store.js
- all (many) menu examples w/ external controller are broken

controller.js could be removed

cjpillsbury:
> * all (many) menu examples w/ external controller are broken

yeah this was another refactor since the last update 💔 let me investigate here.

examples/nextjs-with-typescript/pages/MediaChromeReactSimple.tsx
heff: A couple of worries.
* We bubble events a lot, out of the shadow dom, so I worry about some of our events leaking into other systems and causing problems if they have generic or overlapping names like `play`.
* These events are intended to separate user intentions from actual media state changes, so anything that makes them look more like the related media events I think will cause confusion.

I think I could get behind dropping the `media` part, but I still worry about conflicts down the line. Is `MediaUserEvents.PLAY_REQUEST` really better than `mediaplayrequest`?

src/js/media-controller.js
```diff
@@ -65,60 +70,88 @@ class MediaController extends MediaContainer {
...
-        }
+    if (!this.hasAttribute(Attributes.NO_DEFAULT_STORE)) {
+      this.#setupDefaultStore();
+    }
 
-        MediaUIRequestHandlers[key](this.media, e, this);
-      };
-      this.addEventListener(MediaUIEvents[key], this[handlerName]);
-    });
+    this.enableHotkeys();
```
heff, line 113: Is it ok to move this here instead of connectedCallback?

src/js/media-controller.js
```diff
@@ -278,77 +273,11 @@ class MediaController extends MediaContainer {
...
-    // Reset to paused state
-    // TODO: Can we just reset all state here?
-    // Should hasPlayed refer to the media element or the controller?
-    // i.e. the poster might re-show if not handled by the poster el
-    this.propagateMediaState(MediaUIProps.MEDIA_PAUSED, true);
+    this.#mediaStore?.dispatch({ type: 'mediaelementchangerequest', detail: undefined });
   }
 
   propagateMediaState(stateName, state) {
-    const previousState = getStateValue(this.mediaStateReceivers, stateName);
```
heff, line 339: Glad to be doing away with this hack, though it got the job done.

heff (commented):
Still need to review mediaStore but getting these comments out there for now.

src/js/media-store.js
```diff
@@ -0,0 +1,1567 @@
...
+        // NOTE: Since custom update handlers may depend on *any* state owner
+        // we should apply them whenever any state owner changes (CJP)
+        const prevHandlerTeardown =
+          stateUpdateHandlers[stateName].stateOwnersUpdateHandlers;
+        stateOwnersUpdateHandlers.forEach((fn) => {
+          if (prevHandlerTeardown) {
+            prevHandlerTeardown();
+          }
+          stateUpdateHandlers[stateName].stateOwnersUpdateHandlers = fn(
+            handler,
```
heff, line 1294: I'm getting a little lost in handlers on handlers on handlers.

src/js/media-store.js
```diff
@@ -0,0 +1,1567 @@
...
+            value.toString()
+          );
+        }
+      } catch (err) {
+        // ignore
+      }
+      media.volume = value;
+    },
+    mediaEvents: ['volumechange'],
+    stateOwnersUpdateHandlers: [
```
heff, line 402: I'm a little lost on this general concept. Let's review IRL.

src/js/media-store.js
```diff
@@ -0,0 +1,1567 @@
...
+          stateOwners.media = stateOwnersToKeep.media;
+          stateOwners.fullscreenElement = stateOwnersToKeep.fullscreenElement;
+          stateOwners.rootNode = stateOwnersToKeep.rootNode;
+        }
+      };
+    },
+  };
+};
+
+export default createMediaStore;
```
heff, line 1567: I'm seeing how this achieves the general goal. It's also a daunting file and hard to wrap my head fully around despite (or maybe because of) having written half the code in it. Let's find some time to chat through it and see if we can simplify the concepts or organization some.

src/js/media-controller.js
```diff
@@ -65,60 +70,88 @@ class MediaController extends MediaContainer {
...
+  set mediaStore(value) {
+    if (this.#mediaStore) {
+      this.#mediaStoreUnsubscribe?.();
+      this.#mediaStoreUnsubscribe = undefined;
+    }
+    this.#mediaStore = value;
+
+    if (!this.#mediaStore && !this.hasAttribute(Attributes.NO_DEFAULT_STORE)) {
+      this.#setupDefaultStore();
+      return;
```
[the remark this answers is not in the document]
heff: Discussed IRL: Add a comment explaining that setupDefaultStore calls mediaStore again to set unsubscribe

src/js/media-controller.js
```diff
@@ -128,6 +161,43 @@ class MediaController extends MediaContainer {
...
+      this.#mediaStore.dispatch({
+        type: MediaUIEvents.MEDIA_TOGGLE_SUBTITLES_REQUEST,
+        detail: false
+      });
+    }
+
+    this.#rootNode = undefined;
+
+    if (this.#mediaStoreUnsubscribe) {
+      this.#mediaStore?.dispatch({ type: 'rootnodechangerequest', detail: this.#rootNode });
```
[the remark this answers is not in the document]
heff: Discussed IRL: Add comment, this assumes for an external media store you want media controller to own determining the root node.

src/js/media-controller.js
```diff
@@ -148,128 +218,53 @@ class MediaController extends MediaContainer {
 
     } else if (
       attrName === Attributes.DEFAULT_SUBTITLES &&
-      newValue !== oldValue &&
-      newValue === ''
+      newValue !== oldValue
     ) {
-      this.dispatchEvent(
-        new globalThis.CustomEvent(
-          MediaUIEvents.MEDIA_TOGGLE_SUBTITLES_REQUEST,
```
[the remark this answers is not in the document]
heff: Discussed IRL: This is actually behavioral change, in a positive direction IMO. Now when you change the attribute is will not also immediately toggle captions.

src/js/media-controller.js
heff: A followup note from an IRL conversation with @cjpillsbury: 

`requests` are like Actions in redux. Actions work like events, but are clearly not events because they're Actions. Without the term 'request', and because we're using actual events for this purpose, it's not clear requests are different from other events. It might be clear when you're using the store/provider directly, but not in other contexts.

The store is built in a way where:
1. A change to state is requested
2. That change is calculated
3. The state is updated and propagated

Because of that, it makes sense to include state owner changes as part of that workflow, rather than making a special case for the and potentially adding more complexity to the internal workflow.

So then all that to say, I don't have a good answer here if the goal is to make that event name shorter. I agree it's a big one. Part of me doubts that anyone besides us will ever do anything with this specific event. I know it doesn't mean this topic won't come up again. My gut is to kick the can on this one and leave it as is.

heff:
To-dos from an IRL convo yesterday w/ @cjpillsbury and @AdamJaggard . 
* Break out Facade and util functions from the media store file
* Add all of the text from Adam's internal media store summary doc as comments in the relevant files
* Rename Facade to StateMediator
* Move event-to-state translators to a lookup object, one function per event, that will be given: stateOwners, stateMediator, action (/event)
* Pass in current state to stateMediator getter translators (for hasPlayed)
* Comment the heck out of dispatch and anything else that might not be straight forward to someone diving in to this for the first time.

cjpillsbury:
@heff + @luwes just a heads up: I gutted all of the react-specific src and examples from this branch to get it in approximately "final form" for a v1 merge. I still have those in a branch locally for any v2+ efforts though.

heff:
@cjpillsbury let me know if you're waiting on anything from me on this one.

cjpillsbury:
@heff since this is a larger refactor, I would like to get folks doing a bunch of before/after smoke tests before merging. I'll also need to re-rebase based on some recent changes, including at least one that touches on media state.

src/js/media-container.js
```diff
@@ -599,8 +599,9 @@ class MediaContainer extends globalThis.HTMLElement {
   }
 }
 
-// NOTE: Since MediaContainer is not currently designed to be used directly as
-// a custom element, unlike other components, we do not `define()` it.
+if (!globalThis.customElements.get('media-container')) {
```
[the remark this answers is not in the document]
cjpillsbury: This can be postponed for a separate PR or omitted. Tl;dr - this allows folks to use `<media-container>` without all of the state mgmt crud from `<media-controller>` (relevant for splitting out the `MediaStore`). It's not a pre-req for this effort though, so we can def back this out if you'd prefer.

luwes (approved):
LGTM! lets merge this baby!

PR #785, opened 2024-01-04
feat: add role menu elements

luwes:
- [x] move from aria `select` / `listbox` / `option` pattern to `menu` / `menuitem` pattern
- [x] add `invoketarget` attribute and auto linking between x-menu-button & x-menu elements
- [x] add anchoring logic for button / menu with an explicit `anchor` attribute or auto linking
- [x] close menu when checking item and on focus out

--- 

- it roughly makes the listboxes all menu elements. 
  a menu can hold menu items optionally with a type of radio or checkbox making this more flexible than listbox / option.
- we're doing away with a grouping element similar to `selectmenu` to make the connection between button and menu.
  this improves flexibility, less elements. it's with new upcoming / proposed API's in mind that promote this concept.
  https://developer.mozilla.org/en-US/docs/Web/API/Popover_API
  https://open-ui.org/components/invokers.explainer/
  https://www.w3.org/TR/css-anchor-position-1/
  https://developer.chrome.com/blog/tether-elements-to-each-other-with-css-anchor-positioning

I propose to make this connection automatically for ease of use, less HTML needed but it can be made explicit with the `invoketarget` <-> `id` attributes or `anchor` <-> `id` for the positioning.

example of an automatic connection:

```html
<media-controller>
  <media-captions-menu hidden></media-captions-menu>
  <media-control-bar>
    <media-captions-menu-button></media-captions-menu-button>
  </media-control-bar>
</media-controller>
```

this menu captions button will automatically open the captions menu because the button finds a valid target. 
if there is no menu it falls back to the default behavior.

the captions menu is also automatically anchored to the captions menu button position as this is the most common behavior. it will be possible to opt out with `anchor=""` or similar.


examples/vanilla/control-elements/media-playback-rate-menu.html
```diff
@@ -42,45 +42,75 @@ <h2>Column layout w/ custom icon</h2>
       muted
       crossorigin
     ></video>
+    <media-playback-rate-menu hidden>
+      <!-- Custom select indicator icon -->
+      <span slot="check-indicator" class="check-icon">
```
heff, line 47: I think this should probably be "**checked**-indicator".

src/js/experimental/menu/media-audio-track-menu.js
```diff
@@ -0,0 +1,136 @@
...
+
+  #audioTrackList = [];
+  #prevState;
+
+  attributeChangedCallback(attrName, oldValue, newValue) {
+    super.attributeChangedCallback(attrName, oldValue, newValue);
+
+    if (
+      attrName === MediaUIAttributes.MEDIA_AUDIO_TRACK_ENABLED &&
+      oldValue !== newValue
```
heff, line 38: Will this ever be true? Feels odd if it could be.

heff (approved):
Can't say I did an exhaustive test but nothing looks off to me in the code. Love this direction.

src/js/experimental/menu/media-audio-track-menu.js
luwes: yes, the first condition is the attribute name.
the value is the audio track ID

PR #788, opened 2024-01-12
feat: add settings-menu

luwes:
- [x] keyboard control requires fine tuning. focus to expanded submenu when hidden to shown settings menu

test URL: https://media-chrome-git-fork-luwes-settings-menu-mux.vercel.app/examples/vanilla/control-elements/media-settings-menu.html

SCR-20240124-mvrg

this change adds new settings-menu and settings-menu-item elements and logic to handle submenus with customizable headers, back button, chevron icons, etc...

the case of a settings-menu in another settings-menu is not yet tried or implemented, it's a more rare case I'd like to add later.

the settings-menu-item has a `description` slot which automatically gets populated with the submenu checked item text.
it's possible for the developer to opt-out by setting a slotted description element.

all icons can be customized with slots, the headers of the menu can be customized with a slot or just the title keeping the back button intact.


src/js/experimental/menu/media-settings-menu.js
```diff
@@ -0,0 +1,47 @@
...
+class MediaSettingsMenu extends MediaChromeMenu {
+  static template = template;
+
+  /**
+   * Returns the anchor element when it is a floating menu.
+   * @return {HTMLElement}
+   */
+  get anchorElement() {
+    if (this.anchor != undefined) return super.anchorElement;
+    return getMediaController(this).querySelector('media-settings-menu-button');
```
heff, line 38: Should we make anchor always manual rather than automatic? It's not obvious to me how we'd turn this off in the apple tv player-takeover style, where we don't want anchoring to happen.

heff (approved):
lgtm!

One comment about doing the apple tv style settings menu with this. It might be worth a really quick prototype of that layout too just to make sure it can work with just html & css.

src/js/experimental/menu/media-settings-menu.js
luwes: fair point, I'll add that example back in. I removed that layout for a bit to get the more common use case as quickly as possible ready for review. Removing the anchoring right now would be setting `anchor=""` an empty string but it could also be `noanchor` for example. 

I can see the case for opt-in as well, it would be closer to how that web API would work. 
I'll try out how that looks in the HTML...

AdamJaggard:
If you open a sub-menu and then resize the player the sub menu bleeds out of the player

image

src/js/experimental/menu/media-settings-menu.js
heff: Yeah, I think we definitely need something more clear than empty string. What about `anchor="auto"` or something like that, keeping the magic but making it explicit.

luwes: I made this change, `anchor="auto"` is now required for auto anchoring...

also added the nested column menu 
https://media-chrome-git-fork-luwes-settings-menu-mux.vercel.app/examples/vanilla/control-elements/media-settings-menu.html

luwes:
> If you open a sub-menu and then resize the player the sub menu bleeds out of the player
> 
> image

this one should be fixed

PR #789, opened 2024-01-18
fix: default subs async complexities

cjpillsbury:
While `defaultsubtitles` and user preference subtitles worked for simple cases, there were various async considerations that were causing issues, including e.g. `autoplay`. This PR updates the logic to be potentially less efficient but more reflective of dev + user preferences and avoids some odd state mismatch possibilities that could result from the prior implementation.

**Example**: Consider cases where the developer has enabled `defaultsubtitles`. Let's say my navigator languages includes English, and my local storage preferred language is Japanese. If the first track asynchronously added is Swedish, none of my subtitle language prefs will match, so I'll pick the Swedish track to show. Then, an English track is added asynchronously. I'll end up here again, and, although the "most preferred" language is Japanese, I'll update the showing subtitle track to the English one, since it's more reflective of the user's preferences/environment than Swedish. Finally, a Japanese track is added, so I'll end up selecting that one.

**NOTE**: This code _**could**_ be further optimized to still bail if the current selected showing track is the same as the one that will be selected, but I'd rather hold off unless/until that becomes a real concern (unless folks feel strongly).

To smoke test, use https://media-chrome-9gxm13m21-mux.vercel.app/examples/vanilla/media-elements/mux-video.html

### DIFF
--- a/examples/vanilla/media-elements/mux-video.html
+++ b/examples/vanilla/media-elements/mux-video.html
@@ -8,9 +8,14 @@
       src="https://cdn.jsdelivr.net/npm/@mux/mux-video/+esm"
     ></script>
     <script type="module" src="../../../dist/index.js"></script>
+    <script type="module" src="../../../../dist/experimental/menu/index.js"></script>
     <style>
       .examples {
         margin-top: 20px;
+      }
+
+      mux-video {
+        width: 50vw;
       }
     </style>
   </head>
@@ -22,11 +27,9 @@
           slot="media"
           preload="auto"
           playback-id="ihZa7qP1zY8oyLSQW9TS602VgwQvNdyIvlk9LInEGU2s"
-          metadata-video-id="video-id-12345"
-          metadata-video-title="Mad Max: Fury Road Trailer"
-          metadata-viewer-user-id="user-id-6789"
-          stream-type="on-demand"
           muted
+          crossorigin
+          autoplay
         >
           <track
             label="thumbnails"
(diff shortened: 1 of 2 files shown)


examples/vanilla/media-elements/mux-video.html
```diff
@@ -8,10 +8,15 @@
       src="https://cdn.jsdelivr.net/npm/@mux/mux-video/+esm"
```
cjpillsbury, line 8: NOTE: Updating example to demo some of the more advanced crud. Useful for smoke testing, too.

examples/vanilla/media-elements/mux-video.html
```diff
@@ -22,11 +27,9 @@ <h1>Media Chrome <code>&lt;mux-video/&gt;</code> Example</h1>
           slot="media"
           preload="auto"
           playback-id="ihZa7qP1zY8oyLSQW9TS602VgwQvNdyIvlk9LInEGU2s"
-          metadata-video-id="video-id-12345"
-          metadata-video-title="Mad Max: Fury Road Trailer"
-          metadata-viewer-user-id="user-id-6789"
-          stream-type="on-demand"
           muted
+          crossorigin
```
cjpillsbury, line 31: `crossorigin` should have been added even in the old example.

src/js/controller.js
```diff
@@ -503,15 +503,14 @@ export const MediaUIStates = {
     textTracksEvents: ['addtrack', 'removetrack'],
   },
   MEDIA_SUBTITLES_SHOWING: {
-    get: function (controller) {
+    get: function (controller, event) {
       // NOTE: A bit hacky, but this ensures that HAS-style textTracks (e.g. from mux-video)
       // will also respect `defaultsubtitles` (CJP)
       if (
         controller.hasAttribute('defaultsubtitles') &&
-        !controller.hasAttribute(MediaUIAttributes.MEDIA_HAS_PLAYED) &&
```
cjpillsbury, line 511: NOTE: changed the logic here. Since tracks can be added after media has started playback, particularly with autoplay + HAS (e.g. HLS), we need to approach this differently. Now, we'll end up re-applying default subtitles whenever a track is added or removed. This will result in slightly different behavior when, e.g., swapping out sources or more advanced cases of programmatically adding/removing text tracks.

heff (approved):

PR #795, opened 2024-01-31
feat: add media-chrome-dialog element

luwes:
based on https://github.com/muxinc/media-chrome/pull/788

adds a `media-chrome-dialog` element that could work for the Apple / Netlfix column style menus...

check the last commit for just the element change 
https://github.com/muxinc/media-chrome/pull/795/commits/17c5baa0175a97d6da57bccad33ac8f8e5702423

### DIFF
--- a/src/js/media-container.js
+++ b/src/js/media-container.js
@@ -102,7 +102,7 @@ template.innerHTML = /*html*/`
        * we'll want to add here any slotted elements that shouldn't get pointer-events by default when slotted
        */ ''
     }
-    ::slotted(:not([slot=media]):not([slot=poster]):not(media-loading-indicator)) {
+    ::slotted(:not([slot=media]):not([slot=poster]):not(media-loading-indicator):not([hidden])) {
       pointer-events: auto;
     }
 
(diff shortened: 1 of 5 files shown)


src/js/media-container.js
```diff
@@ -170,7 +170,8 @@ template.innerHTML = /*html*/`
       display: none;
     }
 
-    ::slotted([role=menu]) {
+    ::slotted([role="menu"]),
+    ::slotted([role="dialog"]) {
```
heff, line 174: I'd expect the default of a "dialog" to be centered on the player or aligned top, not the bottom.

heff (approved):
Cool. This makes sense as a component to me. It's not meant to replace the settings menu, right?

So we're on the same page, when I talk about the apple TV menu style, I mean like this:

image

src/js/media-container.js
luwes: updated to be centered for role=dialog.
the vertical centering can't be done in the default slot.
that would have to be done via CSS or putting the dialog in the centered-chrome slot

PR #817, opened 2024-02-13
fix(fullscreen): handleMediaUpdated / rootNode race condition

luwes:
fixes https://github.com/epicweb-dev/kcdshop/issues/189

the async logic in handleMediaUpdated was causing rootNode to be undefined when trying to add event listeners.

this bug was only apparent when rapidly connecting / disconnecting media like in many modern JS frameworks


cjpillsbury (approved):
Approving preemptively pending out of band discussion answers on race condition concerns compared to prior fix.

PR #824, opened 2024-02-15
fix: poster slot fade after play before playing

### LINKED ISSUE #823: slotted poster transition bug
by default, default slotted elements get a opacity transition applied which is causing a weird effect for a slotted poster img.

workarounds are adding a `noautohide` attribute to the img or a `opacity: 1 !important;` style rule.

luwes:
fix #823

### DIFF
--- a/src/js/media-container.js
+++ b/src/js/media-container.js
@@ -144,7 +144,7 @@ template.innerHTML = /*html*/`
     }
 
     ${/* Only add these if auto hide is not disabled */ ''}
-    ::slotted(:not([slot=media]):not([${Attributes.NO_AUTOHIDE}]):not([hidden])) {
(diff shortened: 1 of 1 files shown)


cjpillsbury:
Per @luwes's advice: before/after smoke tested using
* Before: https://media-chrome.mux.dev/examples/vanilla/advanced.html
* After: https://media-chrome-git-fork-luwes-fix-poster-mux.vercel.app/examples/vanilla/advanced.html

Using dev tools, disabled cache and throttled network to slow 3g. Once controls rendered, attempted to play using play button.
* Before: poster image would fade with controls
* After: poster image stays rendered until the media has loaded.
🎉

cjpillsbury (approved):
🚀

PR #825, opened 2024-02-15
fix: buggy chapter segments rendering

luwes:
don't love this but I couldn't get it to work otherwise.

### DIFF
--- a/src/js/controller.js
+++ b/src/js/controller.js
@@ -565,7 +565,15 @@ export const MediaUIStates = {
         .map((/** @type VTTCue */{ text, startTime, endTime }) => ({ text, startTime, endTime }));
     },
     mediaEvents: ['loadstart', 'loadedmetadata'],
-    textTracksEvents: ['addtrack', 'removetrack'],
+    textTracksEvents: ['addtrack', 'removetrack', 'change'],
+    mediaSetCallback(media, callback) {
+      const chaptersTrack = media.querySelector('track[kind="chapters"][default][src]');
+      chaptersTrack?.addEventListener('load', callback);
+    },
+    mediaUnsetCallback(media, callback) {
+      const chaptersTrack = media.querySelector('track[kind="chapters"][default][src]');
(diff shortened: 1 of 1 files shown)


src/js/controller.js
```diff
@@ -564,8 +564,12 @@ export const MediaUIStates = {
       return Array.from(chaptersTrack?.cues  ?? [])
         .map((/** @type VTTCue */{ text, startTime, endTime }) => ({ text, startTime, endTime }));
     },
-    mediaEvents: ['loadstart', 'loadedmetadata'],
-    textTracksEvents: ['addtrack', 'removetrack'],
+    mediaEvents: ['loadstart'],
+    textTracksEvents: ['addtrack', 'removetrack', 'change'],
```
luwes, line 568: I added `change` as well because it makes sense to re-render when the active chapters track changes

cjpillsbury: cool def makes sense to me.

src/js/controller.js
```diff
@@ -565,7 +565,15 @@ export const MediaUIStates = {
         .map((/** @type VTTCue */{ text, startTime, endTime }) => ({ text, startTime, endTime }));
     },
     mediaEvents: ['loadstart', 'loadedmetadata'],
-    textTracksEvents: ['addtrack', 'removetrack'],
+    textTracksEvents: ['addtrack', 'removetrack', 'change'],
+    mediaSetCallback(media, callback) {
+      const chaptersTrack = media.querySelector('track[kind="chapters"][default][src]');
+      chaptersTrack?.addEventListener('load', callback);
+    },
+    mediaUnsetCallback(media, callback) {
```
cjpillsbury, line 573: lol started peeping this before I got pulled away and looks like you anticipated the cleanup concern already. 🔥

cjpillsbury (approved):
Some concerns/tradeoffs with this approach, but we'll address those in a followup.

PR #829, opened 2024-02-18
feat: shifting time box elements with arrow

luwes:
test url: https://media-chrome-docs-git-fork-luwes-range-box-arrow-mux.vercel.app/docs/en/components/media-time-range

- [x] adds logic so children in the time range boxes shift closer to the time position, works for the preview and current box
- [x] adds a built-in arrow which also shifts similarly to `media-time-display` and `media-preview-time-display`


cjpillsbury (approved):
🚀

PR #850, opened 2024-03-11
fix: clean up media-time-display conditions for duration vs. seekable…

cjpillsbury:
…End usage (display and aria label).

Mostly relevant for "DVR" cases.


src/js/media-time-display.js
```diff
@@ -50,7 +55,12 @@ const DEFAULT_MISSING_TIME_PHRASE = 'video not loaded, unknown time.';
 const updateAriaValueText = (el) => {
   const currentTime = el.mediaCurrentTime;
   const [, seekableEnd] = el.mediaSeekable ?? [];
-  const endTime = el.mediaDuration || seekableEnd;
+  let endTime;
```
heff, line 58: We check for `endTime == null` later on, which makes me feel like this should now either default to null, or we should check for `undefined` or `!endTime` later.

heff (approved):
I don't honestly understand what the specific problem was. Maybe `Infinity` was sneaking through and causing problems? But it looks like a non problematic change.

PR #851, opened 2024-03-12
feat: React MediaStore

cjpillsbury:
NOTE: Per a conversation with @heff and prompted by @mmcc's gut example/pseudo-code on importing, at least at this stage, we decided to call an audible and not create multiple npm packages / restructure our codebase to a monorepo. As this was decided between just the two of us, we can certainly discuss more. Additionally, we can merge this as is but hold off on any "official support" and corresponding comms/documentation unless/until we have sufficient consensus on single package vs. multi-package for the new react-specific code, especially since this should have no impact on prior use cases of Media Chrome.

Demo usage/smoke testing: https://media-chrome-demo-nextjs-git-fork-cjpillsbury-experi-b8bc5e-mux.vercel.app/


scripts/react/build.js
```diff
@@ -34,7 +34,7 @@ const toReactComponentStr = (config) => {
   const ReactComponentName = toPascalCase(elementName);
   return `/** @type { import("react").HTMLElement } */
 const ${ReactComponentName} = React.forwardRef(({ children = [], ...props }, ref) => {
-  return React.createElement('${elementName}', toNativeProps({ ...props, ref }), children);
+  return React.createElement('${elementName}', toNativeProps({ ...props, suppressHydrationWarning: true, ref }), children);
```
cjpillsbury, line 37: Adding this to mitigate errors in SSR/RSC scenarios.

examples/nextjs-with-typescript/.gitignore
```diff
@@ -4,6 +4,7 @@
 /node_modules
```
cjpillsbury, line 4: Upgraded example app to Next.JS 13.

package.json
```diff
@@ -79,14 +86,15 @@
...
+    "@types/react": "^18.2.65",
     "@web/test-runner": "^0.15.0",
     "@web/test-runner-playwright": "^0.9.0",
     "cem-plugin-vs-code-custom-data-generator": "^1.4.1",
     "esbuild": "^0.20.0",
     "eslint": "^8.25.0",
     "npm-run-all": "^4.1.5",
     "prettier": "^2.7.1",
-    "react": "^17.0.2",
+    "react": "^18.2.0",
```
cjpillsbury, line 97: Current impl of react hooks for `MediaStore` requires React v18.0 or higher.

src/js/react/useSyncExternalStoreWithSelector.ts
```diff
@@ -0,0 +1,143 @@
...
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+import {
+  useRef,
+  useEffect,
+  useMemo,
+  useDebugValue,
+  useSyncExternalStore,
```
cjpillsbury, line 14: This is the reason we have a hard dependency on React 18 for this implementation. Targeting backwards compatibility before React 18 will require additional, non-trivial complexity.

tsconfig.json
```diff
@@ -9,6 +9,7 @@
     "module": "es2020",
     "allowJs": true,
     "checkJs": true,
+    "jsx": "react",
```
cjpillsbury, line 12: Added to handle new tsx for react-specific code.

package.json
```diff
@@ -57,8 +64,8 @@
     "premanifest": "cp -R src/js/ dist/",
     "manifest": "npx @custom-elements-manifest/analyzer analyze --config scripts/custom-elements-manifest.config.js",
     "build:types": "tsc",
-    "build:esm": "esbuild \"src/js/**/*.js\" --target=es2019 --format=esm --outdir=dist",
-    "build:cjs": "esbuild \"src/js/**/*.js\" --target=es2019 --format=cjs --outdir=dist/cjs",
+    "build:esm": "esbuild \"src/js/**/*.js\" src/js/react/**/* --target=es2019 --format=esm --outdir=dist",
```
cjpillsbury, line 67: These add the react-specific modules to the build process for CJS & ESM (but not for our bundled version).

package.json
```diff
@@ -24,10 +24,17 @@
       "default": "./dist/index.js"
     },
     "./react": {
+      "types": "./dist/react/index.d.ts",
       "import": "./dist/react/index.js",
       "require": "./dist/cjs/react/index.js",
       "default": "./dist/react/index.js"
     },
+    "./react/media-store": {
```
cjpillsbury, line 32: This allows us to directly import the react hooks et al. module without including `/dist` in the pathing for most modern build setups (tl;dr - they need to "know about" [package entry paths/"exports"](https://nodejs.org/api/packages.html#package-entry-points)). There are examples of this in the updated examples applications.

src/js/react/useSyncExternalStoreWithSelector.ts
```diff
@@ -0,0 +1,143 @@
+/**
```
cjpillsbury, line 1: Per the code comment, this is a port (mostly copy) of the official/published https://github.com/facebook/react/blob/main/packages/use-sync-external-store/src/useSyncExternalStoreWithSelector.js from Facebook/Meta. By porting this (which is highly unlikely to change), we sidestep adding additional dependencies to Media Chrome that are only needed for the react-specific module. This is unlike our general `react` module dependency, as we can assume folks using our react code already also have a dependency on react in their application, library, or the like.

examples/vite-react-with-typescript/.eslintrc.cjs
```diff
@@ -0,0 +1,18 @@
+module.exports = {
```
cjpillsbury, line 1: Added a vite project to allow us to test standard setups for client-side only React applications. vite has replaced "create react app" as the de facto build tool for client side react applications.

examples/nextjs-with-typescript/src/app/page.tsx
```diff
@@ -0,0 +1,23 @@
+import Link from 'next/link';
```
cjpillsbury, line 1: New landing page for the next example. Roughly analogous to [our next example app for elements](https://elements-demo-nextjs-mux.vercel.app/)

examples/nextjs-with-typescript/src/app/react-wrappers/player.tsx
```diff
@@ -0,0 +1,83 @@
+import {
```
cjpillsbury, line 1: Minor refactor of our original example app. Note that this new version is a fully server-side rendered component with the new Next 13 component architecture.

TODO: Re-add toggle button to mount/dismount the controller somewhere (which cannot be defined as an RSC) or add an additional page for this to ensure we can still easily smoke test for e.g. GC concerns.

examples/nextjs-with-typescript/package.json
```diff
@@ -1,23 +1,28 @@
...
-    "start": "next start -p 3001",
+    "start": "next start -p 8087",
     "lint": "next lint"
   },
   "dependencies": {
     "media-chrome": "link:../../",
-    "next": "12.0.10",
-    "react": "17.0.2",
-    "react-dom": "17.0.2"
+    "react": "link:../../node_modules/react",
```
cjpillsbury, line 13: NOTE: using links to sidestep acrobatics/build errors that result from duplicate versions of react being pointed to. This is roughly analogous to complexities we encounter when doing a `yarn link` from "react flavored" elements for other projects such as stream.new. While we could alternatively rely on the [yarn file protocol](https://yarnpkg.com/protocol/file) for our media chrome dependency, this has the disadvantage of copying instead of symlinking, which means changes in media chrome won't get picked up without a dependency reinstall on the next example app.

examples/vite-react-with-typescript/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "type": "module",
+  "scripts": {
+    "dev": "vite",
+    "build": "tsc && vite build",
+    "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
+    "preview": "vite preview"
+  },
+  "dependencies": {
+    "media-chrome": "link:../../",
+    "react": "link:../../node_modules/react",
```
cjpillsbury, line 14: See comment for the equivalent declaration in the next.js example app's package json for an explanation of this.

examples/vite-react-with-typescript/src/App.tsx
```diff
@@ -0,0 +1,264 @@
...
+  Provider as MediaChromeProvider,
+  useMediaDispatch as useDispatch,
+  useFullscreenRefCallback,
+  useMediaRefCallback,
+  useMediaSelector as useSelector,
+} from 'media-chrome/react/media-store';
+import { constants } from 'media-chrome';
+const { MediaUIEvents } = constants;
+
+const PlayButton = () => {
```
cjpillsbury, line 13: This is analagous to the next.js example of using the new react context+provider+hooks integrating with `MediaStore`

examples/nextjs-with-typescript/src/app/media-store-hooks/ReactPlayer.tsx
```diff
@@ -0,0 +1,258 @@
...
+// import { useEffect, useState } from 'react';
+import {
+  Provider as MediaChromeProvider,
+  useMediaDispatch as useDispatch,
+  useFullscreenRefCallback,
+  useMediaRefCallback,
+  useMediaSelector as useSelector,
+} from 'media-chrome/react/media-store';
+import { constants } from 'media-chrome';
+const { MediaUIEvents } = constants;
```
cjpillsbury, line 14: Should we consider re-exporting this from `/react/media-store` and renaming to e.g. `MediaActionTypes`?

src/js/react/media-store.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+// These are hooks, which are the norm for working in "modern React". You can see them in action
+// in the PlayButton component and the Video component
+
+// This is a hook to get access to the media ui store's dispatch method, which allows the component to make
+// media state change requests. All it does is grab the context (with the "inherited" value provided from
+// the ancestor Provider) and return the dispatch. Simple.
+export const useMediaDispatch = () => {
+  const store = useContext(MediaContext);
+  const dispatch = store?.dispatch ?? console.log.bind(null, 'fake dispatch!');
```
AdamJaggard, line 62: Just wondering what the scenarios are where the console log fallback will activate, maybe if they use the dispatch hook but haven't added the provider to their tree yet?

examples/nextjs-with-typescript/src/app/media-store-hooks/ReactPlayer.tsx
```diff
@@ -0,0 +1,258 @@
+'use client';
```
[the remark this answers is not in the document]
luwes: is this a jedi mind trick :) it says 'use client' right above :D

package.json
```diff
@@ -57,8 +64,8 @@
     "premanifest": "cp -R src/js/ dist/",
     "manifest": "npx @custom-elements-manifest/analyzer analyze --config scripts/custom-elements-manifest.config.js",
     "build:types": "tsc",
-    "build:esm": "esbuild \"src/js/**/*.js\" --target=es2019 --format=esm --outdir=dist",
-    "build:cjs": "esbuild \"src/js/**/*.js\" --target=es2019 --format=cjs --outdir=dist/cjs",
+    "build:esm": "esbuild \"src/js/**/*.js\" src/js/react/**/* --target=es2019 --format=esm --outdir=dist",
+    "build:cjs": "esbuild \"src/js/**/*.js\" src/js/react/**/* --target=es2019 --format=cjs --outdir=dist/cjs",
```
luwes, line 68: esbuild added support for globs recently so it shouldn't be needed

src/js/react/media-store.tsx
```diff
@@ -0,0 +1,116 @@
+import { createContext, useContext, useEffect, useMemo } from 'react';
```
luwes, line 1: hmm, I think for now we were using TS jsdocs right?

does this file contain jsx or will it in the future?

luwes: oh seeing `<MediaContext.Provider` now, maybe for that one vnode maybe we can use createElement directly?

examples/nextjs-with-typescript/src/app/media-store-hooks/ReactPlayer.tsx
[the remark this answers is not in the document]
cjpillsbury: whoops that was some copypasta. This can't be (fully) a server side component because of the hooks in play. Amending prior comment.

src/js/react/media-store.tsx
cjpillsbury: Any reason why? I'm not seeing any clear advantages to your proposal.

src/js/react/media-store.tsx
cjpillsbury: Yeah I believe this was a timing issue? I def added it while testing. Good callout though I should at least make this less of a toy fallback.

package.json
cjpillsbury: Happy to refactor to whatever you're proposing here, though I'm not sure what the concern is?

examples/nextjs-with-typescript/src/app/react-wrappers/web-component-registration.tsx
```diff
@@ -0,0 +1,8 @@
+'use client';
+import 'media-chrome';
+
+// Clever little hack way to allow our web components' usage to be RSCs (instead of client side components) but ensure
+// the code that needs to run in the browser (e.g. the web component definitions and registration) is actually
+// included & executed on the client side. See usage in player.tsx for complete example. (CJP)
+const DummyComponent = () => <></>
+export default DummyComponent;
```
decepulis, line 8: Would this component still work without these two lines?

cjpillsbury: removed based on larger discussion, but thanks for the knowledge transfer in our meeting!

src/js/react/media-store.tsx
```diff
@@ -0,0 +1,116 @@
...
+} from '../media-store/media-store';
+
+// The context is the thing that allows for a shared... context that propagates outside of the (V)DOM/Component tree.
+// Contexts have values. Our value will be the the media ui store (aka the "guts" of Media Controller state mgmt)
+export const MediaContext: Context<MediaStore | null> =
+  createContext<MediaStore | null>(null);
+
+// The Provider is a (non-visual) React Element that... provides a context for any descendant in the Component tree that may
+// want to use it. As such, it must be an ancestor in the Component Tree for anything that wants to "consume" the context.
+export const Provider = ({
```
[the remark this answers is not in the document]
cjpillsbury: Talked through in meeting. Will be pushing up with discussed names ASAP. 🎉 good callout, @AdamJaggard

package.json
luwes: `\"src/js/**/*.js\" src/js/react/**/*` could be `\"src/js/**/*\"

luwes (approved):
looks good!

AdamJaggard (approved):
🚢

PR #856, opened 2024-03-18
fix: monitor for extended 'seekablechange' event for more reliable me…

cjpillsbury:
…diaSeekable updates.

For an example implementation, see: https://github.com/muxinc/elements/pull/883


luwes (approved):
👍

PR #859, opened 2024-03-18
fix: account for some dynamic changes to media-time-range

cjpillsbury:
(mostly relevant for DVR).

### DIFF
--- a/src/js/media-time-range.js
+++ b/src/js/media-time-range.js
@@ -426,7 +426,10 @@ class MediaTimeRange extends MediaChromeRange {
       this.updateBufferedBar();
     }
 
-    if (attrName === MediaUIAttributes.MEDIA_DURATION) {
+    if (
+      attrName === MediaUIAttributes.MEDIA_DURATION ||
+      attrName === MediaUIAttributes.MEDIA_SEEKABLE
+    ) {
       this.mediaChaptersCues = this.#mediaChaptersCues;
       this.updateBar();
     }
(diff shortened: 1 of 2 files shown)


src/js/utils/range-animation.js
```diff
@@ -38,7 +38,8 @@ export class RangeAnimation {
     // 1. Always allow increases.
     // 2. Allow a relatively large decrease (user action or Safari jumping back :s).
     const increase = start - this.#range.valueAsNumber;
-    if (increase > 0 || increase < -.03) {
+    const durationDelta = Math.abs(duration - this.duration);
```
cjpillsbury, line 41: need to also make sure we're updating if the `duration` value has changed "meaningfully".

src/js/media-time-range.js
```diff
@@ -766,7 +769,7 @@ class MediaTimeRange extends MediaChromeRange {
       return;
     }
 
-    const duration = this.mediaDuration;
+    const duration = this.mediaSeekableEnd;
```
cjpillsbury, line 772: prefer `mediaSeekableEnd` here to make sure we're not early bailing for DVR cases.

src/js/media-time-range.js
```diff
@@ -445,7 +448,7 @@ class MediaTimeRange extends MediaChromeRange {
       && !this.mediaPaused
       && !this.mediaLoading
       && !this.mediaEnded
-      && this.mediaDuration > 0
+      && this.mediaSeekableEnd > 0
```
cjpillsbury, line 451: NOTE: Although `mediaDuration` would probably be fine here, updating to `mediaSeekableEnd` since that's the value that's actually driving the updates under most cases.

luwes (approved):
looks good!

PR #860, opened 2024-03-19
fix: await custom media defined

### LINKED ISSUE #857: custom media element script order breaks media-chrome v3
I got it, it only works if you place the media-chrome script tag *after* mux or hls-video-element's tag, which is the opposite of the example [here](https://www.media-chrome.org/docs/en/media-elements/hls-video). It didn't seem to matter before 3.0.0, is this intended? Thank you for your help.

_Originally posted by @xrun1 in https://github.com/muxinc/media-chrome/discussions/845#discussioncomment-8713403_

luwes:
fix #857


cjpillsbury (approved):
NICE FIND!

PR #889, opened 2024-05-03
fix: avoid dist artifacts in manifest generation

cjpillsbury:


package.json
```diff
@@ -61,7 +61,7 @@
     "format": "prettier --write .",
     "clean": "rimraf dist",
     "lint": "eslint src/js",
-    "premanifest": "cp -R src/js/ dist/",
+    "premanifest": "mkdir -p dist/ && cp -R src/js/media-* dist/",
```
cjpillsbury, line 64: make the `/dist/` directory in case it doesn't already exist (but `-p` to not yell if it does) + use same `cp` target as manifest glob match.

luwes (approved):

PR #892, opened 2024-05-07
Fix React docs link in README.md

ADTC:
Existing link gives 404 error. Change to the Get Started page instead.


luwes (approved):
Thank you!

PR #914, opened 2024-05-22
task/ts-conversion

littlespex:
This is an experiment in using Copilot and other tools to convert the code base to typescript. Remaining tasks:

- [ ] Convert remaining JS files
- [x] Convert tests
- [ ] Verify all tests pass
- [ ] Ensure docs are generated correctly


dylanjha:
@littlespex amazing stuff!

First pass I'm noticing a lot of linting-ish diffs. Spacing, quotes, etc. Stuff like that. We do have a `yarn lint` command and I do see in the commit messages either you or copilot workspace was making lint commits.

Did our existing eslint config not get applied? Or was the eslint config modified as part of this change?

littlespex:
Those are the result of running the `format` script.

cjpillsbury:
> @dylanjha @heff Does anyone know how to address this Playwright error: https://github.com/muxinc/media-chrome/actions/runs/9289849038/job/25566542757#step:10:23
> 
> The `test --all` command works for me locally, but something with the install of playwright is preventing CI from completing. I did update the `package.json` and `yarn.lock`, so I want to make sure these errors weren't introduced by this PR.

I'm re-running now just in case. I can pull your branch down and test locally as well.

cjpillsbury:
@littlespex looking like the buckshot updates did cause some dependency issues. Will require some digging.

cjpillsbury:
@littlespex looks like it's a yarn + playwright binaries issue. https://github.com/microsoft/playwright/issues/5767 Since we had some dependencies still pointing to npmjs registry that were updated to use the yarn registry, I'm wondering that is the root cause. (We can also go down the path of doing the env setup workarounds if that makes sense)

dylanjha:
> Those are the result of running the format script. The code generated by the tools had some sloppy formatting, so I ran npm run format to normalize it. Is that script still valid?

@littlespex yeah, the `format` script should still be valid. But something seems off because it runs prettier and our prettier config specifies single quotes: https://github.com/muxinc/media-chrome/blob/692be61a2a63ba98c7e449d860ce1828b78353a2/package.json#L106-L110

But for some reason this diff is showing a lot of conversions from single quotes to double quotes. That's just one example, but the broader point is that mis-matched linting & formatting is generating a larger diff than necessary.

littlespex:
Re-ran prettier and the quotes are fixed. There are other formatting changes, like print width, that were also applied using prettier's default values.d

src/js/media-chrome-button.ts
```diff
@@ -92,23 +94,29 @@ template.innerHTML = /*html*/ `
  * @cssproperty --media-button-icon-transform - `transform` of button icon.
  * @cssproperty --media-button-icon-transition - `transition` of button icon.
  */
-class MediaChromeButton extends globalThis.HTMLElement {
-  #mediaController;
+class MediaChromeButton extends CustomElement {
```
littlespex, line 97: The `CustomElement` class does *not* use the server safe `globalThis`. This is because when that version of the class is used, typescript no longer sees the components as extensions of `HTMLElement`, and marks any reference to `this` in the class as invalid. Some investigation is needed into how to get around these errors, or if there is a better way to do "sever safe globals".

cjpillsbury: I had some fights against this while I was working on the media-store abstraction. I'm starting to suspect we should do this iteratively, starting with things like constants and server-safe-globals and working our way out. We're definitely in a tricky TS situations with the node/server vs. browser straddling.

littlespex: What is the exact problem that `server-safe-globals.js` solves? (is there an issue and/or PR?)

cjpillsbury: Ok a sub-optimal compromise I'm experimenting with in a local branch is "lying" with our types in our our server-safe-globals definitions. Ideally we'll follow up to make sure that we don't end up with runtime errors for valid server-side usage, but this feels like the right compromise to accelerate/unblock the TS migration effort.

I'll push up a branch for you to see the general approach there. Might make sense for us to touch base on how to combine your effort with my changes.

littlespex: I just ran a test, removed server-safe-globals completely from the source files, only use it in `build/react.js`, and both example apps build without issue.

littlespex: https://github.com/littlespex/media-chrome/pull/4

cjpillsbury: I promise these are necessary shims.

PR #920, opened 2024-05-30
build: lint/format et al updates

cjpillsbury:


package.json
```diff
@@ -58,7 +58,7 @@
     "dist"
   ],
   "scripts": {
-    "format": "prettier --write .",
+    "format": "prettier './src/js/**/*.{js,ts,tsx}' './test/**/*.{js,ts,tsx}' --write",
```
cjpillsbury, line 61: for now, let's only target tests and src

cjpillsbury:
cc @littlespex for visibility

heff (approved):

PR #925, opened 2024-06-12
feat:typescript migration

cjpillsbury:
baseline for the migration efforts


src/js/constants.ts
```diff
@@ -0,0 +1,242 @@
+export const MediaUIEvents = {
```
cjpillsbury, line 1: tried to keep these as tight as possible to make derivation of types automatic and easier to maintain as we add things.

src/js/utils/server-safe-globals.ts
```diff
@@ -79,59 +84,18 @@ const globalThisShim = {
...
- * exitPictureInPicture?,
- * pictureInPictureEnabled?,
- * requestPictureInPicture?,
- * addEventListener?,
- * removeEventListener?,
- * } }
- */
-export const Document =
-  isServer && !isShimmed ? documentShim : globalThis.document;
+export const GlobalThis: typeof globalThis =
```
cjpillsbury, line 94: updated these to "lie" on their export types as if they are always running in the browser and not minimal shim impls when on e.g. node. That does mean we won't get type errors for cases where we have failed to shim/polyfill.

src/js/media-chrome-button.ts
```diff
@@ -95,19 +95,25 @@ template.innerHTML = /*html*/ `
 class MediaChromeButton extends globalThis.HTMLElement {
```
cjpillsbury, line 95: I migrated this one component as part of proving out/validating any issues.

src/js/experimental/media-chrome-option.js
```diff
@@ -109,7 +109,7 @@ class MediaChromeOption extends globalThis.HTMLElement {
 
   enable() {
     if (!this.hasAttribute('tabindex')) {
-      this.setAttribute('tabindex', -1);
+      this.tabIndex = -1;
```
heff, line 112: This is safe, right?

src/js/media-captions-button.js
```diff
@@ -155,4 +155,5 @@ if (!globalThis.customElements.get('media-captions-button')) {
   );
 }
 
+export { MediaCaptionsButton };
```
heff, line 158: Why is this needed?

heff (approved):
A few questions but looks good to me. There's some dark magic in the contants types I don't fully grok if it's worth getting another set of eyes on.

src/js/experimental/media-chrome-option.js
cjpillsbury: should be, but I'll double-confirm 😅

cjpillsbury: yup confirmed!

src/js/media-captions-button.js
cjpillsbury: Plausibly not needed if we migrate to full TypeScript. The `@typedef` imports were getting confused from using the default module export. This is also a pattern in some other of our component modules. There are good reasons to do named exports/both (tl;dr - different build tools can get confused). Might be a good candidate for our development principles (in whatever direction we decide).

src/js/media-chrome-button.ts
littlespex: The issue is that when the utility functions that except `HTMLElement`s are converted to TS, they still don't consider these true `HTMLElement`s:

Screenshot 2024-06-17 at 16 56 34

littlespex: Do we know of a specific environment/configuration that breaks without `server-safe-globals.js`? I have a few ideas for workarounds, but I have no way to verify them without a test case.

luwes: yep, a Next.js app should reproduce the issues that come up without patching the globals.
an example here https://github.com/muxinc/media-chrome/tree/main/examples/nextjs-with-typescript

littlespex: The next's doesn't need the globals at all:

https://github.com/littlespex/media-chrome/pull/4

littlespex:
@cjpillsbury I got permissions errors when trying to commit to this branch, or create a new one, so I copied all of this over to my fork and used it as a base for https://github.com/muxinc/media-chrome/pull/931.

luwes:
closed by https://github.com/muxinc/media-chrome/pull/931

PR #931, opened 2024-06-25
feat: convert JS files to typescript

littlespex:
Typescript migration based off of https://github.com/muxinc/media-chrome/pull/925


littlespex:
@luwes @cjpillsbury This PR has all of the files converted to TS, and works with server globals.

src/js/experimental/media-audio-track-listbox.ts
```diff
@@ -93,7 +98,7 @@ class MediaAudioTrackListbox extends MediaChromeListbox {
     }
   }
 
-  #onChange() {
```
luwes, line 96: was this required? 

I always try to keep the methods on the prototype if possible. 
setting this to a closure on each instance will be less optimized...

(I know in some cases it's needed and using bind sucks)

src/js/experimental/media-captions-listbox.ts
```diff
@@ -129,15 +134,15 @@ class MediaCaptionsListbox extends MediaChromeListbox {
...
-      if (type === 'captions') {
+      if ((type as string) === 'captions') {
         option.append(createIndicator(this, 'captions-indicator'));
       }
 
       container.append(option);
     }
   }
 
-  #onChange() {
```
luwes, line 140: same here, if this change is to satisfy types let me know. I'll help try to find a way so the code doesn't need to change

src/js/experimental/media-audio-track-listbox.ts
littlespex: That was a change made by the AI. It probably noticed that the function is used as an event handler and is not bound. It may not matter as the handler is attached the calling object, so `this` should resolve to the component instance.

src/js/media-audio-track-menu.ts
```diff
@@ -11,11 +11,10 @@ import {
   setStringAttr,
   getMediaController,
 } from './utils/element-utils.js';
+import { TextTrackLike } from './utils/TextTrackLike.js';
 
 /**
- * @extends {MediaChromeMenu}
```
luwes, line 16: this might be needed for custom-elements-manifest/analyzer.

src/js/media-audio-track-menu.ts
```diff
@@ -116,7 +119,7 @@ class MediaAudioTrackMenu extends MediaChromeMenu {
     }
   }
 
-  #onChange() {
+  #onChange = (): void => {
```
luwes, line 122:
```suggestion
  #onChange() {
```

src/js/media-audio-track-menu.ts
```diff
@@ -11,11 +11,10 @@ import {
   setStringAttr,
   getMediaController,
 } from './utils/element-utils.js';
+import { TextTrackLike } from './utils/TextTrackLike.js';
 
 /**
```
littlespex, line 16:
```suggestion
/**
 * @extends {MediaChromeMenu}
```

src/js/media-chrome-menu-item.ts
```diff
@@ -4,10 +4,9 @@ import {
   getDocumentOrShadowRoot,
   containsComposedNode,
 } from './utils/element-utils.js';
+import MediaChromeMenu from './media-chrome-menu.js';
```
luwes, line 7:
```suggestion
import type MediaChromeMenu from './media-chrome-menu.js';
```

src/js/media-audio-track-menu.ts
```diff
@@ -32,69 +33,73 @@ class MediaAudioTrackMenu extends MediaChromeMenu {
...
+    newValue: string | null
+  ): void {
     super.attributeChangedCallback(attrName, oldValue, newValue);
 
     if (
       attrName === MediaUIAttributes.MEDIA_AUDIO_TRACK_ENABLED &&
       oldValue !== newValue
     ) {
-      this.value = newValue;
+      this.value = newValue ?? '';
```
luwes, line 50: this.value == null is checked later on so I think best to leave this null or undefined

src/js/media-rendition-menu.ts
```diff
@@ -11,10 +11,9 @@ import {
   createMenuItem,
   createIndicator,
 } from './media-chrome-menu.js';
+import { Rendition } from './media-store/state-mediator.js';
 
 /**
- * @extends {MediaChromeMenu}
```
luwes, line 16: also best to keep I think

src/js/media-store/request-map.ts
```diff
@@ -11,35 +11,18 @@ import {
...
- *   [K in keyof StateMediator]: ReturnType<StateMediator[K]['get']>
- * } & {
- *   mediaPreviewTime: number;
- *   mediaPreviewImage: string;
- *   mediaPreviewCoords: [number,number,number,number]
- * }} MediaState
- */
+export type MediaUIEventsType =
+  typeof MediaUIEvents[keyof typeof MediaUIEvents];
+export type MediaRequestTypes = Exclude<
```
luwes, line 19: is this correct? Omit to Exclude and uppercast to lowercase

src/js/utils/server-safe-globals.ts
```diff
@@ -0,0 +1,108 @@
...
+  /**
+   * @param {string} media
+   */
+  matchMedia(media) {
+    return {
+      matches: false,
+      media,
+    };
+  },
+} as unknown as typeof globalThis;
```
luwes, line 94: 👍

src/js/utils/template-parts.ts
```diff
@@ -37,37 +51,37 @@ export const defaultProcessor = {
 };
 
 /**
- * @extends {DocumentFragment}
+ *
  */
-export class TemplateInstance extends globalThis.DocumentFragment {
+export class TemplateInstance extends DocumentFragment {
```
luwes, line 56:
```suggestion
export class TemplateInstance extends globalThis.DocumentFragment {
```

src/js/media-store/request-map.ts
littlespex: Yes this is correct. The `requestMap` below uses the values of `MediaUIEvents` as keys, and `MediaUIEventsType` is a union of all the `MediaUIEvents` values.

luwes:
congrats so far, what a massive PR! 

nothing too broken that I can see,
only I'm seeing an issue with controlbar show / hide on player load
https://media-chrome-git-fork-littlespex-task-ts-migration-mux.vercel.app/examples/vanilla/advanced.html

I'll look into this as well. some state issue

src/js/media-container.ts
```diff
@@ -604,24 +618,24 @@ class MediaContainer extends globalThis.HTMLElement {
   #scheduleInactive() {
     this.#setActive();
 
-    clearTimeout(this._inactiveTimeout);
+    clearTimeout(this.#inactiveTimeout);
 
     // Setting autohide to -1 turns off autohide
-    if (this.autohide < 0) return;
+    if (this.#autohide < 0) return;
```
luwes, line 624:
```suggestion
    if (this.autohide < 0) return;
```

src/js/media-container.ts
```diff
@@ -604,24 +618,24 @@ class MediaContainer extends globalThis.HTMLElement {
...
 
     /** @type {ReturnType<typeof setTimeout>} */
-    this._inactiveTimeout = setTimeout(() => {
+    this.#inactiveTimeout = setTimeout(() => {
       this.#setInactive();
-    }, this.autohide * 1000);
+    }, this.#autohide * 1000);
```
luwes, line 629:
```suggestion
    }, this.autohide * 1000);
```

luwes (approved):
awesome! LGTM! thank you for the huge contribution!

luwes:
maybe one nit is I think there's a few places the types can be inferred, we'd probably want to prefer that if possible down the road to make the code less noisy. not a blocker tough.

littlespex:
If you have some specific examples I can take a look. One thing to mention is that declarations were added to all public facing interfaces to be compatible with the [isolated-declarations](https://devblogs.microsoft.com/typescript/announcing-typescript-5-5/#isolated-declarations) compiler flag.

PR #941, opened 2024-07-10
feat!: Tooltips

### LINKED ISSUE #411: Add tooltips to all icon controls
It would be great if media-chrome would ship with a default implementation of tooltips for all of the iconography used for controls.

This could either be a default Browser Chrome implementation (using `title=foo` attributes) or a custom tooltip.

Some examples of what I mean;

Youtube
image

Vimeo
image

I think this would greatly help some less digital-savvy visitors with understanding how to use players built with media-chrome.

AdamJaggard:
Adds a new `media-tooltip` element and includes tooltips by default in anything that extends the `media-chrome-button` class. Tooltips can be placed either `top`, `right`, `bottom`, `left` or `none`.

fix #411

### DIFF
--- a/examples/vanilla/tooltips.html
+++ b/examples/vanilla/tooltips.html
@@ -57,11 +57,11 @@
       <h2>On their own (no arrows when <code>position="none"</code>)</h2>
 
       <div>
-        <media-tooltip position="none">Settings</media-tooltip>
+        <media-tooltip placement="none">Settings</media-tooltip>
         <br><br>
-        <media-tooltip position="none">Play</media-tooltip>
+        <media-tooltip placement="none">Play</media-tooltip>
         <br><br>
-        <media-tooltip position="none">Pause</media-tooltip>
+        <media-tooltip placement="none">Pause</media-tooltip>
       </div>
 
       <h2>As part of button controls</h2>
@@ -106,19 +106,19 @@
       <div style="display: flex; gap: 25px;">
         <div>
           <div>Top</div>
-          <media-play-button tooltipposition="top"></media-play-button>  
+          <media-play-button tooltipplacement="top"></media-play-button>  
         </div>
         <div>
           <div>Right</div>
-          <media-play-button tooltipposition="right"></media-play-button>  
+          <media-play-button tooltipplacement="right"></media-play-button>  
         </div>
         <div>
           <div>Bottom</div>
-          <media-play-button tooltipposition="bottom"></media-play-button>  
+          <media-play-button tooltipplacement="bottom"></media-play-button>  
         </div>
         <div>
           <div>Left</div>
-          <media-play-button tooltipposition="left"></media-play-button>
+          <media-play-button tooltipplacement="left"></media-play-button>
         </div>
       </div>
 
@@ -202,7 +202,7 @@
 
       <h2>Placing within a custom button</h2>
       <div style="display: flex;">
-        <div class="custom-btn">+<media-tooltip position="right">Add</media-tooltip></div>
+        <div class="custom-btn">+<media-tooltip placement="right">Add</media-tooltip></div>
       </div>
 
 	    <div class="examples">
--- a/examples/vanilla/tooltips.html
+++ b/examples/vanilla/tooltips.html
@@ -38,6 +38,10 @@
         <span slot="tooltip">Play the <strong>damn</strong> movie</span>
       </media-play-button>
 
+      <h2>Hiding with CSS</h2>
+
+      <media-play-button style="--media-tooltip-display: none;"></media-play-button>
+
 	    <div class="examples">
         <a href="./">View more examples</a>
       </div>
--- a/examples/vanilla/tooltips.html
+++ b/examples/vanilla/tooltips.html
@@ -25,6 +25,29 @@
       code {
         color: maroon;
       }
+
+      .custom-btn {
+        position: relative;
+        width: 40px;
+        height: 40px;
+        border-radius: 50%;
+        background: lightblue;
+        border: 2px solid #333;
+        color: #333;
+        display: flex;
+        justify-content: center;
+        align-items: center;
+        font-size: 24px;
+        font-weight: bold;
+        cursor: pointer;
+      }
+      .custom-btn media-tooltip {
+        opacity: 0;
+        transition: opacity .3s;
+      }
+      .custom-btn:hover media-tooltip {
+        opacity: 1;
+      }
     </style>
   </head>
   <body>     
@@ -176,6 +199,11 @@
           "></media-play-button>
         </media-control-bar>
       </media-controller>
+
+      <h2>Placing within a custom button</h2>
+      <div style="display: flex;">
+        <div class="custom-btn">+<media-tooltip position="right">Add</media-tooltip></div>
+      </div>
 
 	    <div class="examples">
         <a href="./">View more examples</a>
--- a/examples/vanilla/tooltips.html
+++ b/examples/vanilla/tooltips.html
@@ -27,11 +27,11 @@
       <h2>On their own</h2>
 
       <div>
-        <media-tooltip>Settings</media-tooltip>
+        <media-tooltip position="none">Settings</media-tooltip>
         <br><br>
-        <media-tooltip>Play</media-tooltip>
+        <media-tooltip position="none">Play</media-tooltip>
         <br><br>
-        <media-tooltip>Pause</media-tooltip>
+        <media-tooltip position="none">Pause</media-tooltip>
       </div>
 
       <h2>As part of button controls</h2>
@@ -65,6 +65,27 @@
       <h2>Hiding with CSS</h2>
 
       <media-play-button style="--media-tooltip-display: none;"></media-play-button>
+
+      <h2>Positioning</h2>
+
+      <div style="display: flex; gap: 25px;">
+        <div>
+          <div>Top</div>
+          <media-play-button tooltipposition="top"></media-play-button>  
+        </div>
+        <div>
+          <div>Right</div>
+          <media-play-button tooltipposition="right"></media-play-button>  
+        </div>
+        <div>
+          <div>Bottom</div>
+          <media-play-button tooltipposition="bottom"></media-play-button>  
+        </div>
+        <div>
+          <div>Left</div>
+          <media-play-button tooltipposition="left"></media-play-button>
+        </div>
+      </div>
 
 	    <div class="examples">
         <a href="./">View more examples</a>
--- a/examples/vanilla/tooltips.html
+++ b/examples/vanilla/tooltips.html
@@ -54,7 +54,7 @@
         </media-control-bar>
       </media-controller>
 
-      <h2>Overriding tooltip label</h2>
+      <h2>Overriding tooltip label <small>(and constraining x position)</small></h2>
       <media-controller>
         <video
           slot="media"
--- a/examples/vanilla/tooltips.html
+++ b/examples/vanilla/tooltips.html
@@ -1,0 +1,39 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta name="viewport" content="width=device-width" />
+    <title>Media Chrome Tooltips</title>
+    <script type="module" src="../../dist/index.js"></script>
+    <style>
+      body {
+        background: #eee;
+      }
+      .examples {
+        margin-top: 20px;
+      }
+    </style>
+  </head>
+  <body>
+    <main>
+      <h1>Tooltips</h1>
+      
+      <h2>On their own</h2>
+
+      <div>
+        <media-tooltip>Settings</media-tooltip>
+        <br><br>
+        <media-tooltip>Play</media-tooltip>
+        <br><br>
+        <media-tooltip>Pause</media-tooltip>
+      </div>
+
+      <h2>As part of button controls</h2>
+
+      <media-play-button></media-play-button>
+
+	    <div class="examples">
+        <a href="./">View more examples</a>
+      </div>
+    </main>
+  </body>
+</html>
--- a/examples/vanilla/tooltips.html
+++ b/examples/vanilla/tooltips.html
@@ -154,7 +154,7 @@
             --media-primary-color: #0f0;
             --media-secondary-color: #333;
             --media-tooltip-background: #000;
-            --media-tooltip-arrow-background-color: #0f0;
+            --media-tooltip-arrow-color: #0f0;
             --media-tooltip-arrow-width: 20px;
             --media-tooltip-arrow-height: 5px;
             --media-tooltip-border: 1px solid #0f0;
--- a/examples/vanilla/tooltips.html
+++ b/examples/vanilla/tooltips.html
@@ -11,6 +11,9 @@
       .examples {
         margin-top: 20px;
       }
+      media-controller {
+        max-width: 500px;
+      }
     </style>
   </head>
   <body>
@@ -29,8 +32,17 @@
 
       <h2>As part of button controls</h2>
 
-      <media-play-button></media-play-button>
-      <media-airplay-button></media-airplay-button>
+      <media-controller>
+        <video
+          slot="media"
+          src="https://stream.mux.com/A3VXy02VoUinw01pwyomEO3bHnG4P32xzV7u1j1FSzjNg/high.mp4"
+          muted
+        ></video>
+        <media-control-bar>
+          <media-play-button></media-play-button>
+          <media-airplay-button></media-airplay-button>
+        </media-control-bar>
+      </media-controller>
 
       <h2>Overriding tooltip label</h2>
       <media-airplay-button></media-airplay-button>
--- a/examples/vanilla/tooltips.html
+++ b/examples/vanilla/tooltips.html
@@ -53,6 +53,33 @@
   <body>     
     <main>
       <h1>Tooltips</h1>
+
+      <h2>All buttons</h2>
+
+      <media-controller style="--media-tooltip-container-margin: 10px; min-width: 700px;">
+        <video
+          slot="media"
+          src="https://stream.mux.com/A3VXy02VoUinw01pwyomEO3bHnG4P32xzV7u1j1FSzjNg/high.mp4"
+          muted
+        ></video>
+        <media-control-bar style="display: flexbox; gap: 5px;">
+          <media-airplay-button></media-airplay-button>
+          <media-audio-track-menu-button></media-audio-track-menu-button>
+          <media-captions-menu-button></media-captions-menu-button>
+          <media-captions-button></media-captions-button>
+          <media-cast-button></media-cast-button>
+          <media-fullscreen-button></media-fullscreen-button>
+          <media-mute-button></media-mute-button>
+          <media-pip-button></media-pip-button>
+          <media-play-button></media-play-button>
+          <media-playback-rate-button></media-playback-rate-button>
+          <media-playback-rate-menu-button></media-playback-rate-menu-button>
+          <media-rendition-menu-button></media-rendition-menu-button>
+          <media-seek-backward-button></media-seek-backward-button>
+          <media-seek-forward-button></media-seek-forward-button>
+          <media-settings-menu-button></media-settings-menu-button>
+        </media-control-bar>
+      </media-controller>
       
       <h2>On their own (no arrows when <code>position="none"</code>)</h2>
 
--- a/examples/vanilla/tooltips.html
+++ b/examples/vanilla/tooltips.html
@@ -15,6 +15,9 @@
       media-controller {
         max-width: 300px;
       }
+      media-control-bar {
+        margin: 10px;
+      }
     </style>
   </head>
   <body>     
--- a/examples/vanilla/tooltips.html
+++ b/examples/vanilla/tooltips.html
@@ -61,7 +61,11 @@
           slot="media"
           src="https://stream.mux.com/A3VXy02VoUinw01pwyomEO3bHnG4P32xzV7u1j1FSzjNg/high.mp4"
           muted
-        ></video>
+        >
+          <track label="English" kind="captions" srclang="en" src="../vtt/elephantsdream/captions.en.vtt" default></track>
+        </video>
+        <media-playback-rate-menu anchor="auto" hidden></media-playback-rate-menu>
+        <media-captions-menu anchor="auto" hidden></media-captions-menu>
         <media-control-bar style="display: flex; gap: 5px;">
           <media-airplay-button></media-airplay-button>
           <media-audio-track-menu-button></media-audio-track-menu-button>
--- a/examples/vanilla/tooltips.html
+++ b/examples/vanilla/tooltips.html
@@ -7,16 +7,17 @@
     <style>
       body {
         background: #eee;
+        padding: 50px;
       }
       .examples {
         margin-top: 20px;
       }
       media-controller {
-        max-width: 500px;
+        max-width: 300px;
       }
     </style>
   </head>
-  <body>
+  <body>     
     <main>
       <h1>Tooltips</h1>
       
@@ -40,15 +41,23 @@
         ></video>
         <media-control-bar>
           <media-play-button></media-play-button>
-          <media-airplay-button></media-airplay-button>
         </media-control-bar>
       </media-controller>
 
       <h2>Overriding tooltip label</h2>
-      <media-airplay-button></media-airplay-button>
-      <media-play-button>
-        <span slot="tooltip">Play the <strong>damn</strong> movie</span>
-      </media-play-button>
+      <media-controller>
+        <video
+          slot="media"
+          src="https://stream.mux.com/A3VXy02VoUinw01pwyomEO3bHnG4P32xzV7u1j1FSzjNg/high.mp4"
+          muted
+        ></video>
+        <media-control-bar>
+          <media-play-button>
+            <span slot="tooltip-play"><strong>Play</strong> the movie! <img src="/examples/vanilla/images/spacebar.svg" width="12"></span>
+            <span slot="tooltip-pause">Please <i>stop</i>... <img src="/examples/vanilla/images/spacebar.svg" width="12"></strong></span>
+          </media-play-button>
+        </media-control-bar>
+      </media-controller>
 
       <h2>Hiding with CSS</h2>
 
--- a/examples/vanilla/tooltips.html
+++ b/examples/vanilla/tooltips.html
@@ -30,6 +30,13 @@
       <h2>As part of button controls</h2>
 
       <media-play-button></media-play-button>
+      <media-airplay-button></media-airplay-button>
+
+      <h2>Overriding tooltip label</h2>
+      <media-airplay-button></media-airplay-button>
+      <media-play-button>
+        <span slot="tooltip">Play the <strong>damn</strong> movie</span>
+      </media-play-button>
 
 	    <div class="examples">
         <a href="./">View more examples</a>
--- a/examples/vanilla/tooltips.html
+++ b/examples/vanilla/tooltips.html
@@ -100,41 +100,54 @@
 
       <h2>Entirely custom tooltip element</h2>
 
-      <media-play-button style="margin-top: 1.5em">
-        <style>
-          .my-tooltip {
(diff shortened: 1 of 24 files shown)


src/js/media-tooltip.ts
```diff
@@ -0,0 +1,112 @@
...
+  /**
+   * Get or set offset X value
+   */
+  get offsetX(): number | undefined {
+    return getNumericAttr(this, Attributes.OFFSET_X);
+  }
+
+  set offsetX(value: number | undefined) {
+    setNumericAttr(this, Attributes.OFFSET_X, value);
+  }
```
luwes, line 105: wondering if this could be a CSS var instead?

AdamJaggard: Good shout. I'll wait until I've finalised the whole positioning mechanism and then switch it if nothing else relies on the number being passed in

src/js/media-play-button.ts
```diff
@@ -28,13 +32,43 @@ slotTemplate.innerHTML = /*html*/ `
     <slot name="play">${playIcon}</slot>
     <slot name="pause">${pauseIcon}</slot>
   </slot>
+  <media-tooltip id="tooltip">
```
heff, line 35: I'm wondering if we want something like the [slot template](https://github.com/muxinc/media-chrome/blob/67eab2311a000bbd2a7dbf8be97b43cc392f0d29/src/js/media-play-button.ts#L15) we use for icons here.

Then every button subclass wouldn't have to bring its own tool tip, it'd just have to define the contents of the tooltip. And also the whole tool tip could be replaced with a custom one.

src/js/media-play-button.ts
```diff
@@ -28,13 +32,43 @@ slotTemplate.innerHTML = /*html*/ `
...
+  // not spilling out left
+  if (xDiff > 0) {
+    tooltipEl.offsetX = null;
+    return;
+  }
+
+  tooltipEl.offsetX = xDiff;
+
+  // right edge
+  // TODO: ...
```
heff, line 69: Also make sure it doesn't explode when both left and right are true.

src/js/media-play-button.ts
```diff
@@ -53,13 +87,32 @@ class MediaPlayButton extends MediaChromeButton {
...
     super({ slotTemplate, ...options });
+    this.tooltip = this.shadowRoot.querySelector('#tooltip');
   }
 
   connectedCallback(): void {
     updateAriaLabel(this);
     super.connectedCallback();
+    this.addEventListener(
+      'mouseenter',
+      updateTooltipPosition.bind(null, this.tooltip, 'media-control-bar')
```
heff, line 102: Buttons won't always be contained by a control bar, if that's required here.

src/js/media-play-button.ts
```diff
@@ -53,13 +87,32 @@ class MediaPlayButton extends MediaChromeButton {
...
+
   constructor(options = {}) {
     super({ slotTemplate, ...options });
+    this.tooltip = this.shadowRoot.querySelector('#tooltip');
   }
 
   connectedCallback(): void {
     updateAriaLabel(this);
     super.connectedCallback();
+    this.addEventListener(
```
heff, line 100: This feels like generic tooltip logic that should go in media-chrome-button, in case that's not already the plan.

heff (commented):
Great start!

src/js/media-chrome-button.ts
```diff
@@ -61,6 +62,19 @@ template.innerHTML = /*html*/ `
     max-height: 100%;
     min-width: 100%;
   }
+
+  media-tooltip {
+    position: absolute;
+    bottom: calc(100% + 12px);
```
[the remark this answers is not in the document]
AdamJaggard: This is something I've been thinking a bit about. There's some complexity in being able to both dynamically position the tooltip as well as stop it going out of "bounds" (if defined). Maybe not too complex though and just lots of cases to account for.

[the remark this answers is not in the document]
AdamJaggard: Definitely! I left a TODO comment in `media-tooltip.ts` about replacing some hard coded values with css vars, it should apply to this style too

src/js/media-play-button.ts
AdamJaggard: Won't be a hard requirement, this is just for testing out the logic of it being contained by "something", as we know that's needed for Mux Player. How and when this is invoked isn't defined yet

src/js/media-play-button.ts
AdamJaggard: That is the plan! just working within this control while the logic for everything is worked out. Still some refactoring to do

src/js/media-tooltip.ts
```diff
@@ -0,0 +1,239 @@
...
+    if (leftDiff < 0) {
+      this.style.setProperty('--media-tooltip-offset-x', `${leftDiff}px`);
+      return;
+    }
+
+    // out of right bounds
+    if (rightDiff > 0) {
+      this.style.setProperty('--media-tooltip-offset-x', `${rightDiff}px`);
+      return;
+    }
```
heff, line 205: These can both be true at the same time, but it seems to be working ok. At a certain point the left edge becomes dominant, which seems fine. Or at least I can't think of how I'd want it to work differently.

image
image

src/js/media-tooltip.ts
```diff
@@ -0,0 +1,239 @@
...
+      : 0;
+
+    // we might have already offset the tooltip previously so we remove it's
+    // current offset from our calculations
+    const leftDiff = tooltipX - containerX + currOffsetX;
+    const rightDiff = tooltipRight - containerRight + currOffsetX;
+
+    // out of left bounds
+    if (leftDiff < 0) {
+      this.style.setProperty('--media-tooltip-offset-x', `${leftDiff}px`);
```
heff, line 197: How this is all working is hard to understand from scratch so don't be shy with code comments. Maybe a summary of how `media-tooltip-offset-x` is ultimately used to set both sides.

src/js/media-tooltip.ts
```diff
@@ -0,0 +1,239 @@
...
+    // we don't offset against tooltips coming out of left and right sides
+    if (position === 'left' || position === 'right') {
+      // could have been offset before switching to a new position
+      this.style.removeProperty('--media-tooltip-offset-x');
+      return;
+    }
+
+    const containingSelector = this.container;
+    const containingEl = closestComposedNode(this, containingSelector);
+    if (!containingEl) return;
```
heff, line 179: I think it'd be good to default this to `media-controller` specifically. The main thing you don't want is the tooltip going outside of the player, especially in full screen mode, and the media-controller best represents that most of the time. The control-bar is a little too optional to rely on I think, and the control-bar left/right margin can need to be zero for some designs, as opposed to always relying on the control bar providing the margin between the tooltip and the edge of the screen.

But so with that, I think we'd want some kind of default margin between the tooltip and the edge of the media controller or supplied container. Like `--media-tooltip-container-margin`. How does that sound?

src/js/media-tooltip.ts
```diff
@@ -0,0 +1,239 @@
...
+
+  set position(value: TooltipPosition | undefined) {
+    setStringAttr(this, Attributes.POSITION, value);
+  }
+
+  /**
+   * Get or set tooltip container selector that will constrain the tooltips horizontal
+   * position. A css selector that matches one of the tooltips parents.
+   */
+  get container(): string | undefined {
```
heff, line 226: `container` feels a little vague here, or at least I'd expect an element from it not string. It's hard to think of something better. Maybe `containWithin`. I like that because it's a little clearer what the functionality is and I don't expect any specific values from it, so query selector values seem fine to use.

heff (commented):
It's coming along and looking really good. Happy to jump on a call and talk through any of the comments here.

src/js/media-tooltip.ts
AdamJaggard: This is a CSS var now!

examples/vanilla/tooltips.html
```diff
@@ -0,0 +1,159 @@
...
+          <media-play-button>
+            <span slot="tooltip-play">Play......</span>
+            <span slot="tooltip-pause">Pause........</span>
+          </media-play-button>
+        </media-control-bar>
+      </media-controller>
+
+      <h2>Hiding with CSS</h2>
+
+      <media-play-button style="--media-tooltip-display: none;"></media-play-button>
```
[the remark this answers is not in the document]
AdamJaggard: This example completely hides the tooltip where as `tooltipposition="none"` leaves the tooltip visible but doesn't position it in any way (position `relative` essentially). Where you imagining that ``tooltipposition="none"` would also `display: none` the tooltip? something I hadn't considered!

examples/vanilla/tooltips.html
```diff
@@ -0,0 +1,159 @@
...
+            font-family: "Comic Sans MS", "Comic Sans", cursive;
+            font-size: 18px;
+          }
+          .my-tooltip marquee { display: block; }
+          media-play-button:hover .my-tooltip {
+            opacity: 1;
+            scale: 1;
+          }
+        </style>
+        <div slot="tooltip" class="my-tooltip">
```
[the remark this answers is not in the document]
AdamJaggard: Good shout

src/js/media-tooltip.ts
AdamJaggard: That makes sense. Currently `media-xyz-button`'s are defining the containing element, but we could switch it to a default (media-controller) that can be optionally overridden and manually adjusted with a margin.

src/js/media-tooltip.ts
AdamJaggard: Yeah I gave a little though to this but couldn't come up with a clean "this is how it _should_ look" in these types of scenarios and left it currently as undefined behaviour or a design issue instead of an implementation issue. We could try and detect this and fall back to centering? but it would still be visually strange.

src/js/media-tooltip.ts
AdamJaggard: I had some more verbose names for this initially so was hoping someone would have an idea for something descriptive and succinct. I think `containWithin` is pretty good. @cjpillsbury also mentioned that to follow existing patterns we should accept an element as well as a string potentially.

src/js/media-tooltip.ts
```diff
@@ -0,0 +1,239 @@
...
+      bottom: calc(100% + var(--media-tooltip-distance, 12px));
+      left: 50%;
+      transform: translate(calc(-50% - var(--media-tooltip-offset-x, 0px)), 0);
+    }
+    :host(:not([position])) #arrow,
+    :host([position="top"]) #arrow {
+      top: 100%;
+      left: 50%;
+      border-width: 5px 6px 0 6px;
+      border-color: var(--_tooltip-background) transparent transparent transparent;
```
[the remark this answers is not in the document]
AdamJaggard: Have pushed an update for this. I also added n new `--media-tooltip-border` var even though we don't use it ourselves, but means you can do something similar to this if someone wanted.
image

examples/vanilla/tooltips.html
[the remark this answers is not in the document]
AdamJaggard: Have pushed up an update that includes this

src/js/media-chrome-button.ts
```diff
@@ -189,6 +208,12 @@ class MediaChromeButton extends globalThis.HTMLElement {
       } else {
         this.disable();
       }
+    } else if (attrName === 'tooltipposition' && newValue !== oldValue) {
+      // If the tooltip isn't defined, then we could accidentally overwrite
+      // the '.position' property
+      if (this.tooltip && globalThis.customElements.get('media-tooltip')) {
+        this.tooltip.position = newValue;
+      }
```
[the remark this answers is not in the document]
AdamJaggard: The issue I was seeing that prompted this workaround was that position was not applying ever, if `.position` was attempting to be set before the element was defined. I need to revisit this as the code has changed a bit since but it would be nice not to have to do the check.

I think your last point is covered by the `setupTooltip` method, which runs after `whenDefined` resolves (setup in `connectedCallback`), regardless of if attribute changes have applied yet or not. That method grabs the latest tooltipposition value and sets it on the tooltip so any attribute change misses will be caught at this point. So I _think_ this is fine... but I will do some tests.

src/js/media-tooltip.ts
AdamJaggard: Have pushed up some changes for this. Defaulting to `media-controller` and with an optional margin css var to pad it out.

src/js/media-tooltip.ts
AdamJaggard: Have added some more comments to this that that hopefully explains this better. The only difference between needing to be brought in from the left or from the right is if you need a positive offset (push to the right) or negative offset (pull it further left) so it just re-uses the variable for both cases, setting a positive or negative offset depending on what needs to be done.

src/js/media-tooltip.ts
AdamJaggard: Have updated this to `containWithin`

src/js/media-chrome-button.ts
```diff
@@ -214,13 +239,27 @@ class MediaChromeButton extends globalThis.HTMLElement {
...
 
   disconnectedCallback() {
     this.disable();
     // Use cached mediaController, getRootNode() doesn't work if disconnected.
     this.#mediaController?.unassociateElement?.(this);
     this.#mediaController = null;
+
+    this.removeEventListener('mouseenter', this.tooltip?.updateXOffset);
+    this.removeEventListener('focus', this.tooltip?.updateXOffset);
+    // TODO: how to remove this correctly?
```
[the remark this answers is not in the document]
AdamJaggard: All sorted. We already had a click handler method in the class so I just piggybacked on that

examples/vanilla/tooltips.html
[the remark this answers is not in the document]
heff: Yeah, that's what I meant by the "none" option. The css var works but it's a tad bit advanced for how important that feature is, and the attribute option will surface the feature in docs better. Otherwise I can't see much value in just an unpositioned tooltip. 

Open to other naming/setting directions for that, but it felt nice to not introduce another attribute.

src/js/media-chrome-button.ts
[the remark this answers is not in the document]
heff: Ahh right, that makes sense now. The tooltip instance would `hasOwnProperty` `position` so the prototype version would get ignored. (Confirmed in [this writeup](https://nolanlawson.com/2021/08/03/handling-properties-in-custom-element-upgrades/))

This is gonna happen for other custom elements that use tooltips too, so I'm tempted to fix this in Tooltip. You could use the writeup fix and just do the `delete this.position` part since setupTooltip will handle it from there.

src/js/media-tooltip.ts
heff: Yeah, works for now. 👍  If there's real issues with this we could look into a max-width and line wrapping.

src/js/media-tooltip.ts
heff: > we should accept an element as well as a string potentially

I think that's right in general but this is a situation kind of like the `mediacontroller` attribute where I think it'd be good if we didn't hold on to a reference of the element. Unless we need to do that anyway.

cjpillsbury: if we're sticking with generic class selectors (instead of ids) this should work for any of our "automagic container picking" solutions

examples/vanilla/tooltips.html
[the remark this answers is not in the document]
AdamJaggard: I think the value in an un-positioned tooltip is if you want to be in control of positioning yourself. If a 'none' position hides the tooltip, there's no way to opt out of the positioning that is baked in (position absolute, transforms, etc) unless we add a lot more css vars. There's also something slightly odd to me about a `position` property controlling visibility, but I don't feel particularly strongly about any of it so happy to go in either direction. Maybe @luwes or @cjpillsbury can offer a tie breaker? 🙂

src/js/media-chrome-button.ts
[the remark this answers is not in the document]
AdamJaggard: Very interesting! I implemented the `hasOwnProperty` fix in full and it looks like it works well. I kept the idea of caching the value and setting it after deletion just in case we change our media button implementation one day, and I want to avoid the the tooltip having to have knowledge about how it is being implemented exactly.

src/js/media-tooltip.ts
```diff
@@ -0,0 +1,278 @@
...
+  /**
+   * Get or set tooltip position
+   */
+  get position(): TooltipPosition | undefined {
+    return getStringAttr(this, Attributes.POSITION);
+  }
+
+  set position(value: TooltipPosition | undefined) {
+    setStringAttr(this, Attributes.POSITION, value);
+  }
```
luwes, line 259: what do you think of renaming `position` to `placement` or `place`?

- the reasoning is in CSS the `position` property is something else, there are other props with `place-*` that are more similar to what we're doing here.
- the most popular floating library (floating UI) uses `placement` as well, as does Shoelace and Spectrum web comps

src/js/media-tooltip.ts
```diff
@@ -0,0 +1,239 @@
...
+    :host([position="right"]) {
+      position: absolute;
+      left: calc(100% + var(--media-tooltip-distance, 12px));
+      top: 50%;
+      transform: translate(0, -50%);
+    }
+    :host([position="right"]) #arrow {
+      top: 50%;
+      right: 100%;
+      border-width: 6px 5px 6px 0;
```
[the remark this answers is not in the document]
AdamJaggard: Have added some more css vars for this `--media-tooltip-arrow-height`, `--media-tooltip-arrow-width`

examples/vanilla/tooltips.html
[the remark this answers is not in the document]
luwes: I would also not overload the `tooltipposition` attr with a hiding functionality.

Weren't we going to add an easy attribute to change the text of the tooltips within the button. 
Could it be an empty `tooltip=""` attribute?

src/js/media-tooltip.ts
cjpillsbury: > what do you think of renaming `position` to `placement` or `place`?

`placement` seems like a good one for me. Add to the list:
- MUI - https://mui.com/material-ui/react-tooltip/#positioned-tooltips
- Chakra - https://v2.chakra-ui.com/docs/components/tooltip/usage#placement
- Ant Design - https://ant.design/components/tooltip#tooltip-demo-placement

Probably others as well.

cjpillsbury: The one thing in the background that I mentioned to Adam that's (hopefully) upcoming, re: "position" is - https://developer.chrome.com/blog/anchor-positioning-api

src/js/media-tooltip.ts
luwes: sorry, this will need additional changes. 
we've used the `bounds` attribute for several elements already (time-range, menu). 
can we align on that? 

see https://github.com/muxinc/media-chrome/blob/main/src/js/experimental/media-chrome-selectmenu.ts#L346-L362

src/js/media-tooltip.ts
```diff
@@ -0,0 +1,278 @@
...
+   * Get or set tooltip container selector that will constrain the tooltips horizontal
+   * position. A css selector that matches one of the tooltips parents.
+   */
+  get containWithin(): string | undefined {
+    return getStringAttr(this, Attributes.CONTAIN_WITHIN);
+  }
+
+  set containWithin(value: string | undefined) {
+    setStringAttr(this, Attributes.CONTAIN_WITHIN, value);
+  }
```
luwes, line 271: see other comment, we should try to align on this naming. we've used `bounds` in other elements

src/js/media-chrome-button.ts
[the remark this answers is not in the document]
luwes: we use this technique in the theme element too
https://github.com/muxinc/media-chrome/blob/ac8bf0b3d4d59f9875337a626cd1175ab5d0dd43/src/js/media-theme-element.ts#L113-L126

src/js/media-play-button.ts
```diff
@@ -28,6 +27,12 @@ slotTemplate.innerHTML = /*html*/ `
     <slot name="play">${playIcon}</slot>
     <slot name="pause">${pauseIcon}</slot>
   </slot>
+  <slot name="tooltip">
+    <media-tooltip container="media-control-bar">
+      <slot name="tooltip-play">Play</slot>
+      <slot name="tooltip-pause">Pause</slot>
```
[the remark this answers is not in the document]
AdamJaggard: I've pushed a change to this up that moves the tooltip definition into the generic button class, with the subclass just passing in what it wants to be used as the tooltip content (the play and pause slots). Take a look if this is what you where thinking.

src/js/media-tooltip.ts
luwes: yep, the anchor API is a good callout. one complex puppy of an API :) but will be powerful stuff

src/js/media-chrome-button.ts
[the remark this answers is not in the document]
AdamJaggard: I don't know if support is good enough to use it but I noticed this also exists as a replacement for `Object.prototype.hasOwnProperty.call(this, prop)`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn

src/js/media-tooltip.ts
AdamJaggard: Have updated this to placement 👍

examples/vanilla/tooltips.html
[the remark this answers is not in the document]
AdamJaggard: I've pushed up a tiny CSS change that would allow people to just use the standard global `hidden` attribute e.g. `<media-tooltip hidden>`, is there a reason not to use that? admitted I haven't used it myself in a long time, but it seems to be using it as intended.

src/js/media-tooltip.ts
AdamJaggard: @luwes, do you think someone might want to set different bounds for the tooltips compared to what the `bounds` attribute is already added to? e.g. someone wants all the menus to contain within one thing, but the tooltips to be contained by something else (maybe something wrapping just the buttons). I'm happy to piggy back on the `bounds` attribute if that doesn't sound like something that's needed. I also prefer the consistency if we did use it if there's no issue with the above.

src/js/media-tooltip.ts
AdamJaggard: Just resolving this convo in favour of the other one above

examples/vanilla/tooltips.html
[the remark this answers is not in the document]
heff: `hidden` works for a stand-alone tooltip 👍 We also need an attribute for the media-play-button etc.

To Wes's point, I do think we need a `tooltip=""` attribute for these buttons as an easy way to set that. I know the pushback has been "two ways of doing the same thing", but @mmcc commented that changing the content and swapping an element are really two different use cases, and I agree with that.

From there, I don't have a strong opinion if it's `tooltip=""` or `notooltip`. Either one. `notooltip` feels a little cleaner since we're going to default to them being on, but `tooltip=""` means one less attr.

src/js/media-chrome-button.ts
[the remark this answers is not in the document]
heff: Looks safe enough to me. 👍

src/js/media-tooltip.ts
luwes: @AdamJaggard they could still do that I think, the `bounds` attribute is set on the timerange element or the menu element or on the tooltip element. 
https://github.com/muxinc/media-chrome/blob/main/src/js/media-chrome-menu.ts#L956-L960

AdamJaggard: Hey @luwes, have updated this to `bounds`, let me know if I've missed anything matching current implementation. I didn't re-use the `getBoundsElement` function as it falls back to the immediate parent (or itself) as the bounds element if it can't find anything and this breaks tooltip assumptions.

examples/vanilla/tooltips.html
```diff
@@ -0,0 +1,246 @@
...
+          <media-fullscreen-button></media-fullscreen-button>
+          <media-mute-button></media-mute-button>
+          <media-pip-button></media-pip-button>
+          <media-play-button></media-play-button>
+          <media-playback-rate-button></media-playback-rate-button>
+          <media-playback-rate-menu-button></media-playback-rate-menu-button>
+          <media-rendition-menu-button></media-rendition-menu-button>
+          <media-seek-backward-button></media-seek-backward-button>
+          <media-seek-forward-button></media-seek-forward-button>
+          <media-settings-menu-button></media-settings-menu-button>
```
luwes, line 80: could the menu be added in the example so the interaction between tooltip and menu showing can be tested?

src/js/media-airplay-button.ts
```diff
@@ -66,7 +79,7 @@ class MediaAirplayButton extends MediaChromeButton {
   }
 
   constructor(options: { slotTemplate?: HTMLTemplateElement } = {}) {
-    super({ slotTemplate, ...options });
+    super({ slotTemplate, tooltipContent, ...options });
```
luwes, line 82: (non-blocking) I think we should move away from passing args to the constructor in general.
reason being in the DOM when creating a CE there is no such a thing e.g. `document.createElement('x-button')`. attrs or props are always added after construction. cc @cjpillsbury @heff

src/js/media-captions-button.ts
```diff
@@ -36,6 +41,11 @@ slotTemplate.innerHTML = /*html*/ `
   </slot>
 `;
 
+const tooltipContent = /*html*/ `
+  <slot name="tooltip-enable">Enable captions</slot>
+  <slot name="tooltip-disable">Disable captions</slot>
```
luwes, line 46: (non-b) fine to iterate and add these later to a constants file for i18n etc

src/js/media-cast-button.ts
```diff
@@ -18,16 +19,23 @@ slotTemplate.innerHTML = /*html*/ `
...
-  }])) slot:not([name=enter]):not([name=icon]) {
+  }])) slot[name=icon] slot:not([name=enter]) {
     display: none !important;
   }
+
+  :host([${MediaUIAttributes.MEDIA_IS_CASTING}]) slot[name=tooltip-enter],
+    :host(:not([${
+      MediaUIAttributes.MEDIA_IS_CASTING
+    }])) slot[name=tooltip-exit] {
+      display: none;
```
luwes, line 37: this prettier formatting 😅

src/js/media-chrome-button.ts
```diff
@@ -1,12 +1,21 @@
...
+import {
+  getOrInsertCSSRule,
+  getStringAttr,
+  setStringAttr,
+} from './utils/element-utils.js';
 import { globalThis, document } from './utils/server-safe-globals.js';
 
-const template = document.createElement('template');
+const Attributes = {
+  TOOLTIP_PLACEMENT: 'tooltipplacement',
```
luwes, line 11: 👍

src/js/media-chrome-button.ts
```diff
@@ -61,14 +70,36 @@ template.innerHTML = /*html*/ `
...
+  :host([notooltip]) slot[name="tooltip"] {
+    display: none;
+  }
 </style>
+
+<slot name="tooltip">
+  <media-tooltip>
+    <slot name="tooltip-content"></slot>
+  </media-tooltip>
+</slot>
```
luwes, line 93: 👍

src/js/media-chrome-button.ts
```diff
@@ -61,14 +70,36 @@ template.innerHTML = /*html*/ `
     max-height: 100%;
     min-width: 100%;
   }
+
+  media-tooltip {
+    opacity: 0;
+    transition: opacity .3s;
```
luwes, line 76: can be an iteration but I'd probably like to change the:
- transition delay
- transition duration
- maybe the transform of the tooltip 
- maybe the easing of the transition

src/js/media-captions-button.ts
AdamJaggard: Was chatting to Christian about this and have added these as constants to the labels util file now as a simple dictionary, as re-using the verbs dictionary wasn't quite correct and was missing entries

PR #946, opened 2024-07-25
fix: refactor fullscreen-api and state mediator to make fewer assumpt…

cjpillsbury:
…ions on env-specific divergences. Fixes bug in modern Safari iOS fullscreen.

### DIFF
--- a/src/js/utils/fullscreen-api.js
+++ b/src/js/utils/fullscreen-api.js
@@ -1,27 +1,36 @@
 import { document } from './server-safe-globals.js';
 
 export const fullscreenApi = {
-  enter: 'requestFullscreen',
-  exit: 'exitFullscreen',
-  rootEvents: ['fullscreenchange'],
-  mediaEvents: [],
-  element: 'fullscreenElement',
-  error: 'fullscreenerror',
-  enabled: 'fullscreenEnabled',
-};
-
-if (document.fullscreenElement === undefined) {
-  fullscreenApi.enter = 'webkitRequestFullScreen';
-  fullscreenApi.exit =
-    document.webkitExitFullscreen != null
+  enter:
+    'requestFullscreen' in document
+      ? 'requestFullscreen'
+      : 'webkitRequestFullScreen' in document
+      ? 'webkitRequestFullScreen'
+      : undefined,
+  exit:
+    'exitFullscreen' in document
+      ? 'exitFullscreen'
+      : 'webkitExitFullscreen' in document
       ? 'webkitExitFullscreen'
-      : 'webkitCancelFullScreen';
-  fullscreenApi.rootEvents = ['webkitfullscreenchange'];
-  (fullscreenApi.mediaEvents = [
-    'webkitbeginfullscreen',
-    'webkitendfullscreen',
-  ]),
-    (fullscreenApi.element = 'webkitFullscreenElement');
-  fullscreenApi.error = 'webkitfullscreenerror';
-  fullscreenApi.enabled = 'webkitFullscreenEnabled';
-}
+      : 'webkitCancelFullScreen' in document
+      ? 'webkitCancelFullScreen'
+      : undefined,
+  element:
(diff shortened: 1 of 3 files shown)


src/js/utils/fullscreen-api.js
```diff
@@ -1,27 +1,36 @@
...
-};
-
-if (document.fullscreenElement === undefined) {
-  fullscreenApi.enter = 'webkitRequestFullScreen';
-  fullscreenApi.exit =
-    document.webkitExitFullscreen != null
+  enter:
+    'requestFullscreen' in document
+      ? 'requestFullscreen'
+      : 'webkitRequestFullScreen' in document
```
luwes, line 7: are these always available in the document if the target is actually the media element?

cjpillsbury: Thanks for the catch (discussing out of band)

src/js/utils/fullscreen-api.js
```diff
@@ -1,27 +1,179 @@
...
+  } else if (media?.webkitEnterFullscreen) {
+    // Media element fullscreen using iOS API
+    media.webkitEnterFullscreen();
+  } else if (media?.requestFullscreen) {
+    // So media els don't have to implement multiple APIs.
+    media.requestFullscreen();
+  }
+};
+
+const exitFullscreenKey =
```
cjpillsbury, line 62: Per out of band convo, should these "prefer" browser prefix versions or standard versions?

luwes: like this seems good to me, prefer prefix-less

luwes (approved):
LGTM!

src/js/utils/fullscreen-api.js
cjpillsbury: cool would love to get @heff's and/or @mmcc's gut on this as well

heff: Yeah, can't say I'm thinking too deeply on this one, but I also prefer prefixless, in the hopes we can just delete code later without changing much more.

src/js/utils/fullscreen-api.js
```diff
@@ -1,27 +1,179 @@
...
+
+/**
+ * @typedef {object} StateOwners
+ * @property {MediaStateOwner} [media]
+ * @property {RootNodeStateOwner} [documentElement]
+ * @property {FullScreenElementStateOwner} [fullscreenElement]
+ */
+
+/** @type {(stateOwners: StateOwners) => Promise<undefined> | undefined} */
+export const enterFullscreen = (stateOwners) => {
```
heff, line 32: This is definitely a nit pick, so take it or leave it, but if we're keeping these as `Utils` it'd feel better if they weren't tied to the concept of stateOwners and media + fullscreenEl were specific arguments.

heff (approved):
Thanks for fixing this!

src/js/utils/fullscreen-api.js
cjpillsbury: Yeah for me I just wasn't sure if there was an expectation that, if both are around, which would be more reliable. That said, I think the status quo wins by vote, so I'm gonna mash that merge.

src/js/utils/fullscreen-api.js
cjpillsbury: Yeah I def get that intuition. Here was my reasoning:

- If you squint, this is just a common convention for passing in multiple optional arguments in JS (an object with optional key + value pairs).
- The "shape" of that argument object **_happens_** to use the same names as `StateOwners`, in large part out of convenience (since, aside from one case in platform checks, it's only used by the Media Store architecture)
- Even there, it isn't **_identical_** to `StateOwners`, because it only expects/requires a small subset of what has to be defined on them

I'm gonna leave for now but we can certainly revisit in next week's MCOH (feels like it would fit in the "best practices" category if we wanna define something either way here)

Issue #933, opened 2024-07-02
V4 Tracking


heff:
- [x] Tool tips (https://github.com/muxinc/media-chrome/pull/941)
- [x] Typescript conversion (https://github.com/muxinc/media-chrome/pull/931)
- [x] Redo component imports to be conservative on default bundle size 
	(https://github.com/muxinc/media-chrome/pull/950)
	(https://github.com/muxinc/media-chrome/pull/951)
- [x] include updates from v3 branch
	https://github.com/muxinc/media-chrome/compare/main...3.x
- [x] fix docs menu comp reference section 
	https://www.media-chrome.org/docs/en/components/media-audio-track-menu#reference

mmcc:
Consider making the baobun design our default theme out of the box for Media Chrome.

PR #951, opened 2024-08-08
feat!: remove selectmenu & co for v4

luwes:
this removes the older experimental selectmenu and related components

related https://github.com/muxinc/media-chrome/issues/933


examples/vanilla/disabled.html
```diff
@@ -73,7 +73,6 @@ <h1>Media Chrome Disabled-Controls Video Usage Example</h1>
           <media-time-range disabled aria-disabled="true"></media-time-range>
           <media-time-display showduration remaining disabled aria-disabled="true"></media-time-display>
           <media-captions-button default-showing disabled aria-disabled="true"></media-captions-button>
-          <media-captions-selectmenu disabled aria-disabled="true"></media-captions-selectmenu>
```
cjpillsbury, line 76: issue(non-blocking): Realizing looking here that we don't demo disabled behavior for new menu/menu button architecture. Does that work as expected? If yes, we should probably add example(s) here.

cjpillsbury (approved):
One nb callout but otherwise LGTM

Issue #923, opened 2024-06-11
Add seekToLiveOffset attribute


k-g-a:
### The problem

Currentrly, if we have a paused live stream and issue `MEDIA_PLAY_REQUEST` (i.e. via pressing `play` button), we are thrown straight to the end of seekable range:

```ts
const liveEdgeTime = stateMediator.mediaSeekable.get(stateOwners)?.[1];

if (notDvr && liveEdgeTime) {
  stateMediator.mediaCurrentTime.set(liveEdgeTime, stateOwners);
}
```

Unfortunatelly, this often causes a live stream to stall afterwards (not necessary immediately) due to the fact that it's playing too close to the edge - new video fragments just don't catch up in time, especially if poor network conditions are in place.

### Proposed solution

The first thing that comes to mind is to use `liveEdgeOffset` within `MEDIA_PLAY_REQUEST` handler:

```ts
const liveEdgeTime = stateMediator.mediaSeekable.get(stateOwners)?.[1];

if (notDvr && liveEdgeTime) {
  const liveEdgeOffset = stateOwners.options.liveEdgeOffset ?? 0;
  stateMediator.mediaCurrentTime.set(liveEdgeTime - liveEdgeOffset, stateOwners);
}
```

But this may lead up to the other issue: if we are close enough to the moment of the next video segment arrival, as soon as it arrives we will fall back behind the `liveEdgeOffset` (i.e. `duration`/`seekableEnd` jump forward by the amount of the segment length). This will cause `media-live-button` to show that we are "not live" (i.e. become gray), but in a few moments the playback will catch up with  `liveEdgeOffset` so the `media-live-button` becomes "live" (i.e. red). This process will continue to repeat and the  `media-live-button` will "flicker". This is especially the case for segments with small durations (i.e. 3-5 seconds).

So, it's more reasonable to introduce another attribute called `SEEK_TO_LIVE_OFFSET: 'seektoliveoffset'` (with default of `+this.getAttribute(Attributes.LIVE_EDGE_OFFSET)` for example) and use it within `MEDIA_PLAY_REQUEST` nahdler. 

In this case we could provide, for example of a 3-seconds video fragment length, the following values `<media-controller liveedgeoffset="15" seektoliveoffset="11">` and get both: reliable `media-live-button` indication and enough segments ahead to overcome network problems.

_In case this proposal is accepted, I can make a PR._

### P.S.  

We used to "hack" this behaviour in previous versions of `media-chrome` in the following way:

```ts

[...]

PR #958, opened 2024-08-16
feat: Add seektoliveoffset attribute

k-g-a:
1. Implements https://github.com/muxinc/media-chrome/issues/923, adds documentation (with motivation).
2. Fixes a regression of `noautoseektolive` not being used at all
3. Fixes a couple of broken links in docs.


luwes:
thank you for the contribution! I'll bring this up with the team.

k-g-a:
Hi, @luwes! 

I'd like to fix the coverage issue of this PR, but, unfortunatelly, I could not find any examples of working with live streams in test cases. So I'm open to ideas/suggestions to implement.

luwes:
@k-g-a thanks so much for this contribution! And appologies for the late action on our part.

This concept makes sense to me, my colleague pointed out that if everything would go perfectly as in the HLS spec this seek offset wouldn't be needed but real world is often different and Media Chrome could have different media elements like DASH, etc.

PR #964, opened 2024-09-03
docs: change doc urls

arturparkhisenko:
Hi Mux team, I've found some broken URLs after TypeScript migration and tried to address some of them (renaming from js to ts).
Please review, I've updated some lines of code links to match the new code.
Also, I've found that some MDX files have that problem but it's a story for another time.


AdamJaggard (approved):
Thank you for finding all of these @arturparkhisenko! will get these merged and deployed shortly

PR #966, opened 2024-09-10
fix: position menu absolute in control bar

luwes:
to prevent taking space


cjpillsbury (approved):
pre-approving/deferring to you on pref, but we should probably update docs for this. Mind either adding a followup issue or updating:
- Add a section to control bar docs a la https://www.media-chrome.org/docs/en/components/media-control-bar#use-with-media-time-range
- Update docs for each menu component, either linking to control bar section :point_up:, updating examples to be more like https://www.media-chrome.org/docs/en/components/media-time-range, or something similar

cjpillsbury:
We should also update our migration guide to explain how to go from old select menu architecture to your new menu architecture (not sure if an issue was ever created).

luwes:
@AdamJaggard updated the migration guide here https://github.com/muxinc/media-chrome/pull/965

PR #967, opened 2024-09-10
fix: NaN valueAsNumber Safari error

### LINKED ISSUE #947: range.valueAsNumber operation is not supported

luwes:
fix #947


AdamJaggard (approved):
Good catch

PR #982, opened 2024-09-24
fix: Safari (webkit) style tag styles not applied

luwes:
regression in the previous fix. 

Safari!!!! 

SCR-20240924-mgkd


cjpillsbury (approved):
🚢 🇮🇹

PR #988, opened 2024-09-26
fix: tooltip type error

### LINKED ISSUE #986: TypeError: Cannot read properties of null (reading 'updateXOffset')
I'm getting this error on Sentry, running an App on Next.js, everything is up-to-date.

Triggered on this line:
https://github.com/muxinc/media-chrome/blob/78c303221439b097eb4c2ffffc3fefd25b5e9722/src/js/media-chrome-button.ts#L324

Sampling:
55% Chrome Mobile WebView 117.0.5938 
14% Chrome Mobile 128.0.0
10% Mobile Safari 17.6
3% Chrome Mobile 129.0.0

I don't have replay sessions, but it seems this code chunk is called when the component is being unloaded.

luwes:
fix #986

### DIFF
--- a/src/js/media-chrome-button.ts
+++ b/src/js/media-chrome-button.ts
@@ -171,9 +171,9 @@ class MediaChromeButton extends globalThis.HTMLElement {
       this.nativeEl.appendChild(slotTemplate.content.cloneNode(true));
 
       this.shadowRoot.appendChild(buttonHTML);
-
-      this.tooltipEl = this.shadowRoot.querySelector('media-tooltip');
     }
+
+    this.tooltipEl = this.shadowRoot.querySelector('media-tooltip');
   }
 
   #clickListener = (e) => {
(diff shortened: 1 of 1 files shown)


src/js/media-chrome-button.ts
```diff
@@ -171,9 +171,9 @@ class MediaChromeButton extends globalThis.HTMLElement {
       this.nativeEl.appendChild(slotTemplate.content.cloneNode(true));
 
       this.shadowRoot.appendChild(buttonHTML);
-
-      this.tooltipEl = this.shadowRoot.querySelector('media-tooltip');
     }
+
+    this.tooltipEl = this.shadowRoot.querySelector('media-tooltip');
```
luwes, line 176: we still want to get the media-tooltip if the shadow dom was SSR'd

src/js/media-chrome-button.ts
```diff
@@ -292,10 +294,9 @@ class MediaChromeButton extends globalThis.HTMLElement {
     this.#mediaController?.unassociateElement?.(this);
     this.#mediaController = null;
 
-    this.removeEventListener('mouseenter', this.tooltipEl?.updateXOffset);
-    this.removeEventListener('focus', this.tooltipEl?.updateXOffset);
+    this.removeEventListener('mouseenter', this.#positionTooltip);
+    this.removeEventListener('focus', this.#positionTooltip);
     this.removeEventListener('click', this.#clickListener);
-    this.tooltipEl = null;
```
luwes, line 298: on disconnected we don't want to empty this property because we only query it in the constructor, not in connectedCallback

src/js/media-chrome-button.ts
cjpillsbury: question(non-blocking): given the error, should we just move `tooltipEl` to a getter that lazy evals `this.shadowRoot.querySelector('media-tooltip');`?

cjpillsbury (approved):
One nb question/suggestion. Otherwise lgtm.

luwes:
good idea. for now it's fine I think because we don't reset it anymore. we cache it right after the DOM is appended so the query will always return the media-tooltip element.

PR #990, opened 2024-09-26
feat: add range CSS parts

### LINKED ISSUE #921: Customizing the buffering bar
It appears that there's no way to impact the buffering bar (`#highlight`). I'd like to add a border radius to this to match the rest of the items impacted by `--media-range-track-border-radius`

### LINKED ISSUE #983: media-time-range height transition on hover
add support for transitioning the height of the track.

we want this in multiple themes but doesn't work currently.

luwes:
fix #921, fix #983

- Brushed up mobile example https://media-chrome-git-fork-luwes-range-css-parts-mux.vercel.app/examples/vanilla/mobile.html
- Removes the background element, we can use the track element for setting the CSS background.
- Adds CSS parts: `thumb`, `track`, `progress`, `buffered`.
- `--media-range-track-color` CSS var is removed from JS docs, this was left over from the old structure with CSS linear-gradient where it represented the color of the range section on the right of the thumb.

1. a smooth track height transition on hover looks like

```css
  media-time-range::part(track) {
    transition: transform 0.2s;
    transform: scaleY(1);
  }

  media-time-range:hover::part(track) {
    transform: scaleY(1.5);
  }
```

2. adding border-radius to track buffered and track progress elements

```css
media-time-range::part(buffered),
media-time-range::part(progress) {
  border-radius: 9999px;
}
```

### DIFF
--- a/src/js/media-chrome-range.ts
+++ b/src/js/media-chrome-range.ts
@@ -130,32 +130,22 @@ template.innerHTML = /*html*/ `
       will-change: transform;
     }
 
-    #background,
     #track {
-      border-radius: var(--media-range-track-border-radius, 1px);
-      position: absolute;
-      width: 100%;
-      height: 100%;
-    }
-
-    #background {
       background: var(--media-range-track-background, rgb(255 255 255 / .2));
-      backdrop-filter: var(--media-range-track-background-backdrop-filter);
-      -webkit-backdrop-filter: var(--media-range-track-background-backdrop-filter);
-    }
-
-    #track {
+      border-radius: var(--media-range-track-border-radius, 1px);
       border: var(--media-range-track-border, none);
       outline: var(--media-range-track-outline);
       outline-offset: var(--media-range-track-outline-offset);
       backdrop-filter: var(--media-range-track-backdrop-filter);
       -webkit-backdrop-filter: var(--media-range-track-backdrop-filter);
       box-shadow: var(--media-range-track-box-shadow, none);
+      position: absolute;
+      width: 100%;
+      height: 100%;
       overflow: hidden;
     }
 
     #progress,
-    #highlight,
     #pointer {
       position: absolute;
       height: 100%;
@@ -207,7 +197,6 @@ template.innerHTML = /*html*/ `
       height: var(--media-range-segment-hover-height, 7px);
     }
 
-    #background,
     #track {
       clip-path: url(#segments-clipping);
     }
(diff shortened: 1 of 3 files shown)


src/js/media-chrome-range.ts
```diff
@@ -121,6 +121,7 @@ template.innerHTML = /*html*/ `
 
     #appearance {
       height: var(--media-range-track-height, 4px);
+      transition: var(--media-range-track-transition);
```
luwes, line 124: this was functionality was removed with the re-architecture. 
just making it work again here.

it's possible to animate the track via height but it's not so smooth as doing it with a transform.

src/js/media-chrome-range.ts
```diff
@@ -240,13 +240,13 @@ template.innerHTML = /*html*/ `
     <div id="startpoint"></div>
     <div id="endpoint"></div>
     <div id="appearance" part="appearance">
```
heff, line 242: Is `appearance` an undocumented part? Or should it not be exposed? It's kind of an odd concept.

heff (approved):
lgtm

src/js/media-chrome-range.ts
luwes: I tend to agree now but it was documented unfortunately so I don't think we can remove it :(

cjpillsbury: I know we don't frequently do this, but I think this is a good case to deprecate (and document the deprecation). I think several of these names may be confusing or at least not intuitive. The track-specific parts are not obviously track specific. I'm not sure why the div indicating what's been buffered is called `"highlight"`? I don't have a clear sense of what `"pointer"` actually is from the name or the description? Might be worth some time/thought/discussion getting these names right, since, per @luwes's last comment, once we add these parts + names it's a bigger deal to change or remove them.

If we _**do**_ deprecate `"appearance"` and decide to rename it in the deprecation process, I think we can just:
1. add the new part name to the attribute
2. update the docs indicating that appearance is deprecated and will be removed in a future release and they should use whatever the new part name is (which should also be documented).

luwes: Sounds good yes, I don't want to rush this. 

There's much to think about also heading in the future CSS vars vs CSS parts. And if we ever want to deprecate CSS vars in favor of a CSS part.

`highlight` is called like that because it's in the base class MediaChromeRange. 
It doesn't really know about buffering or even if it's a time range.

the appearance naming is from CSS and webkit uses this as well. it's the visual UI, the appearance :)

heff (approved):
lgtm

PR #995, opened 2024-10-03
fix: chapters event w/ preload none

luwes:
this fixes an issue where the chapters were not being loaded in the UI because the track event was not fired.


cjpillsbury (approved):
👍

PR #996, opened 2024-10-04
fix: add button tooltip part for more styling

luwes:
options like transition delay, backdrop-filter, etc...


heff (approved):
Makes sense!

PR #997, opened 2024-10-04
fix: seeking when preload is none

### LINKED ISSUE #910: Media-Time-Range scrub and skimming is broken if video is not pre-loaded, happens often on iOS 17.5
just ran into another bug? seems that the media-time-range scrub and skimming is broken with the latest iOS update? Running iOS 17.5. It looks like it skims and updates, but one play it reverts back to the original playhead position?
Seems to be only an issue before the video starts play. And leaves the UI timecode vulnerable to being completely out of sync and just wrong

---

Some further research and testing from my end, it seems to happen most frequently when the video hasn't started loading for some reason. 

You can see it happen even on desktop in the media-chrome examples:
https://media-chrome.mux.dev/examples/vanilla/state-change-events-demo.html
1. Try to scrub
2. Hit play
3. Notice scrub goes back to 0.

luwes:
related #910

this fixes an issue where when you seek before the video is loaded we don't set the video.currentTime.

it's fine to set the currentTime anytime. it doesn't throw for a valid number value. 
but users have to set the `defaultduration` in some cases. not sure with iOS and native HLS.

### DIFF
--- a/examples/vanilla/mobile.html
+++ b/examples/vanilla/mobile.html
@@ -84,7 +84,7 @@
   <body>
     <main>
       <h1>Media Chrome Standard Mobile Video Usage Example</h1>
-      <media-controller autohide="2">
+      <media-controller autohide="2" defaultduration="134">
         <video
           slot="media"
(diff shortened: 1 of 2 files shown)


cjpillsbury (approved):
👌

PR #1005, opened 2024-10-08
fix: sticky chapter on mouseout

### LINKED ISSUE #999: sticky timeline chapters
Image

they don't go back to normal height sometimes

luwes:
fix #999

I can't reproduce it anymore so hopefully fixed.

the bug happened when `updateSegments` was called multiple times, easier to reproduce with preload none. 
`insertCSSRule` created a rule every time `updateSegments` was called so it possibly happened while the user hovered over a chapter.

### DIFF
--- a/src/js/media-chrome-range.ts
+++ b/src/js/media-chrome-range.ts
@@ -1,7 +1,6 @@
 import { MediaStateReceiverAttributes } from './constants.js';
 import { globalThis, document } from './utils/server-safe-globals.js';
 import {
-  insertCSSRule,
   getOrInsertCSSRule,
   getPointProgressOnLine,
 } from './utils/element-utils.js';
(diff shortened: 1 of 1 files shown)


src/js/media-chrome-range.ts
```diff
@@ -388,6 +387,10 @@ class MediaChromeRange extends globalThis.HTMLElement {
     this.#cssRules.pointer = getOrInsertCSSRule(this.shadowRoot, '#pointer');
     this.#cssRules.progress = getOrInsertCSSRule(this.shadowRoot, '#progress');
     this.#cssRules.thumb = getOrInsertCSSRule(this.shadowRoot, '#thumb');
+    this.#cssRules.activeSegment = getOrInsertCSSRule(
```
cjpillsbury, line 390: SMRT.

cjpillsbury (approved):
Makes sense as a change per earlier out of band convo.

PR #1012, opened 2024-10-24
fix: iOS volume support bug

### LINKED ISSUE #1011: mediavolumeunavailable not added on iOS
this should be added as an attribute if changing volume via software is not available. on most mobile devices.

open https://player-style-e61wlc1tb-mux.vercel.app/themes/halloween on iOS see volume range candle not hiding!

luwes:
fix #1011


cjpillsbury (approved):
works for me!

PR #1013, opened 2024-10-24
feat: add range `thumb` slot

luwes:
adds a `thumb` slot for range elements. 

this is required for some CSS features like animations because the declared keyframes have to be in the same scope as the element to which they are applied.

### DIFF
--- a/src/js/media-chrome-range.ts
+++ b/src/js/media-chrome-range.ts
@@ -177,16 +177,19 @@ template.innerHTML = /*html*/ `
       width: var(--media-range-thumb-width, 10px);
       height: var(--media-range-thumb-height, 10px);
       margin-left: calc(var(--media-range-thumb-width, 10px) / -2);
-      border: var(--media-range-thumb-border, none);
-      border-radius: var(--media-range-thumb-border-radius, 10px);
-      background: var(--media-range-thumb-background, var(--media-primary-color, rgb(238 238 238)));
-      box-shadow: var(--media-range-thumb-box-shadow, 1px 1px 1px transparent);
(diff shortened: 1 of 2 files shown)


src/js/media-chrome-range.ts
```diff
@@ -172,7 +172,8 @@ template.innerHTML = /*html*/ `
       }
     }
 
-    #thumb {
+    #thumb,
+    ::slotted([slot=thumb]) {
```
luwes, line 176: a question here is, do we want to apply default styles to the slotted thumb element? I went ahead with yes but I could see reasons for not wanting that. or part of the styles, like background color, border-radius...

luwes: we decided to not apply default stylistic styles like color, border, border-radius to the slotted thumb

src/js/media-chrome-range.ts
```diff
@@ -172,14 +172,11 @@ template.innerHTML = /*html*/ `
...
       width: var(--media-range-thumb-width, 10px);
       height: var(--media-range-thumb-height, 10px);
       margin-left: calc(var(--media-range-thumb-width, 10px) / -2);
```
heff, line 179: I guess we need to keep height/width in here because of this placement math, but that feels like it'll be annoying. Is there any other CSS magic we could use where the margin could be based off whatever width the thumb happens to be?

luwes: this worked via a CSS translate (or transform) with -50% since it's relative to itself. not to the parent like position absolute...

heff (approved):
This looks good to merge to me.

PR #1018, opened 2024-11-05
fix: tooltip overflow causing scrollbar

### LINKED ISSUE #1017: menu button tooltip creates scroll bar
the scrollbar is created by defocusing the open menu clicking outside the menu and menu button.

Image

luwes:
fix #1017


AdamJaggard (approved):
Are there any scenarios where someone might want to explicitly/purposefully position a tooltip before revealing it? which I think this solution would block. Approved either way as I don't know if that is a real problem in practice and this looks like it fixes the known issue.

PR #1024, opened 2024-11-08
feat: add media-error-dialog

luwes:
https://media-chrome-git-fork-luwes-error-dialog-mux.vercel.app/examples/vanilla/control-elements/media-error-dialog.html

use of `<media-chrome-dialog>` here:
https://media-chrome-git-fork-luwes-error-dialog-mux.vercel.app/examples/vanilla/control-elements/media-settings-menu.html

- Simplified the media-chrome-dialog removing many CSS vars that are not essential.
- Format the error message by overriding `MediaErrorDialog.formatErrorMessage(error)`
- Change the error message via slots
```html
<media-error-dialog mediaerrorcode="2">
  <div slot="error-2">This is a custom message</div>
</media-error-dialog>
```

- Supports custom media elements errors with any code and message.
- Built with React SSR compatible `getTemplateHTML` static methods (could be nice to start working towards this)


examples/vanilla/control-elements/media-error-dialog.html
```diff
@@ -0,0 +1,114 @@
...
+    <media-controller defaultsubtitles>
+      <custom-video
+        slot="media"
+        src="https://stream.mux.com/Sc89iWAyNkhJ3P1rQ02nrEdCFTnfT01CZ2KmaEcxXfB00/low.mp4"
+        muted
+        crossorigin
+      ></custom-video>
+      <img slot="poster" src="https://image.mux.com/Sc89iWAyNkhJ3P1rQ02nrEdCFTnfT01CZ2KmaEcxXfB008/thumbnail.webp?time=13" />
+      <media-loading-indicator slot="centered-chrome" noautohide></media-loading-indicator>
+      <media-error-dialog hidden></media-error-dialog>
```
luwes, line 67: I don't love that `hidden` has to be added by default. it was used on `media-chrome-dialog` similar to `menu` but maybe an `open` attribute like the native dialog could be nicer here.

src/js/media-error-dialog.ts
```diff
@@ -0,0 +1,108 @@
...
+];
+
+/**
+ * @attr {number} mediaerrorcode - (read-only) The error code for the current media error.
+ * @attr {string} mediaerrormessage - (read-only) The error message for the current media error.
+ *
+ * @cssproperty --media-control-background - `background` of control.
+ */
+class MediaErrorDialog extends MediaChromeDialog {
+  static getSlotTemplateHTML = getSlotTemplateHTML;
```
cjpillsbury, line 63: question: Any reason we're implementing this as a class-level instead of an instance level that could be overridden?

src/js/media-chrome-dialog.ts
```diff
@@ -94,19 +96,26 @@ class MediaChromeDialog extends globalThis.HTMLElement {
...
-        this.constructor as typeof MediaChromeDialog
-      ).template.content.cloneNode(true) as HTMLElement;
-      this.shadowRoot.append(this.nativeEl);
-    }
+      const attrs = namedNodeMapToObject(this.attributes);
 
-    this.addEventListener('invoke', this);
-    this.addEventListener('focusout', this);
-    this.addEventListener('keydown', this);
+      this.shadowRoot.innerHTML = /*html*/ `
```
cjpillsbury, line 114: question: this is a pretty different implementation/architecture from other "base" components (e.g. the relationship between `media-chrome-button` vs. `media-play-button`, where the `slotTemplate` is provided via the subclass's `constructor`. Any reason(s) for this different implementation approach?

src/js/labels/labels.ts
```diff
@@ -1,5 +1,14 @@
 export type LabelOptions = { seekOffset?: number; playbackRate?: number };
 
+// Setting a code explicitely to null makes the error not show up in the UI.
+export const errors = {
```
cjpillsbury, line 4: question: is there a benefit to not implementing this as a function that accepts e.g. an "error-shaped object" or similar and yields a (potentially empty) string? That would be consistent with other label code below and should simplify the error dialog impl

src/js/media-chrome-dialog.ts
```diff
@@ -1,49 +1,61 @@
...
-        opacity .2s ease-out,
-        transform .15s ease-out
-      ) !important;
-      ${/* ^^Prevent transition override by media-container */ ''}
-      visibility: var(--media-dialog-visibility, visible);
-      opacity: var(--media-dialog-opacity, 1);
-      transform: var(--media-dialog-transform-in, translateY(0) scale(1));
-    }
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+function getTemplateHTML(_attrs: Record<string, string>) {
```
cjpillsbury, line 10: question: why not handle this more like we handle `media-chrome-button` with subclasses? Any particular advantages/disadvantages?

src/js/media-chrome-dialog.ts
```diff
@@ -94,19 +96,26 @@ class MediaChromeDialog extends globalThis.HTMLElement {
...
-      ).template.content.cloneNode(true) as HTMLElement;
-      this.shadowRoot.append(this.nativeEl);
-    }
+      const attrs = namedNodeMapToObject(this.attributes);
 
-    this.addEventListener('invoke', this);
-    this.addEventListener('focusout', this);
-    this.addEventListener('keydown', this);
+      this.shadowRoot.innerHTML = /*html*/ `
+        ${(this.constructor as typeof MediaChromeDialog).getTemplateHTML(attrs)}
```
cjpillsbury, line 115: question: assuming we keep something roughly resembling this architecture, why use attributes instead of the type-accurate properties for these functions?

src/js/labels/labels.ts
cjpillsbury: I'm now seeing we're also relying on this string to be a mini "domain specific language"/syntax in the error dialog parsing code. My S1 is to break these out into two functions - one for a default title by error shape (code and/or message), another for a default verbose message by error shape (code and/or message).

src/js/media-error-dialog.ts
```diff
@@ -0,0 +1,108 @@
...
+    </slot>
+  `;
+}
+
+function shouldHideErrorDialog(errorCode?: number) {
+  return !errorCode || errors[errorCode] === null;
+}
+
+function getErrorMessage(errorCode?: number | string, errorMessage?: string) {
+  const message: string = errors[+errorCode] ?? errorMessage ?? '';
```
cjpillsbury, line 39: question: does this mean we have to provide a message in the format `Title:Verbose Message` for a title to be used? Also, if we treat `errors[+errorCode]` as the preference, doesn't that mean folks modeling more verbose/context specific error messages will be ignored for the more generic ones in this case?

src/js/utils/element-utils.ts
```diff
@@ -1,6 +1,14 @@
 import { MediaStateReceiverAttributes } from '../constants.js';
 import type MediaController from '../media-controller.js';
 
+export function namedNodeMapToObject(namedNodeMap: NamedNodeMap) {
```
cjpillsbury, line 4: question: (per earlier comment) is this needed if we simply rely on properties instead of attributes?

cjpillsbury (changes requested):
Some starting questions (unrelated to prior central focus of prior convo). May not be blockers, depending

cjpillsbury:
> ```js
> import { errors } from 'media-chrome/dist/labels/labels'; 
> errors[1] = 'Abort Error'; 
> errors[2] = 'Network Error';
> ```

I may be wrong, but it's my impression that mutating module-level exports is typically considered an anti-pattern unless there are good reasons to do it. If we want a recommended update, this should probably be treated as a default implementation (I think ideally as functions defining default title translation and default content/verbose message translation) on media-error-dialog that can be overridden.

examples/vanilla/control-elements/media-error-dialog.html
cjpillsbury: Yeah modeling it more like https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog feels right to me.

cjpillsbury:
In the example, the default UI/behavior:
1. does not extend to the full size of the `media-controller`
2. allows users to interact with controls (minimally because they aren't covered by the error dialog)

idea: we can use our convention of adding default styles to media-container for media-error-dialog to make it cover the entirety of media-controller by default potentially allowing those styles to be overridable via e.g. css vars

src/js/media-chrome-dialog.ts
luwes: yes, the reason is that passing args through the constructor is not really a good practice. I might have mentioned this before. basically if you create a DOM element there is no possibility to pass arguments. `document.createElement('my-button')`. I know it's a base class and it's possible with the constructor like we do but still this is more of an unexpected API than using a static method and overriding static methods in the child class. I've been wanting to change those base classes for a while and this new element seemed like a good candidate to start this process.

src/js/media-error-dialog.ts
luwes: yes, it's on the class level so it could be called by React for example to render DSD (declarative shadow dom) without having to create an instance of this element.

overriding is still possible by child classes by redeclaring the static method.

src/js/media-chrome-dialog.ts
luwes: yes, this is forward looking for SSR DSD in React components like we do in the media elements repo.

src/js/media-error-dialog.ts
luwes: yes, the message will need a colon for a title to be used.

yes, my first thought was that it's a nicer experience to have consistent error messages across browsers because they're different. 

it's a valid concern, it is possible to clear the errors labels and then it will fallback to the message in the error object but that's not up to the custom media element. we could potentially make use of detecting if the error event is a native event or user event (isTrusted)?

heff (commented):
Partial review here

src/js/media-chrome-dialog.ts
luwes: this will also be used to SSR so it should output plain HTML

examples/vanilla/control-elements/media-error-dialog.html
```diff
@@ -0,0 +1,114 @@
...
+        }
+      }
+
+      globalThis.customElements.define('custom-video', CustomVideo);
+    </script>
+
+    <style>
+      /* Hide custom elements that are not defined yet */
+      :not(:defined) {
+        display: none;
```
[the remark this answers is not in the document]
luwes: it wouldn't work properly because the JS of media-chrome takes a while to load. 
that CSS has to be set while parsing the page

examples/vanilla/control-elements/media-error-dialog.html
```diff
@@ -0,0 +1,114 @@
...
+      }
+
+      h2 {
+        margin-top: 2rem;
+      }
+
+      /** add styles to prevent CLS (Cumulative Layout Shift) */
+      media-controller:not([audio]) {
```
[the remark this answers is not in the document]
luwes: same here, that CSS has to be set while parsing the page and the elements

examples/vanilla/control-elements/media-settings-menu.html
```diff
@@ -101,7 +101,6 @@ <h2>menu children</h2>
       <media-loading-indicator slot="centered-chrome" noautohide></media-loading-indicator>
       <media-chrome-dialog
         id="settings"
-        style="position: absolute; top: 5%;"
         hidden
```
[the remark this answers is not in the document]
luwes: I think we're all for. I'll change it

src/js/media-chrome-dialog.ts
```diff
@@ -59,19 +71,7 @@ export const Attributes = {
  * @cssproperty --media-secondary-color - Default color of background.
  * @cssproperty --media-text-color - `color` of text.
  *
- * @cssproperty --media-control-background - `background` of control.
  * @cssproperty --media-dialog-display - `display` of dialog.
- * @cssproperty --media-dialog-background - `background` of dialog.
```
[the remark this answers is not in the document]
luwes: yes, most of these can be set on the element itself without CSS vars.
I wanted to start with as little CSS vars as possible with this new element.

src/js/media-chrome-dialog.ts
```diff
@@ -94,19 +96,26 @@ class MediaChromeDialog extends globalThis.HTMLElement {
   constructor() {
     super();
 
+    this.addEventListener('invoke', this);
+    this.addEventListener('focusout', this);
+    this.addEventListener('keydown', this);
```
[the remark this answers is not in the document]
luwes: this makes it call `handleEvent` on the this object.

it can be handy so child classes can override the handleEvent method. 
this technique is used in a few different components.

examples/vanilla/control-elements/media-error-dialog.html
```diff
@@ -0,0 +1,114 @@
...
+        get error() {
+          return this.#error ?? super.error;
+        }
+
+        set error(value) {
+          this.#error = value;
+        }
+      }
+
+      globalThis.customElements.define('custom-video', CustomVideo);
```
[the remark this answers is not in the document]
luwes: yes, custom-media-element doesn't define those by default anymore because it's kind of meant to be extended.

src/js/media-chrome-dialog.ts
```diff
@@ -124,16 +133,16 @@ class MediaChromeDialog extends globalThis.HTMLElement {
...
     }
   }
 
-  attributeChangedCallback(
-    attrName: string,
-    oldValue: string | null,
-    newValue: string | null
-  ) {
+  attributeChangedCallback(attrName: string, oldValue: string | null, newValue: string | null) {
+    this.#init();
```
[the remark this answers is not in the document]
luwes: yep, we want to render the shadow dom HTML as soon as the attributes are ready or when connected to the DOM

luwes:
> > ```
> > <media-error-dialog mediaerrorcode="2">
> >   <div slot="error-2">This is a custom message</div>
> > </media-error-dialog>
> > ```
> 
> Does this mean that the slot-based override only supports a single hard-coded string for errors with a particular code?

no here you're only limited by HTML. you can add multiple elements with the error-2 slot attribute or include any HTML in that div above, images, svg's etc.

src/js/media-chrome-dialog.ts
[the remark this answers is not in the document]
heff: lol, how have I gone this long without knowing about `handleEvent`! 👍

src/js/media-chrome-dialog.ts
cjpillsbury: So would your recommendation longer term be to e.g. refactor our button class defs to more resemble this architecture?

src/js/media-chrome-dialog.ts
cjpillsbury: Cool that def makes sense with the additional arguments. Sold.

src/js/media-error-dialog.ts
cjpillsbury: I'm thinking particularly about the far more verbose/"bespoke" error messages and titles in `mux-player`, fwiw

src/js/utils/element-utils.ts
cjpillsbury: Resolving this per prior comments. 👌

cjpillsbury:
> no here you're only limited by HTML. you can add multiple elements with the error-2 slot attribute or include any HTML in that div above, images, svg's etc.

I was maybe unclear. Let's take a concrete scenario:
1. I have a custom media element (let's call it `mux-video`) where I have enough context to have more detailed messages for a variety of things that would fall under `MEDIA_ERR_NETWORK` (e.g. I know the network request failed because of details about the JWT, or that it was specifically an error while requesting a DRM license, etc. etc.)
2. I want to be able to provide that granular, technical error message to some analytics integration (let's call it Mux Data)
3. I want to be able to provide some more "user friendly" error message to end users via an error dialog

How does this work for either
1. the mutative case for the errors object from the labels module (which I'm still not a fan of? I understand there are examples in the wild)
2. slots or similar component-level customization
?

src/js/media-error-dialog.ts
cjpillsbury: the `isTrusted` is an interesting solution... I think I like it (independent of whether this should be handled as a function vs. a lookup).

luwes:
> > no here you're only limited by HTML. you can add multiple elements with the error-2 slot attribute or include any HTML in that div above, images, svg's etc.
> 
> I was maybe unclear. Let's take a concrete scenario:
> 
> 1. I have a custom media element (let's call it `mux-video`) where I have enough context to have more detailed messages for a variety of things that would fall under `MEDIA_ERR_NETWORK` (e.g. I know the network request failed because of details about the JWT, or that it was specifically an error while requesting a DRM license, etc. etc.)
> 2. I want to be able to provide that granular, technical error message to some analytics integration (let's call it Mux Data)
> 3. I want to be able to provide some more "user friendly" error message to end users via an error dialog
> 
> How does this work for either
> 
> 1. the mutative case for the errors object from the labels module (which I'm still not a fan of? I understand there are examples in the wild)
> 2. slots or similar component-level customization
>    ?

I think the Mux Data case shouldn't be part of Media Chrome, that's up to the video engine logic to translate the error message to a technical analytics error. we do this in mux-video as well.

yes, currently to translate the error dialog message is to mutate the error labels.

I think I'm on board with changing it on the component level. it would work similarly as how the menu item text can be changed. 
https://github.com/muxinc/media-chrome/blob/f99f8edf2ab1440b4f8b4d014320e784c4e746d3/src/js/menu/media-chrome-menu.ts#L308-L308
https://github.com/muxinc/media-chrome/blob/92b6bd6e93cfbcfcd06e24dc5507c37acd6b1491/src/js/menu/media-rendition-menu.ts#L141-L144

luwes:
> In the example, the default UI/behavior:
> 
> 1. does not extend to the full size of the `media-controller`
> 2. allows users to interact with controls (minimally because they aren't covered by the error dialog)
> 
> idea: we can use our convention of adding default styles to media-container for media-error-dialog to make it cover the entirety of media-controller by default potentially allowing those styles to be overridable via e.g. css vars

updated here [`07cb4cf` (#1024)](https://github.com/muxinc/media-chrome/pull/1024/commits/07cb4cf4c40b65f7425fb1f981a54d5193bc06f0)
I didn't add new CSS vars because it still possible to override the CSS just via an element selector.

src/js/media-container.ts
```diff
@@ -200,6 +200,12 @@ template.innerHTML = /*html*/ `
     ::slotted([role="dialog"]) {
       align-self: center;
     }
+
+    ::slotted(media-error-dialog) {
+      position: absolute;
+      inset: 0;
+      z-index: 2;
+    }
```
luwes, line 208: @cjpillsbury did you have this in mind with covering all controls?

src/js/media-store/util.ts
```diff
@@ -85,6 +85,8 @@ export const toggleSubtitleTracks = (stateOwners, force: boolean): void => {
 export const areValuesEq = (x: any, y: any): boolean => {
   // If both are strictly equal, they're equal
   if (x === y) return true;
+  // If either is null, they're not equal
+  if (x == null || y == null) return false;
```
luwes, line 89: added tests below for areValuesEq

examples/vanilla/control-elements/media-error-dialog.html
```diff
@@ -0,0 +1,114 @@
...
+    </script>
+
+    <style>
+      /* Hide custom elements that are not defined yet */
+      :not(:defined) {
```
cjpillsbury, line 27: nitpick(non-blocking): This should probably be applied specifically to `media-error-dialog` for what we're demonstrating, right? Or is there a reason you hide every custom element for this demo?

src/js/labels/labels.ts
cjpillsbury: Still not a fan of this being an object with the assumption of potentially mutating it, but I won't press here.

src/js/media-chrome-dialog.ts
```diff
@@ -1,52 +1,68 @@
...
-        visibility 0s,
-        opacity .2s ease-out,
-        transform .15s ease-out
-      ) !important;
-      ${/* ^^Prevent transition override by media-container */ ''}
-      visibility: var(--media-dialog-visibility, visible);
-      opacity: var(--media-dialog-opacity, 1);
-      transform: var(--media-dialog-transform-in, translateY(0) scale(1));
-    }
+function getTemplateHTML(_attrs: Record<string, string>) {
```
cjpillsbury, line 11: suggestion: add a code comment documenting that this is intended to be used as a mixin static method on a class that also has a static method called `getSlotTemplateHTML()` somewhere in the class inheritance chain. I suspect the relationships between these two when subclasses are involved will not be super intuitive (definitely required some chewing on my part for the different permutations).

src/js/media-chrome-dialog.ts
```diff
@@ -80,12 +86,14 @@ export const Attributes = {
  * @cssproperty --media-text-content-height - `line-height` of text.
  */
 class MediaChromeDialog extends globalThis.HTMLElement {
-  static template: HTMLTemplateElement = template;
+  static getTemplateHTML = getTemplateHTML;
+  static getSlotTemplateHTML = getSlotTemplateHTML;
 
   static get observedAttributes() {
-    return [Attributes.HIDDEN, Attributes.ANCHOR];
+    return [Attributes.OPEN, Attributes.ANCHOR];
```
cjpillsbury, line 93: 😎

src/js/media-chrome-dialog.ts
```diff
@@ -1,52 +1,68 @@
...
+    <style>
+      :host {
+        font: var(--media-font,
+          var(--media-font-weight, normal)
+          var(--media-font-size, 14px) /
+          var(--media-text-content-height, var(--media-control-height, 24px))
+          var(--media-font-family, helvetica neue, segoe ui, roboto, arial, sans-serif));
+        color: var(--media-text-color, var(--media-primary-color, rgb(238 238 238)));
+        display: var(--media-dialog-display, inline-flex);
+        width: 100%;
```
cjpillsbury, line 22: question: should these be defined here or in `media-container` via a selector?

src/js/media-chrome-dialog.ts
```diff
@@ -1,52 +1,68 @@
...
+        height: 100%;
+        justify-content: center;
+        align-items: center;
+        transition: visibility .15s ease-in, opacity .15s ease-in, transform .15s ease-in !important;
+        visibility: hidden;
```
cjpillsbury, line 27: 😎 but also
question: this means that the dialog will still take up space but just not be visible. Seems sub-optimal for the broader idea here I think? cc @heff

package.json
```diff
@@ -163,7 +163,8 @@
     ],
     "rules": {
       "@typescript-eslint/ban-ts-comment": "off",
-      "@typescript-eslint/no-explicit-any": "off"
+      "@typescript-eslint/no-explicit-any": "off",
+      "@typescript-eslint/no-unused-vars": ["error", { "argsIgnorePattern": "^_" }]
```
cjpillsbury, line 167: praise: 😍 thank you for gutting all of my one offs with this single line.

cjpillsbury (changes requested):
Haven't reviewed elements side yet but here's what I've got here. Overall this is looking positive, though!

src/js/media-error-dialog.ts
```diff
@@ -0,0 +1,130 @@
...
+ */
+class MediaErrorDialog extends MediaChromeDialog {
+  static getSlotTemplateHTML = getSlotTemplateHTML;
+  static formatErrorMessage = formatErrorMessage;
+
+  static get observedAttributes() {
+    return [...super.observedAttributes, ...observedAttributes];
+  }
+
+  #mediaError: MediaErrorLike | null = null;
```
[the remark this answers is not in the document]
cjpillsbury: Resolving this one. We inconsistently use `null` and `undefined`. Different pros/cons for each. Should probably discuss in MCOH.

src/js/media-chrome-dialog.ts
```diff
@@ -1,52 +1,68 @@
...
+        ${/* Prevent opacity override by media-container */ ''}
+        opacity: 0 !important;
```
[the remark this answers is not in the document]
luwes: ah yes, we could possibly use [role=dialog] to exclude. will do that

src/js/media-container.ts
```diff
@@ -298,6 +304,7 @@ class MediaContainer extends globalThis.HTMLElement {
               MediaUIAttributes.MEDIA_CHAPTERS_CUES,
               MediaUIAttributes.MEDIA_WIDTH,
               MediaUIAttributes.MEDIA_HEIGHT,
+              MediaUIAttributes.MEDIA_ERROR_MESSAGE,
```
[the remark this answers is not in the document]
luwes: yes, `mediaerrorcode` and `mediaerrormessage` on the media-error-dialog and only `mediaerrorcode` on the media-controller. if it's in one attribute we couldn't exclude the message separately from media-controller. 
`mediaerrorcode` feels like it could be useful on media-controller but the message seems noisy and not very useful there.

src/js/media-container.ts
luwes: yes right, adding new slot / layer? cc @heff

src/js/media-store/util.ts
luwes: how so? we want check both on `null` and `undefined`. with the condition right above, if either is null or undefined we can safely say they are not equal...

src/js/media-error-dialog.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  attributeChangedCallback(attrName: string, oldValue: string | null, newValue: string | null) {
+    super.attributeChangedCallback(attrName, oldValue, newValue);
+
+    // Add this conditional to prevent endless loop by setting the hidden attribute.
+    if (!observedAttributes.includes(attrName)) return;
```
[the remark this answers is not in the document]
luwes: still required, updated comment

src/js/media-chrome-dialog.ts
luwes: good call, I think in this case because it's a dialog element which generally covers the whole window (the dialog background at least) it seems correct.
see https://media-chrome-git-fork-luwes-error-dialog-mux.vercel.app/examples/vanilla/control-elements/media-settings-menu.html for an example of `media-chrome-dialog`

src/js/media-store/util.ts
cjpillsbury: ah you're right since this is specifically for an early bail, we should be good. Resolving.

src/js/media-container.ts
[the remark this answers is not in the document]
cjpillsbury: yup agree. good call.

src/js/media-chrome-dialog.ts
cjpillsbury: cool just wanted to call it out. I can def see arguments for defaulting to 100% though, so no need to bikeshed.

src/js/media-chrome-dialog.ts
luwes: I ended up using `transition-behavior: allow-discrete;`. Otherwise it'll become messy to update the CSS props via JS and on transition end event etc.

src/js/media-store/media-store.ts
```diff
@@ -495,7 +492,8 @@ const createMediaStore = ({
       // For any state change request "actions"/"events" of media (and related) state,
       // these are handled by the `RequestMap`, which defines a function for a given change request type
       // that is responsible for what should happen as a result
-      if (requestMap[type]) {
+      // If a fatal error occurred, we should not process any more state change requests.
```
cjpillsbury, line 495: suggestion(non-blocking):

```suggestion
      // If a fatal error occurred, we should not process any more state change requests,
      // but we should process updates to state owners or options/defaults, handled below,
      // which is why we don't simply early bail here.
```

cjpillsbury (approved):
One code comment update suggestion for clarity/additional context. Otherwise 🚀 lookin' 😎.

heff (commented):
lgtm except this should show the error title
image

Also we're swallowing the custom error code and that doesn't seem right. I think we should instead create a title for unknown errors like "Error Code 42".

image

In the future we could give devs a way to override the title of a custom code.

examples/vanilla/control-elements/media-error-dialog.html
```diff
@@ -0,0 +1,109 @@
...
+
+      <h2>Error 4 (MEDIA_ERR_SRC_NOT_SUPPORTED)</h2>
+      <media-error-dialog mediaerrorcode="4"></media-error-dialog>
+
+      <h2>Error 5 (MEDIA_ERR_ENCRYPTED)</h2>
+      <media-error-dialog mediaerrorcode="5"></media-error-dialog>
+
+      <h2>Error 2 with custom message via <code>error-2</code> slot</h2>
+      <media-error-dialog mediaerrorcode="2">
+        <div slot="error-2">This is a custom message</div>
```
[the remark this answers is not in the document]
cjpillsbury: @heff This is related to a prior convo on whether the slot and formatting should more explicitly split apart things like:
1. strings of title + message vs. (string representation of) HTML
2. html for title vs. html for message (regardless of 1)
3. separate slots for each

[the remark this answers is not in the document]
luwes: yes, here they have full control of the content that is shown in the dialog. they can easily add a title, message, links, etc.

```html
<h3 slot="error-2">Network Error</h3>
<p slot="error-2">This is the error message</p>
```

or wrap it in a div 

```html
<div slot="error-2">
  <h3>Network Error</h3>
  <p>This is the error message</p>
</div>
```

src/js/media-error-dialog.ts
```diff
@@ -0,0 +1,108 @@
...
+      #content {
+        background: var(--media-control-background, var(--media-secondary-color, rgb(20 20 30 / .8)));
+        display: block;
+        padding: 1.2em 1.5em;
+      }
+
+      h3,
+      p,
+      ::slotted(h3),
+      ::slotted(p) {
```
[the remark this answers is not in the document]
luwes: I'll update it. I added it to easily get the same style when using slotted error content elements.

cjpillsbury (approved):

src/js/media-container.ts
```diff
@@ -188,18 +188,24 @@ template.innerHTML = /*html*/ `
...
-}]) slot[name=poster] {
+      MediaUIAttributes.MEDIA_HAS_PLAYED
+    }]) slot[name=poster] {
       display: none;
     }
 
-    ::slotted([role="menu"]) {
-      align-self: end;
+    slot[name=dialog]::slotted(*) {
+      z-index: 2;
```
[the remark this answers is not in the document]
luwes: in media-chrome-range the input[type=range] element is used as a hover zone which is 14px higher than the actual media-chrome-range. helps with dragging on mobile. it's primarily for layouts where the timeline goes right above the buttons and then it's better that the timeline slightly overlaps the buttons.

I can see the concerns though, we already had a report from Phil on this in the Mux dashboard with Mux players dialog being on `z-index: 100`.

I'll see if there is a workaround.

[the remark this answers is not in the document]
luwes: I did some testing on Mux player with the z-indexes removed and that worked fine because we set them in the themes. maybe that's right approach. leave it up to the consumer of the MC elements? 
I removed all the z-indexes in the last commit.

heff (approved):

Issue #886, opened 2024-04-30
Can't import CJS version using Jest


tzachyrm:
We're importing some React components, e.g.

```js
import { MediaController } from "media-chrome/react";
```

In our production build (Webpack, ESM) this works fine, but when using Jest (which uses CJS), we receive the following error:

```
TypeError: Class extends value #<Object> is not a constructor or null

at Object.<anonymous> (node_modules/.pnpm/media-chrome@3.2.1/node_modules/media-chrome/dist/cjs/media-preview-chapter-display.js:56:68)
      at Object.<anonymous> (node_modules/.pnpm/media-chrome@3.2.1/node_modules/media-chrome/dist/cjs/index.js:103:52)
      at Object.<anonymous> (node_modules/.pnpm/media-chrome@3.2.1/node_modules/media-chrome/dist/cjs/react/index.js:75:16)
```

I've narrowed it down to a single problematic line in `media-chrome/dist/cjs/media-preview-chapter-display.js`:

```
var import_media_text_display = __toESM(require("./media-text-display.js"), 1);
```

Changing it to  `var import_media_text_display = require("./media-text-display.js");` seems to fix the issue, at least in my case.

I'm not an expert, but this seems to be an esbuild issue.

media-chrome version: 3.2.1
node version: v18.14.2
jest version: 29.7.0

JuhG:
We were running into the same issue.

Changing our babel config fixed this. This was our old config:
```tsx
 presets: [
    [
      "@babel/preset-env",
      { modules: "commonjs" },
    ],
    // ...
  ],
```

Removing the `modules: "commonjs"` part fixed the issue (which now defaults to `auto`), but not sure how it affects everything else. Looks like there's an issue with running `media-chrome` in a CJS module system.

PR #1040, opened 2024-12-18
Made createMediaStore a named import

szenasitibor:
I'm having the same problem, as the described in [this](https://github.com/muxinc/media-chrome/issues/886) issue.

The error I receive:
```
media-controller.js:440 Uncaught TypeError: (0 , import_media_store.default) is not a function
    at MediaController.setupDefaultStore_fn (media-controller.js:440:1)
    at MediaController.connectedCallback (media-controller.js:219:1)
```

Which shows me, that the problem is here:
`var import_media_store = __toESM(require("./media-store/media-store.js"), 1);`

By making **createMediaStore** a named export, and using that, this problem disappears, and the video player works perfectly.

### DIFF
--- a/src/js/media-controller.ts
+++ b/src/js/media-controller.ts
@@ -28,7 +28,7 @@ import {
   setNumericAttr,
   setStringAttr,
 } from './utils/element-utils.js';
-import createMediaStore, { MediaStore } from './media-store/media-store.js';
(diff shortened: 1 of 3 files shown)


luwes (approved):
LGTM! sorry for the delay

Issue #874, opened 2024-04-02
Variable for setting transition duration when hiding media controls


intelligence:
AFAIK these values are hardcoded and cannot be changed via CSS variables at the moment?

My workaround:
```
.mux-player[userinactive]:not([mediapaused]) media-control-bar {
  transition: opacity 0.125s ease-in-out;
}
```

PR #1051, opened 2025-01-14
feat: add CSS variable to customize the <media-controller> hide transition

R-Delfino95:
Related to #874
- Added `--media-control-transition-out` CSS variable to optionally change the transition effect when hiding the `<media-controller>` component.
- Added `--media-control-transition-in` CSS variable to optionally change the transition effect when showing the `<media-controller>` component.
- Added example of use in `basic-video.html`

### DIFF
--- a/src/js/media-container.ts
+++ b/src/js/media-container.ts
@@ -153,7 +153,7 @@ template.innerHTML = /*html*/ `
       Attributes.NO_AUTOHIDE
     }]):not([hidden]):not([role=dialog])) {
       opacity: 1;
-      transition: opacity 0.25s;
+      transition: var(--media-control-transition-in, opacity 0.25s);
     }
 
     ${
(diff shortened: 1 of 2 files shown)


src/js/media-container.ts
```diff
@@ -169,7 +169,7 @@ template.innerHTML = /*html*/ `
       Attributes.NO_AUTOHIDE
     }]):not([role=dialog])) {
       opacity: 0;
-      transition: opacity 1s;
+      transition: var(--media-hide-transition, opacity 1s);
```
luwes, line 172: this is good but I would change the naming to include control. we've also used `in` and `out` naming for transitions to make that distinction. @heff @cjpillsbury thoughts?

I propose making this CSS var `--media-control-transition-out` and then the transition above we can make `--media-control-transition-in`

R-Delfino95: Great! I renamed the variable and created the variable for the show transition

luwes (approved):
LGTM! thanks

PR #1057, opened 2025-01-23
fix: set breakpoints on connect

luwes:


mihar-22:
I'd recommend moving the entire [resize and mutation observer setup](https://github.com/muxinc/media-chrome/blob/1565149577061ec78ac69b2921245c70bff77181/src/js/media-container.ts#L326-L402) to the `connectedCallback`.

Few points:

1. Observers in the constructor are surprising, is there a reason? Usually the constructor is one-time setup logic that doesn't require cleaning up. In addition, any size measurements or layout calculations in the constructor would be incorrect or incomplete, since the element doesn’t have its final rendering context yet (no offset parent, no actual dimensions, etc.). One last small point is that if our elements are registered async then heavy work done in the constructor will block browser parsing.

2. I'm not seeing the cleanup logic on disconnect?

3. We should probably refactor the callbacks as class methods to avoid creating these functions for every instance.

luwes:
> PR looks good! This is the earliest time in which we can accurately get the container sizing.
> 
> I'd also recommend moving the entire [resize and mutation observer setup](https://github.com/muxinc/media-chrome/blob/1565149577061ec78ac69b2921245c70bff77181/src/js/media-container.ts#L326-L402) to the `connectedCallback`.
> 
> Few points:
> 
> 1. Observers in the constructor are surprising, is there a reason? Usually the constructor is one-time setup logic that doesn't require cleaning up. In addition, any size measurements or layout calculations in the constructor would be incorrect or incomplete, since the element doesn’t have its final rendering context yet (no offset parent, no actual dimensions, etc.). One last small point is that if our elements are registered async then heavy work done in the constructor will block browser parsing.
> 2. I'm not seeing the cleanup logic on disconnect?
> 3. We should probably refactor the callbacks as class methods to avoid creating these functions for every instance.

valid points though the constructor is called on an upgrade and most of the time that is on connected. I believe we put it there because it is just setup once but it could be an oversight. it doesn't cause any serious issues but a new PR putting it in connectedCallback seems good to me. 

last I checked the cleanup logic is not absolutely required because the browser cleans that up automatically when the element is garbage collected.

let's keep those changes for a new PR.

PR #1061, opened 2025-01-28
fix: remove media container observers on disconnect

mihar-22:
As discussed in https://github.com/muxinc/media-chrome/pull/1057.

Apologies, not sure why it's auto-formatting the innerHTML template string, am I missing an extension or something?

To answer a few questions from the other PR:

> valid points though the constructor is called on an upgrade and most of the time that is on connected.

Generally true, but not when the element is created in-memory (`document.createElement` or via template) and not yet attached to the DOM. Quite common practice in modern frameworks (e.g., Vue/Solid/Svelte).

> I believe we put it there because it is just setup once but it could be an oversight. it doesn't cause any serious issues but a new PR putting it in connectedCallback seems good to me. last I checked the cleanup logic is not absolutely required because the browser cleans that up automatically when the element is garbage collected.

That is true but it's still best practice to clean up on disconnect. Element may be held in memory as it's being moved in the DOM, might be offloaded in a scroll container as it goes out of view, detached by a framework as a performance optimization, and many other reasons.

### DIFF
--- a/src/js/media-container.ts
+++ b/src/js/media-container.ts
@@ -10,7 +10,7 @@
 import { globalThis, document } from './utils/server-safe-globals.js';
 import { MediaUIAttributes, MediaStateChangeEvents } from './constants.js';
 import { nouns } from './labels/labels.js';
-import { observeResize } from './utils/resize-observer.js';
+import { observeResize, unobserveResize } from './utils/resize-observer.js';
 // Guarantee that `<media-gesture-receiver/>` is available for use in the template
 import './media-gesture-receiver.js';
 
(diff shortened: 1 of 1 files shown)


luwes:
> Apologies, not sure why it's auto-formatting the innerHTML template string, am I missing an extension or something?

no worries about that, I think I manually corrected that last time. Prettier is making that quite ugly :)

> Generally true, but not when the element is created in-memory (document.createElement or via template) and not yet attached to the DOM. Quite common practice in modern frameworks (e.g., Vue/Solid/Svelte).

that's not correct, yes it is possible but it requires a manual https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry/upgrade for that to happen. by default the upgrade is only done on connecting to the DOM, and the constructor will also be called at that time.

LGTM! I'll leave it for 1 other person to review

cjpillsbury (approved):
It's been awhile since thinking about these contortionist moves 😅. I can't recall whether or not one of the motivations for doing this in the constructor was to handle the possibility of doing a "first round" computation earlier than `connectedCallback`/DOM attachment (given the async crud) to reduce chances of CLS. Overall this approach makes sense to me and we can always re-solve the first pass case with this refactor if needed. AKA LGMT.

Pintorado (approved):

mihar-22:
> that's not correct, yes it is possible but it requires a manual https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry/upgrade for that to happen. by default the upgrade is only done on connecting to the DOM, and the constructor will also be called at that time.

Sorry I wasn't clear. I was referring to the `constructor` and `connectedCallback` timing _after_ the initial upgrade. It's all cursed voodoo magic anyway. You can see below that it can be constructed without being connected to the DOM in-memory.

Calling `upgrade`:

image

<br />

Simply appending to DOM:

image

PR #1065, opened 2025-01-31
fix: hide error dialog on playback recovery

luwes:
this change hides the error dialog when playback recovers (playing event fires and media state indicates playing).

### DIFF
--- a/src/js/media-store/util.ts
+++ b/src/js/media-store/util.ts
@@ -2,10 +2,6 @@ import { TextTrackKinds, TextTrackModes } from '../constants.js';
 import { getTextTracksList, updateTracksModeTo } from '../utils/captions.js';
 import { TextTrackLike } from '../utils/TextTrackLike.js';
 
-export const isMediaPlaying = (media) => {
-  return !media.paused && !media.ended && media.readyState > 2;
-}
-
 export const getSubtitleTracks = (stateOwners): TextTrackLike[] => {
   return getTextTracksList(stateOwners.media, (textTrack) => {
     return [TextTrackKinds.SUBTITLES, TextTrackKinds.CAPTIONS].includes(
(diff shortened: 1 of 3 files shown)


src/js/media-store/util.ts
```diff
@@ -2,6 +2,10 @@ import { TextTrackKinds, TextTrackModes } from '../constants.js';
 import { getTextTracksList, updateTracksModeTo } from '../utils/captions.js';
 import { TextTrackLike } from '../utils/TextTrackLike.js';
 
+export const isMediaPlaying = (media) => {
+  return !media.paused && !media.ended && media.readyState > 2;
```
heff, line 6: Do we need this? None of these should ever be false when the `playing` event is fired.

heff (approved):
👍 Feel free to merge but one question about the necessity of `isMediaPlaying`.

heff:
For `<mux-video>` we still need to/are planning to implement a `warning` API for non-fatal errors, right?

src/js/media-store/util.ts
luwes: true, good call. I later added the `playing` event.type check and didn't realize

luwes:
> For `<mux-video>` we still need to/are planning to implement a `warning` API for non-fatal errors, right?

in mux-video it's possible to log out warnings with the `debug` attribute directly from hls.js.

from my PR here https://github.com/muxinc/elements/pull/1014

the warning API is not implemented because after closer inspection we don't really do much with non-fatal errors.

the only thing is doing a console.warn here:
https://github.com/muxinc/elements/blob/2343df80fddcbf05485ed9d4ab27f1bb0fac04d0/packages/mux-player/src/index.ts#L502-L502

we can easily do that in playback-core if we want?

PR #1071, opened 2025-02-11
Add multi language support for labels

ronalduQualabs:
Screenshot 2025-02-11 at 3 09 23 PM

Add multi language support for labels by: 

- Create a Translations Helper that:

   - Read language of the browser 
   - Set language for the browser or set english as default

- Add string languages files for:

  - English (Default)
  - Spanish
  - Portuguese 
  - French


src/js/utils/translation-helper.ts
```diff
@@ -0,0 +1,79 @@
+import { En } from '../lang/en.js';
+import { Es } from '../lang/es.js';
+import { Fr } from '../lang/fr.js';
+import { Pt } from '../lang/pt.js';
```
luwes, line 4: I'll have a closer look later but a blocker here is that all the languages are bundled up which will make the bundle size increase too much. 

en.js is the default and has to be included but the rest should be separately importable via 
`media-chrome/lang/xx.js` for example

they should also be able to provide their own lang file with a simple method.

```js
import { addTranslation } from 'media-chrome';

export const Pt: TranslationType = {
  tooltips: {
    enterAirplay: 'Iniciar AirPlay',
  },
  nouns: {
    audioPlayer: 'reprodutor de áudio',
  },
  verbs: {
    play: 'reproduzir',
  },
};

addTranslation(Pt);
```

luwes (changes requested):

src/js/utils/translation-helper.ts
luwes: yes, for media-chrome's own translation files we can call that `addTranslation()` call in the file.
the imports or the script tags (for HTML only) have to come before the Media Chrome custom elements markup.
that should make it possible I think right.

@ronalduQualabs make sure the lookup of the translated string is done at time of accessing the string in the UI component. Not during the file execution like it is now for `tooltipLabels` for example. It needs to be a lazy getter for this to work.

heff: > have to come before the Media Chrome custom elements markup. that should make it possible I think right
Could be. Any issues using `<script type="module">` for either in this case? Could you write out how to see this working in the HTML use case?

luwes: true the order has to be correct then, that's a bit of a bummer.

```html
<script type="module" src="https://cdn.jsdelivr.net/npm/media-chrome@4/lang/nl.js/+esm"></script>
<script type="module" src="https://cdn.jsdelivr.net/npm/media-chrome@4/+esm"></script>

<media-controller>...</media-controller>
```

nl.js

```js
import { addTranslation } from 'media-chrome/utils/translation.js';

addTranslation({
...
});
```

heff: Ok so then how does addTranslation work if media-chrome itself isn't in the page yet?

luwes: it's just a normal esm module (separate import) with a dictionary object that is shared.
first we populate the dictionary, later we access it when the main bundle is run.

src/js/utils/translation/en.ts
```diff
@@ -0,0 +1,60 @@
+import { TranslateDictionary } from '../translation-helper.js';
+
+export const En: TranslateDictionary = {
+  tooltips: {
+    enterAirplay: 'Start airplay',
```
luwes, line 5: this is starting to look good. some thoughts that came to mind.

1. this grouping / categories (tooltips, nouns, ...) is not needed and even the indexing with special keys as well
  let's make this a simple dict like
```js
export const En = {
  'Start airplay': 'Start airplay',
  ...
```
 - this reduces byte size slightly because of gzip compression 
 - it's easier for translation because the literal english word is on the left side as key.

src/js/utils/translation-helper.ts
```diff
@@ -0,0 +1,78 @@
...
+    exitFullscreen: string;
+    enterPip: string;
+    exitPip: string;
+    seekLive: string;
+    playingLive: string;
+    seekBack: string;
+    seekForward: string;
+    seconds: string;
+  };
+};
```
luwes, line 60: this type is not needed, we can infer it from the en.js structure...

src/js/utils/translation-helper.ts
```diff
@@ -0,0 +1,78 @@
...
+export const addTranslation = (
+  langCode: string,
+  languageDictionary: TranslateDictionary
+) => {
+  translationsLanguages[langCode] = languageDictionary;
+};
+
+export const getCurrentTranslations = () => {
+  const getBrowserLanguage = () => navigator.language.split('-')[0] || 'en';
+  const getPreferredLanguage = () => getBrowserLanguage();
```
luwes, line 75: let's make this function `t(key: string) => string` with key being the english variant of the translation
e.g. `const translation = i18n.t('Enter airplay')`

src/js/labels/labels.ts
```diff
@@ -25,68 +27,72 @@ export const formatError = (error: MediaErrorLike) => {
...
-  UNMUTE: 'Unmute',
-  ENTER_PIP: 'Enter picture in picture mode',
-  EXIT_PIP: 'Enter picture in picture mode',
-  PLAY: 'Play',
-  PAUSE: 'Pause',
-  PLAYBACK_RATE: 'Playback rate',
-  RENDITIONS: 'Quality',
-  SEEK_BACKWARD: 'Seek backward',
-  SEEK_FORWARD: 'Seek forward',
-  SETTINGS: 'Settings',
```
luwes, line 53: all these conversions can be removed once the previous steps are implemented

src/js/utils/translation-helper.ts
luwes: also move the inline functions outside if possible

src/js/labels/labels.ts
```diff
@@ -25,68 +27,70 @@ export const formatError = (error: MediaErrorLike) => {
...
-    `seek forward ${seekOffset} seconds`,
+    `${translation('seek forward')} ${seekOffset} ${translation('seconds')}`,
   SEEK_BACK_N_SECS: ({ seekOffset = 30 } = {}) =>
-    `seek back ${seekOffset} seconds`,
-  SEEK_LIVE: () => 'seek to live',
-  PLAYING_LIVE: () => 'playing live',
+    `${translation('seek back')} ${seekOffset} ${translation('seconds')}`,
+  SEEK_LIVE: () => translation('seek to live'),
+  PLAYING_LIVE: () => translation('playing live'),
 };
```
luwes, line 94: could you make it so these all can be removed? we can call `t()` directly in those files that need them...

make the `t()` function accept an object as second argument to fill in vars. example:

```js
`t('seek back {seekOffset} seconds', { seekOffset: 30 });`
```

here is an example of the replace regex
https://github.com/muxinc/elements/blob/2fbb2c157aa694ed82fc2b81b989cca6c0f790af/packages/playback-core/src/util.ts#L175-L177

src/js/utils/i18n.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export const addTranslation = (
+  langCode: string,
+  languageDictionary: TranslateDictionary
+) => {
+  translationsLanguages[langCode] = languageDictionary;
+};
+
+export const t = (key: TranslateKeys) => {
+  const getBrowserLanguage = () => navigator.language.split('-')[0] || 'en';
```
luwes, line 15: can this be moved in the top scope?

src/js/utils/i18n.ts
```diff
@@ -0,0 +1,19 @@
...
+  const getPreferredLanguage = () => getBrowserLanguage();
```
luwes, line 16: this is not needed, remove and use 1 name

src/js/utils/translation/en.ts
luwes, line 1: I would keep this in the `/lang` folder as well. is there a reason it's in `/utils/translation`?

src/js/lang/es.ts
```diff
@@ -0,0 +1,57 @@
+import { addTranslation } from '../utils/i18n.js';
+import { TranslateDictionary } from '../utils/translation/en.js';
```
heff, line 2: Should this be a types import only?

src/js/utils/translation/en.ts
ronalduQualabs: moved

src/js/utils/i18n.ts
ronalduQualabs: done

src/js/utils/i18n.ts
ronalduQualabs: done

src/js/labels/labels.ts
ronalduQualabs: done

src/js/utils/i18n.ts
luwes: not sure I'm seeing it

src/js/utils/i18n.ts
luwes: not seeing it

src/js/labels/labels.ts
```diff
@@ -25,68 +27,70 @@ export const formatError = (error: MediaErrorLike) => {
...
-    `seek forward ${seekOffset} seconds`,
+    `${(translation('seek forward {seekOffset} seconds'), { seekOffset })}`,
   SEEK_BACK_N_SECS: ({ seekOffset = 30 } = {}) =>
-    `seek back ${seekOffset} seconds`,
-  SEEK_LIVE: () => 'seek to live',
-  PLAYING_LIVE: () => 'playing live',
+    `${translation('seek back {seekOffset} seconds', { seekOffset })}`,
+  SEEK_LIVE: () => translation('seek to live'),
+  PLAYING_LIVE: () => translation('playing live'),
 };
```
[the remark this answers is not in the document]
ronalduQualabs: done

src/js/lang/es.ts
ronalduQualabs: done

src/js/utils/translation/en.ts
```diff
@@ -0,0 +1,59 @@
...
+  'media loading': 'media loading',
+
+  settings: 'settings',
+  'audio tracks': 'audio tracks',
+  quality: 'quality',
+  play: 'play',
+  pause: 'pause',
+  mute: 'mute',
+  unmute: 'unmute',
+  LIVE: 'LIVE',
```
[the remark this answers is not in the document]
ronalduQualabs: Done

luwes:
just resolve conflict and package.json exports entry and it should be good to go!

src/js/utils/i18n.ts
```diff
@@ -0,0 +1,27 @@
...
+};
+
+export const addTranslation = (
+  langCode: string,
+  languageDictionary: TranslateDictionary
+) => {
+  translationsLanguages[langCode] = languageDictionary;
+};
+
+const getBrowserLanguage = () => navigator.language.split('-')[0] || 'en';
```
luwes, line 14:
```suggestion
const getBrowserLanguage = () => globalThis.navigator?.language.split('-')[0] || 'en';
```

luwes (approved):
LGTM 👍

src/js/utils/translation-helper.ts
davidvanleeuwen: @luwes Maybe I'm not completely understanding this, because wouldn't this cause a race condition? 

Screenshot 2025-03-25 at 17 56 46

see: https://codepen.io/davidvanleeuwen/pen/raNrzqj

And most linters would turn around the order, so this is quite a frustrating setup now.

ps: [chrome allows you to change locale](https://www.aleksandrhovhannisyan.com/blog/changing-locale-in-chrome-with-dev-tools/)

Issue #1069, opened 2025-02-05
Bug: Firefox PiP button activates when clicked through on settings menu


daniemo2:
On Firefox, this can be reproduced on the 'Toggling menu-item children' of the settings menu example: https://www.media-chrome.org/docs/en/examples/settings-menu

If I open the settings menu, and click somewhere that is in the same position as the PiP button, the video will move to PiP and the menu click will be ignored, despite (I believe) the menu being above the button.

Strangely, it seems that if the menu has a solid opacity background, it avoids this behaviour. However, a slightly transparent background is preferred.

ronalduQualabs:
@daniemo2 I could reproduce it, thanks for the information about the background I'm already looking for a solution.

ronalduQualabs:
@daniemo2 

Checking this error I discovered that it is more complicated than it seems. Firefox has its own [picture in picture system](https://support.mozilla.org/en-US/kb/about-picture-picture-firefox) in which from its interface they add that floating button that currently intervenes with our clicks in the media-chrome interface.

Besides that our picture in picture implementation does not work because firefox does not support the [picture-in-picture api](https://developer.mozilla.org/en-US/docs/Web/API/Picture-in-Picture_API). 

I'm still investigating ways to make the button not interrupt menu clicks without the need to change the color to solid since in custom players the error could return.

daniemo2:
Thanks for continuing to investigate @ronalduQualabs! I noticed that youtube doesn't suffer from the same issue, despite having transparent menus. I know they're not using Media Chrome but perhaps you can see if there's anything they've done to overcome the issue - although I couldn't see anything obvious.

ronalduQualabs:
@daniemo2 That's exactly what I trying to figure out how Youtube did that. I'll let you know the updates

ronalduQualabs:
After a research we discover that[ Firefox uses their own PiP system](https://support.mozilla.org/en-US/kb/about-picture-picture-firefox) adding a toggle overlaying videos and other elements if they dont have background color to 100% and opacity to 1 this make our menus unclickeable if they are under this toggle.

We resolve this issue by adding a css class for background solid color that only applies into firefox and we created this [ticket](https://bugzilla.mozilla.org/show_bug.cgi?id=1947825) for get a better solution from mozilla them since I also discover Youtube had the same problem but they created an exception that prevent this UX bug for elements with transparencies from 0.7 or higher for them identifying its **youtube** by the domain.
 
For now this can be partially solved.

PR #1074, opened 2025-02-12
Fix/firefox pip button activates when clicked through on settings menu

ronalduQualabs:
Fix to this [issue ](https://github.com/muxinc/media-chrome/issues/1069) After a research we discover that Firefox uses their own PiP system adding a toggle overlaying videos which identify if the element hovering it has opacity 100% or not and if it doesnt the click event will trigger the toggle instead out menu.

For now this can be partially solved adding a css class for background that only applies into firefox and we created this [ticket](https://bugzilla.mozilla.org/show_bug.cgi?id=1947825) for get a better solution for them since we also discover Youtube had the same problem but tey created an exception for transparencies from 0.7 or higher for them identifying youtube by the domain. 

Screenshot 2025-02-12 at 4 26 29 PM

### DIFF
--- a/src/js/menu/media-chrome-menu.ts
+++ b/src/js/menu/media-chrome-menu.ts
@@ -103,6 +103,19 @@ template.innerHTML = /*html*/ `
       position: relative;
       bottom: var(--_menu-bottom);
       box-sizing: border-box;
+    } 
+
+    @-moz-document url-prefix() {
+      :host::after{
+        content: "";
+        position: absolute;
+        top: 0; left: 0; right: 0; bottom: 0;
+        background: var(--media-settings-menu-background,
+        var(--media-menu-background,
+        var(--media-control-background,
+        var(--media-secondary-color, rgb(20 20 30)))));
+        z-index: -1;
+      }
     }
 
     :host([hidden]) {
@@ -463,7 +476,10 @@ class MediaChromeMenu extends globalThis.HTMLElement {
   }
 
   formatMenuItemText(text: string, data?: any) {
-    return (this.constructor as typeof MediaChromeMenu).formatMenuItemText(text, data);
+    return (this.constructor as typeof MediaChromeMenu).formatMenuItemText(
+      text,
+      data
+    );
   }
 
   get anchor() {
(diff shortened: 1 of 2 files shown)


ronalduQualabs:
@cjpillsbury Hi, could you help me to check this ?

src/js/menu/media-chrome-menu.ts
```diff
@@ -103,6 +103,12 @@ template.innerHTML = /*html*/ `
       position: relative;
       bottom: var(--_menu-bottom);
       box-sizing: border-box;
+    } 
+
+    @-moz-document url-prefix() {
+      :host{
+        background: rgb(20 20 30);
+      }
```
luwes, line 111: we still want to honor the CSS vars if they are set `var(--media-menu-background, var(--media-control-background, var(--media-secondary-color` I think. 

so maybe something like
https://codepen.io/luwes-the-looper/pen/NPWNmbY?editors=1100

```css
div {  
  --_menu-bg-color: rgb(20 20 30 / .8);
  
  background: var(--media-menu-background, var(--media-control-background, var(--media-secondary-color, var(--_bg-color))));
}

@-moz-document url-prefix() {
  div {
    --_menu-bg-color: rgb(20 20 30);
  }
}
```

src/js/menu/media-settings-menu.ts
```diff
@@ -16,6 +15,12 @@ template.innerHTML = MediaChromeMenu.template.innerHTML + /*html*/`
       overflow: hidden;
     }
 
+      @-moz-document url-prefix() {
+      :host{
+        background: rgb(20 20 30);
+      }
+    }
```
luwes, line 22: here as well.
some indentation fixes are needed in this file it seems.

ronalduQualabs: done

src/js/menu/media-settings-menu.ts
```diff
@@ -7,15 +7,22 @@ const template: HTMLTemplateElement = document.createElement('template');
...
   <style>
     :host {
+      --_menu-bg: rgb(20 20 30 / .8);
       background: var(--media-settings-menu-background,
-        var(--media-menu-background,
-        var(--media-control-background,
-        var(--media-secondary-color, rgb(20 20 30 / .8)))));
+          var(--media-menu-background,
+            var(--media-control-background,
+              var(--media-secondary-color, rgb(20 20 30 / .8)))));
```
luwes, line 14: missing the CSS var `--_menu-bg` here

ronalduQualabs: done

luwes (approved):
LGTM 👍

PR #1075, opened 2025-02-13
show Auto(mediaHeight) in rendition menu

### LINKED ISSUE #1052: Feature Request: Show 'Auto (mediaHeight)'  in Rendition Menu
If the rendition menu is a sub-menu inside of the settings menu, and 'Auto' is currently selected, it will have a description that shows the current media height playing, e.g. screenshot:

Image

However, if the rendition menu is standalone, it won't show this description.

Image

Would it be possible to implement this? 

I can almost piece it together with the `formatMenuItemText` function, however due to the rendering, when I switch between Auto to the same quality, the menu doesn't re-render as the `mediaheight` attribute hasn't changed. E.g. I have Auto selected and it's currently playing 1080p. If I select the 1080p menu item, it won't run `formatMenuItemText` to remove the '(1080p)' from auto.

Pintorado:
Resolves #1052

## Changes
I added conditions to check if the `Auto` option is selected when the menu is rendered and when it detects a change. If `Auto` is selected, the `mediaHeight` will be shown in the option.
## Example video
https://github.com/user-attachments/assets/2e8513f1-8c16-420d-a60b-94722f79999e

### DIFF
--- a/src/js/menu/media-rendition-menu.ts
+++ b/src/js/menu/media-rendition-menu.ts
@@ -17,7 +17,7 @@ import { Rendition } from '../media-store/state-mediator.js';
 
 /**
  * @extends {MediaChromeMenu}
- * 
+ *
  * @slot - Default slotted elements.
  * @slot header - An element shown at the top of the menu.
  * @slot checked-indicator - An icon element indicating a checked menu-item.
@@ -153,9 +153,13 @@ class MediaRenditionMenu extends MediaChromeMenu {
       this.defaultSlot.append(item);
     }
 
+    const text = isAuto
+      ? this.formatMenuItemText(`Auto (${this.mediaHeight}p)`)
+      : this.formatMenuItemText('Auto');
+
     const item = createMenuItem({
       type: 'radio',
-      text: this.formatMenuItemText('Auto'),
+      text,
       value: 'auto',
       checked: isAuto,
     });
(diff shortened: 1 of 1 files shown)


ronalduQualabs (approved):

src/js/menu/media-rendition-menu.ts
```diff
@@ -168,6 +172,10 @@ class MediaRenditionMenu extends MediaChromeMenu {
   }
 
   #onChange(): void {
+    const autoOption = this.defaultSlot.querySelector('[value="auto"]');
+    const isAuto = autoOption.part.contains('checked');
+    autoOption.textContent = isAuto ? `Auto (${this.mediaHeight}p)` : 'Auto';
```
luwes, line 177: This works but it would be nice to keep the declarative "render" method for updating all state in the UI. Under this line `this.value = newValue ?? 'auto';` if you add a new line under it with `this.#render();` you get the same final result though all menu items will be rendered again.

My original intention was to optimize the render method later doing a simple diff of the menu items and updating only where needed. If you like you can try to add that.

High level an algorithm that compares the length of the menu items. Add or remove if needed.
I wouldn't bother with swapping and order, that's less likely an issue for our menus I think.
Once the length is correct, just compare the data with the DOM properties. If they don't match, set them (overwrite).

luwes (changes requested):
it would be nice to keep the declarative render approach I think.
easy fix with adding `this.#render();` after the .value is set instead of the mutative logic.

option to optimize the "diff" algorithm but fine to do in a follow up.

luwes (approved):
LGTM 👍

PR #1082, opened 2025-02-27
fix: dialog accessibility hidden, page load bug

luwes:
related https://github.com/muxinc/elements/pull/1075

I tried using `@starting-style {}` but couldn't get that to work so let's do it old school.


heff (approved):
Looks ok to me

Issue #1083, opened 2025-02-27
4.6.0+ no longer has typings for React


Jiralite:
As of 4.6.0, it seems there are no more React typings.

To reproduce, install 4.6.0 and import anything from `media-chrone/react`.

A quick look at the commits makes #1068 the likely culprit.

Downgrading to 4.5.0 continues to work.

ronalduQualabs:
Hi @Jiralite, I just checked and ran into the same error you did at first. Restarting the editor or the TypeScript server fixed it for me. Can you try that and let me know if it works for you?

Jiralite:
This is not an IDE issue.

luwes:
Ah yes, it's complaining at the import on my end, tough the types on the components do work for me.

<img width="996" alt="Image" src="https://github.com/user-attachments/assets/67fc7f34-bce2-4e6b-b92e-6588910c1f2a" />

wesbos:
confirming this was fixed in `4.7.0` - this can be closed

PR #1085, opened 2025-02-28
fix: 4.6.0+ no longer has typings for React

ronalduQualabs:
Solves [This Issue](https://github.com/muxinc/media-chrome/issues/1083)

- Add ".d.ts" definitions for react components and menu react components


luwes (approved):
nice, that works too!

PR #1088, opened 2025-03-04
feat: persist mute state

### LINKED ISSUE #1072: Mute Setting not saving on refresh
When I muted the video with 100% volume this volume is storage in local storage then when refresh the page the muted setting its not enabled and audio its again to 100% percent.

Ideally the muted state should be storage as volume level in order the audio not re enable at refresh.

Pintorado:
Closes #1072
## Changes
- `MEDIA_UNMUTE_REQUEST`: When unmuting, if the preferred volume is too low (<= 0.25), set the volume to 0.25. Otherwise, restore the preferred volume.
- `MEDIA_VOLUME_REQUEST`: When volume is 0 and are unmuted, automatically mute.
- `mediaMuted` set behavior: When muting, set `media-chrome-pref-muted` to `true` in localStorage and when unmuting set it to `false`.
- `stateOwnersUpdateHandlers` behavior: calls `mediaMuted`'s set with the `mutedPref` value.

I replicated the same user experience as yt when muting and unmuting (restoring the previously selected volume), with the exception that if the preferred volume is too low (<=0.25), it is set to 0.25 to ensure the audio is clearly perceptible. I'm open to any other suggestions for doing this :)

### DIFF
--- a/src/js/media-store/request-map.ts
+++ b/src/js/media-store/request-map.ts
@@ -15,7 +15,7 @@ import { StateMediator, StateOwners } from './state-mediator.js';
 import { MediaState } from './media-store.js';
 
 export type MediaUIEventsType =
-typeof MediaUIEvents[keyof typeof MediaUIEvents];
+  typeof MediaUIEvents[keyof typeof MediaUIEvents];
 export type MediaRequestTypes = Exclude<
   MediaUIEventsType,
   | 'registermediastatereceiver'
--- a/src/js/media-store/request-map.ts
+++ b/src/js/media-store/request-map.ts
@@ -161,15 +161,13 @@ export const requestMap: RequestMap = {
   [MediaUIEvents.MEDIA_UNMUTE_REQUEST](stateMediator, stateOwners) {
     const key = 'mediaMuted';
     const value = false;
-    const preferredVolume =
+    const volumePref =
       +globalThis.localStorage.getItem('media-chrome-pref-volume') || +'0.25';
     // If we've unmuted but the current volume is 0, restore the preferred volume or set it to some low volume
-    if (!stateMediator.mediaVolume.get(stateOwners)) {
-      stateMediator.mediaVolume.set(
-        preferredVolume > 0.25 ? preferredVolume : 0.25,
-        stateOwners
-      );
-    }
+    stateMediator.mediaVolume.set(
+      volumePref > 0.25 ? volumePref : 0.25,
+      stateOwners
+    );
     stateMediator[key].set(value, stateOwners);
   },
   [MediaUIEvents.MEDIA_VOLUME_REQUEST](stateMediator, stateOwners, { detail }) {
--- a/src/js/media-store/request-map.ts
+++ b/src/js/media-store/request-map.ts
@@ -178,10 +178,6 @@ export const requestMap: RequestMap = {
     if (value && stateMediator.mediaMuted.get(stateOwners)) {
       stateMediator.mediaMuted.set(false, stateOwners);
     }
-    // If we've adjusted the volume to 0 and are unmuted, automatically mute.
-    if(value === '0' && !stateMediator.mediaMuted.get(stateOwners)){
-      stateMediator.mediaMuted.set(true, stateOwners);
-    }
     stateMediator[key].set(value, stateOwners);
   },
   [MediaUIEvents.MEDIA_SEEK_REQUEST](stateMediator, stateOwners, { detail }) {
--- a/src/js/media-store/request-map.ts
+++ b/src/js/media-store/request-map.ts
@@ -161,13 +161,10 @@ export const requestMap: RequestMap = {
   [MediaUIEvents.MEDIA_UNMUTE_REQUEST](stateMediator, stateOwners) {
     const key = 'mediaMuted';
     const value = false;
-    const volumePref =
-      +globalThis.localStorage.getItem('media-chrome-pref-volume');
-    // If we've unmuted but the current volume is 0, restore the preferred volume or set it to some low volume
-    stateMediator.mediaVolume.set(
-      volumePref > 0.25 ? volumePref : 0.25,
-      stateOwners
-    );
+    // If we've unmuted but our volume is currently 0, automatically set it to some low volume
+    if (!stateMediator.mediaVolume.get(stateOwners)) {
+      stateMediator.mediaVolume.set(0.25, stateOwners);
+    }
     stateMediator[key].set(value, stateOwners);
   },
   [MediaUIEvents.MEDIA_VOLUME_REQUEST](stateMediator, stateOwners, { detail }) {
--- a/src/js/media-store/request-map.ts
+++ b/src/js/media-store/request-map.ts
@@ -161,10 +161,6 @@ export const requestMap: RequestMap = {
   [MediaUIEvents.MEDIA_UNMUTE_REQUEST](stateMediator, stateOwners) {
     const key = 'mediaMuted';
     const value = false;
-    // If we've unmuted but our volume is currently 0, automatically set it to some low volume
-    if (!stateMediator.mediaVolume.get(stateOwners)) {
-      stateMediator.mediaVolume.set(0.25, stateOwners);
-    }
     stateMediator[key].set(value, stateOwners);
   },
   [MediaUIEvents.MEDIA_VOLUME_REQUEST](stateMediator, stateOwners, { detail }) {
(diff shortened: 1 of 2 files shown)


src/js/media-store/request-map.ts
```diff
@@ -161,10 +161,13 @@ export const requestMap: RequestMap = {
   [MediaUIEvents.MEDIA_UNMUTE_REQUEST](stateMediator, stateOwners) {
     const key = 'mediaMuted';
     const value = false;
-    // If we've unmuted but our volume is currently 0, automatically set it to some low volume
-    if (!stateMediator.mediaVolume.get(stateOwners)) {
-      stateMediator.mediaVolume.set(0.25, stateOwners);
-    }
+    const volumePref =
+      +globalThis.localStorage.getItem('media-chrome-pref-volume') || +'0.25';
```
luwes, line 165:
```suggestion
      +globalThis.localStorage.getItem('media-chrome-pref-volume');
```

that fallback doesn't seem needed because below you're already doing a check
`NaN > 0.25 = false`

src/js/media-store/state-mediator.ts
```diff
@@ -389,6 +389,26 @@ export const stateMediator: StateMediator = {
     set(value, stateOwners) {
       const { media } = stateOwners;
       if (!media) return;
+
+      try {
+        if (value) {
+          globalThis.localStorage.setItem('is-muted', 'true');
```
luwes, line 395: let's keep the same naming `media-chrome-pref-muted`

src/js/media-store/state-mediator.ts
```diff
@@ -389,6 +389,26 @@ export const stateMediator: StateMediator = {
...
+          globalThis.localStorage.setItem(
+            'media-chrome-pref-volume',
+            media.volume.toString()
+          );
+          media.volume = 0;
```
luwes, line 400: I wouldn't conflate muted and volume. what the reason behind this?

src/js/media-store/state-mediator.ts
```diff
@@ -389,6 +389,26 @@ export const stateMediator: StateMediator = {
...
+        } else {
+          globalThis.localStorage.setItem('is-muted', 'false');
+          const volumePref = globalThis.localStorage.getItem(
+            'media-chrome-pref-volume'
+          );
+          media.volume = +volumePref;
```
luwes, line 406: remove if muted and volume is kept separate

src/js/media-store/state-mediator.ts
```diff
@@ -428,14 +448,25 @@ export const stateMediator: StateMediator = {
...
           const volumePref = globalThis.localStorage.getItem(
             'media-chrome-pref-volume'
           );
+          const isMuted =
+            globalThis.localStorage.getItem('is-muted') === 'true';
+
+          media.muted = isMuted;
+          if (!isMuted && volumePref !== null) {
+            media.volume = +volumePref;
+          }
```
luwes, line 463: I think this shouldn't be here, the muted localStorage logic should be part of the the `mediaMuted` property https://github.com/muxinc/media-chrome/pull/1088/files#diff-333677ad6c1f12477baa3069c5713d46dfcfe6d53826b6743860ca2d7eecb49cR383

it'll require a `stateOwnersUpdateHandlers` as well.
basically you need to clone how mediaVolume does it but then for mediaMuted. no need to mix the 2 I think.

src/js/media-store/state-mediator.ts
Pintorado: My bad, this slipped my attention. I think this was necessary when `MEDIA_UNMUTE_REQUEST` checked if the media volume was 0 instead of `media-chrome-pref-volume`. Not necessary now so I'll remove it

src/js/media-store/request-map.ts
```diff
@@ -175,6 +178,10 @@ export const requestMap: RequestMap = {
     if (value && stateMediator.mediaMuted.get(stateOwners)) {
       stateMediator.mediaMuted.set(false, stateOwners);
     }
+    // If we've adjusted the volume to 0 and are unmuted, automatically mute.
```
luwes, line 181: I don't think this is needed, part of separation of volume and muted.

src/js/media-store/state-mediator.ts
```diff
@@ -428,14 +438,22 @@ export const stateMediator: StateMediator = {
...
           const volumePref = globalThis.localStorage.getItem(
             'media-chrome-pref-volume'
           );
+          const mutedPref =
+            globalThis.localStorage.getItem('media-chrome-pref-muted') ===
+            'true';
+
           if (volumePref == null) return;
           stateMediator.mediaVolume.set(+volumePref, stateOwners);
+          stateMediator.mediaMuted.set(mutedPref, stateOwners);
```
luwes, line 453: this logic should go in a new `stateOwnersUpdateHandlers` property of the `mediaMuted` key above I believe. I think you can almost clone the logic of 
 the mediaVolume property value but replace "volume" with "muted" if that makes sense. of course volume is a number and muted is a boolean...

src/js/media-store/request-map.ts
Pintorado: It worked fine without this, but `media-chrome-pref-muted` wasn't updating to `true` when muting through the volume range, that's why I added this to force it just in case it caused any problem. Should I remove it?

src/js/media-store/state-mediator.ts
Pintorado: Done! I'm not sure if I should also add `noMutedPref` in the stateOwners options as we have `noVolumePref`. What do you think?

src/js/media-store/request-map.ts
luwes: remove it because muting through the volume range shouldn't be possible. 
you can set volume to 0% but that is different than the muted state.

Pintorado: Got it!

luwes (approved):
LGTM! 👍

src/js/media-store/state-mediator.ts
luwes: oh yes, we can follow up with that.

Pintorado:
> some tests are failing, could you have a look at that @Pintorado ?

Yes I'm already working on it!

src/js/media-store/request-map.ts
```diff
@@ -15,7 +15,7 @@ import { StateMediator, StateOwners } from './state-mediator.js';
 import { MediaState } from './media-store.js';
 
 export type MediaUIEventsType =
-  typeof MediaUIEvents[keyof typeof MediaUIEvents];
+typeof MediaUIEvents[keyof typeof MediaUIEvents];
```
luwes, line 18:
```suggestion
  typeof MediaUIEvents[keyof typeof MediaUIEvents];
```

src/js/media-store/request-map.ts
```diff
@@ -161,10 +161,13 @@ export const requestMap: RequestMap = {
...
-    if (!stateMediator.mediaVolume.get(stateOwners)) {
-      stateMediator.mediaVolume.set(0.25, stateOwners);
-    }
+    const volumePref =
+      +globalThis.localStorage.getItem('media-chrome-pref-volume');
+    // If we've unmuted but the current volume is 0, restore the preferred volume or set it to some low volume
+    stateMediator.mediaVolume.set(
+      volumePref > 0.25 ? volumePref : 0.25,
+      stateOwners
+    );
```
luwes, line 170:
```suggestion
    // If we've unmuted but our volume is currently 0, automatically set it to some low volume
    if (!stateMediator.mediaVolume.get(stateOwners)) {
      stateMediator.mediaVolume.set(0.25, stateOwners);
    }
```

this actually introduces a regression because we're making volume 0.25 with every unmute request. even if volume was 1

src/js/media-store/state-mediator.ts
```diff
@@ -389,9 +390,37 @@ export const stateMediator: StateMediator = {
...
+
       media.muted = value;
     },
     mediaEvents: ['volumechange'],
+    stateOwnersUpdateHandlers: [
+      (handler, stateOwners) => {
+        const {
+          options: { noMutedPref },
+        } = stateOwners;
+        if (noMutedPref) return;
```
luwes, line 411:
```suggestion
        const { media } = stateOwners;
        // The muted enabled attribute should still override the preference.
        if (!media || media.muted || noMutedPref) return;
```

src/js/media-store/request-map.ts
Pintorado: Actually this didn't set it to 0.25 if volume was 1. This only happened if `volumePref` was lower than 0.25. I made this UX decision so that the audio is clearly perceptible when unmuting and having a really low `volumePref`. But without this it works fine anyway

PR #1090, opened 2025-03-12
tools: replace yarn v1 with npm

luwes:


decepulis (commented):
The code you've written looks good, but I can't run it locally.
`npm run dev` errors because `npm run build:types -w` is looking for a workspace after `-w`. Did you mean to pass the `-w` flag to `build:types` with `--`? I wonder what other flags like that might've been missed

decepulis (approved):
Nice, this is looking really good

PR #1098, opened 2025-03-19
fix: timerange thumb and tooltip drift mobile

### LINKED ISSUE #883: timerange thumb and tooltip drift mobile
the time tooltip and thumb of the time range seem to drift apart if scrubbing on mobile, especially if you hold on and move vertically.

image_from_ios

Pintorado:
Fixes #883 
## Changes
Modified the pointer position calculation in `getPointerRatio`

### DIFF
--- a/src/js/media-chrome-range.ts
+++ b/src/js/media-chrome-range.ts
@@ -2,6 +2,7 @@ import { MediaStateReceiverAttributes } from './constants.js';
 import { globalThis, document } from './utils/server-safe-globals.js';
 import {
   getOrInsertCSSRule,
+  getPointProgressOnLine,
 } from './utils/element-utils.js';
 import { observeResize, unobserveResize } from './utils/resize-observer.js';
 
@@ -516,9 +517,13 @@ class MediaChromeRange extends globalThis.HTMLElement {
   getPointerRatio(evt) {
     const startRect = this.#startpoint.getBoundingClientRect();
     const endRect = this.#endpoint.getBoundingClientRect();
-    const pointerRatio =
-      (evt.clientX - startRect.left) / (endRect.left - startRect.left);
-    return Math.max(0, Math.min(1, pointerRatio));
+    
+    return getPointProgressOnLine(
+      evt.clientX,
+      evt.clientY,
+      { x: startRect.left, y: startRect.bottom },
+      { x: endRect.left, y: endRect.top }
+    );
   }
 
   get dragging() {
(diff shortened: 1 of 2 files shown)


src/js/utils/element-utils.ts
```diff
@@ -173,14 +173,20 @@ export function getPointProgressOnLine(
...
+  if (isVertical) {
+    // For vertical lines, calculate progress from bottom to top
+    const totalHeight = p2.y - p1.y;
+    const progress = (y - p1.y) / totalHeight;
+    return Math.max(0, Math.min(1, progress));
+  } else {
+    // For horizontal lines, calculate progress from left to right
+    const totalWidth = p2.x - p1.x;
+    const progress = (x - p1.x) / totalWidth;
+    return Math.max(0, Math.min(1, progress));
```
luwes, line 188: this is a good try but it excludes the very edge case where a slider would be not vertical or horizontal. I asked AI to see what was wrong with the original function and it came back with this. It seems on mobile to fix the issue as well. Could you verify it works with the vertical slider case?

```js
/**
 * Get progress ratio of a point on a line segment.
 * @param x - The x coordinate of the point.
 * @param y - The y coordinate of the point.
 * @param p1 - The first point of the line segment.
 * @param p2 - The second point of the line segment.
 */
export function getPointProgressOnLine(
  x: number,
  y: number,
  p1: Point,
  p2: Point
): number {
  const dx = p2.x - p1.x;
  const dy = p2.y - p1.y;
  const lengthSquared = dx * dx + dy * dy;

  if (lengthSquared === 0) return 0; // Avoid division by zero if p1 === p2

  const projection = ((x - p1.x) * dx + (y - p1.y) * dy) / lengthSquared;

  return Math.max(0, Math.min(1, projection)); // Clamp between 0 and 1
}
```

src/js/media-chrome-range.ts
```diff
@@ -515,13 +515,15 @@ class MediaChromeRange extends globalThis.HTMLElement {
...
+    const startRect = this.#startpoint.getBoundingClientRect();
+    const endRect = this.#endpoint.getBoundingClientRect();
+    
+    return getPointProgressOnLine(
       evt.clientX,
       evt.clientY,
-      this.#startpoint.getBoundingClientRect(),
-      this.#endpoint.getBoundingClientRect()
+      { x: startRect.left, y: startRect.bottom },
+      { x: endRect.left, y: endRect.top }
```
luwes, line 525: this change shouldn't be needed. please revert.

src/js/utils/element-utils.ts
Pintorado: Yes this works, I'll update, thank you

luwes (approved):
👍 thank you!

Issue #756, opened 2023-10-31
Hide mouse cursor (and controls) on video element after user inactivity


luwes:
feature requested by multiple users

- [x] hide mouse cursor on video element when controls fade out
- [ ] add option that even when mouse cursor is on the controls (button, timerange etc) the controls and mouse cursor will still fade out after a delay.
- [ ] be smart on play button click and hiding controls / cursor like explained by @cjpillsbury below

kentcdodds:
Can I request that the cursor be hidden after inactivity even if it's over video controls? It bugs the heck out of me when someone's sharing their screen and they move their mouse over the controls and the controls stay up through the whole video. Pet peeve I guess.

luwes:
@kentcdodds thanks for the feedback! I'm not sure we can make that the default. 
generally the controls stay active when the mouse is over the controls. I'll bring it up as an option w/ the team though.

have you seen this behavior with another player?

kentcdodds:
I've definitely seen players do both, but can't think of a specific example

cjpillsbury:
Appreciate the feedback, @kentcdodds and I've definitely seen that as well. We'd want to be cautious about adding it as default, minimally for accessibility considerations (not all folks are as quick on the draw with interactivity), but I like the idea of making it optional. Another thing we should probably consider in relationship to that is having our delayed hide behavior kick off if a user takes an action: e.g. if the user clicks the play button but doesn't then move the cursor, we can assume (optionally, by default, or as the de facto behavior) that they'd like the controls to eventually automatically hide. We should probably first do an ad hoc survey of other players in the wild to see how they behave around these things (iirc youtube will do this, but maybe only for play/pause? would have to recheck).

kentcdodds:
That sounds awesome to me 👍

PR #1101, opened 2025-03-24
feat:hide-cursor-controls-on-inactivity

ronalduQualabs:
This PR resolves [issue #756](https://github.com/muxinc/media-chrome/issues/756), addressing the case where controls and the cursor should hide even when hovering over the controls.

### Changes Introduced
- Added the alwaysHideControls attribute to the media controller.
- When this attribute is present, controls and the cursor will hide after user inactivity, even if the user is hovering over the controls only if video is playing and not casting.

##### With alwaysHideControls:
- After clicking play, if the user does not move the cursor, both the controls and cursor will hide—even when hovering over the controls.
#### Without alwaysHideControls (default behavior):
- Controls and cursor will only hide if the user is not hovering over the controls is playing and is not casting

### DIFF
--- a/src/js/extras/media-clip-selector/index.ts
+++ b/src/js/extras/media-clip-selector/index.ts
@@ -35,7 +35,7 @@ template.innerHTML = `
     }
 
     #startHandle, #endHandle {
-      cursor: pointer;
+      cursor: var(--custom-cursor, pointer);
       height: 80%;
       width: ${HANDLE_W}px;
       border-radius: 4px;
--- a/src/js/extras/media-clip-selector/index.ts
+++ b/src/js/extras/media-clip-selector/index.ts
@@ -35,7 +35,7 @@ template.innerHTML = `
     }
 
     #startHandle, #endHandle {
-      cursor: var(--custom-cursor, pointer);
+      cursor: var(--media-cursor, pointer);
       height: 80%;
       width: ${HANDLE_W}px;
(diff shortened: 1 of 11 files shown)


src/js/media-container.ts
```diff
@@ -685,6 +700,14 @@ class MediaContainer extends globalThis.HTMLElement {
     setBooleanAttr(this, Attributes.NO_AUTOHIDE, value);
   }
 
+  get alwaysHideControls(): boolean | undefined {
```
luwes, line 703: let's discuss this in open source office hours.

we have the attribute `autohide` which is a number in seconds and -1 turns it off.

there is also the `noautohide` attribute which can be added to controls to not get automatically hidden.

src/js/extras/media-clip-selector/index.ts
```diff
@@ -35,7 +35,7 @@ template.innerHTML = `
     }
 
     #startHandle, #endHandle {
-      cursor: pointer;
+      cursor: var(--custom-cursor, pointer);
```
luwes, line 38: most public Media Chrome CSS vars are prefixed with `media-`
so for this one it should be `--media-cursor`.

ronalduQualabs: changed

luwes (approved):
awesome! thanks

Issue #1097, opened 2025-03-19
Update Docs to include usage of languages API in media-chrome


ronalduQualabs:
Update documentation to explain how to use the languages in media-chrome for labels and tooltips according to [PR #1071](https://github.com/muxinc/media-chrome/pull/1071).

PR #1102, opened 2025-03-24
docs: adding-language-support

ronalduQualabs:
Includes the documentations update for include multiple languages into media-chrome [Issue](https://github.com/muxinc/media-chrome/issues/1097)

### DIFF
--- a/docs/src/pages/docs/en/audio-player.md
+++ b/docs/src/pages/docs/en/audio-player.md
@@ -24,7 +24,7 @@ Since audio players don’t use the same visual layout as video players do, we c
 </media-controller>
 
 ```html
-<script type="module" src="https://cdn.jsdelivr.net/npm/media-chrome@4/+esm"></script>
+<script type="module" src="https://cdn.jsdelivr.net/npm/media-chrome@4.8.0/+esm"></script>
 
 <media-controller audio>
   <audio
--- a/docs/src/pages/docs/en/audio-player.md
+++ b/docs/src/pages/docs/en/audio-player.md
@@ -24,7 +24,7 @@ Since audio players don’t use the same visual layout as video players do, we c
 </media-controller>
 
 ```html
-<script type="module" src="https://cdn.jsdelivr.net/npm/media-chrome@4.8.0/+esm"></script>
+<script type="module" src="https://cdn.jsdelivr.net/npm/media-chrome@4/+esm"></script>
 
 <media-controller audio>
   <audio
--- a/docs/src/pages/docs/en/get-started.mdx
+++ b/docs/src/pages/docs/en/get-started.mdx
@@ -54,7 +54,7 @@ Load the module in the `<head>` of your HTML page. Note the `type="module"`, tha
 
 <CopyToClipboard>
 ```html
-<script type="module" src="https://cdn.jsdelivr.net/npm/media-chrome@4.8.0/+esm"></script>
+<script type="module" src="https://cdn.jsdelivr.net/npm/media-chrome@4/+esm"></script>
 ```
 </CopyToClipboard>
 
--- a/docs/src/pages/docs/en/get-started.mdx
+++ b/docs/src/pages/docs/en/get-started.mdx
@@ -54,7 +54,7 @@ Load the module in the `<head>` of your HTML page. Note the `type="module"`, tha
 
 <CopyToClipboard>
 ```html
-<script type="module" src="https://cdn.jsdelivr.net/npm/media-chrome@4/+esm"></script>
+<script type="module" src="https://cdn.jsdelivr.net/npm/media-chrome@4.8.0/+esm"></script>
 ```
 </CopyToClipboard>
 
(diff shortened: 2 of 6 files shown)


docs/src/pages/docs/en/audio-player.md
```diff
@@ -24,7 +24,7 @@ Since audio players don’t use the same visual layout as video players do, we c
 </media-controller>
 
 ```html
-<script type="module" src="https://cdn.jsdelivr.net/npm/media-chrome@4/+esm"></script>
+<script type="module" src="https://cdn.jsdelivr.net/npm/media-chrome@4.8.0/+esm"></script>
```
luwes, line 27: these can src url changes can be reverted because https://cdn.jsdelivr.net/npm/media-chrome@4/+esm will return the latest v4. sometimes it takes a while before the jsdelivr cache is purged

docs/src/pages/docs/en/internationalization/adding-language-support.mdx
```diff
@@ -0,0 +1,56 @@
...
+- **French** (`fr`)
+- **Portuguese** (`pt`)
+
+## How to Include a Language
+
+### Using HTML
+To add support for a specific language in an HTML file, include the corresponding language module using a `<script>` tag:
+
+```html
+<script type="module" src="https://cdn.jsdelivr.net/npm/media-chrome@4.8.0/dist/lang/pt.js/+esm"></script>
```
luwes, line 23:
```suggestion
<script type="module" src="https://cdn.jsdelivr.net/npm/media-chrome@4.8.0/dist/lang/pt.js"></script>
```

we have to use the non /+esm endpoint for this to work.

also here we should mention that for this to work they have to use the non-bundled media-chrome dist
this one https://cdn.jsdelivr.net/npm/media-chrome@4
instead of https://cdn.jsdelivr.net/npm/media-chrome@4/+esm

this is needed because we need to set state in the `i18n.js` module which only works if it can be imported separately.

docs/src/pages/docs/en/internationalization/adding-language-support.mdx
```diff
@@ -0,0 +1,56 @@
...
+```
+
+Replace `pt.js` with the appropriate language code (`es.js`, `fr.js`, etc.) for other languages.
+
+### Using JavaScript
+If you are working with JavaScript, you can import the language module as follows:
+
+```js
+import "media-chrome/dist/lang/pt.js";
```
luwes, line 32:
```suggestion
import "media-chrome/lang/pt.js";
```

remove the dist part if possible. I noticed we might need to add one more exports block before the other one to make this work. this one

```
    "./lang/*.js": {
      "import": "./dist/lang/*.js",
      "require": "./dist/lang/*.js",
      "default": "./dist/lang/*.js"
    }
```

luwes (changes requested):
looking good but it needs a few changes to make everything work smoothly.

docs/src/pages/docs/en/get-started.mdx
```diff
@@ -54,7 +54,7 @@ Load the module in the `<head>` of your HTML page. Note the `type="module"`, tha
 
 <CopyToClipboard>
 ```html
-<script type="module" src="https://cdn.jsdelivr.net/npm/media-chrome@4/+esm"></script>
+<script type="module" src="https://cdn.jsdelivr.net/npm/media-chrome@4.8.0/+esm"></script>
```
luwes, line 57: revert please

docs/src/pages/docs/en/internationalization/adding-language-support.mdx
```diff
@@ -0,0 +1,57 @@
...
+### Using HTML
+To add support for a specific language in an HTML file, include the corresponding language module using a `<script>` tag:
+
+```html
+<script type="module" src="https://cdn.jsdelivr.net/npm/media-chrome@4.8.0/lang/pt.js"></script>
+```
+
+Replace `pt.js` with the appropriate language code (`es.js`, `fr.js`, etc.) for other languages.
+
+> **Note:** This only works with the non-bundled version of Media Chrome (`https://cdn.jsdelivr.net/npm/media-chrome@4`) and not the **`/+esm`** version. This is required because the `i18n.js` module must be imported separately to properly manage state.
```
luwes, line 27: 👍

package.json
```diff
@@ -86,7 +86,7 @@
       "require": "./dist/cjs/*.js",
       "default": "./dist/*.js"
     },
-    "./lang/*": {
+    "./dist/lang/*": {
```
luwes, line 89: does this work to resolve both `/lang/pt.js` and `/lang/pt` in a bundler setup? 

also just noticed we should order the `exports` keys from specific to non-specific because the first match gets chosen and stops. basically the lang related should go above the `./*` and `./*.js`

docs/src/pages/docs/en/get-started.mdx
luwes, line 57:
```suggestion
<script type="module" src="https://cdn.jsdelivr.net/npm/media-chrome@4/+esm"></script>
```

PR #1105, opened 2025-03-25
feat: set i18n through media-controller

davidvanleeuwen:
This adds ~~a singleton class~~ `let` to manage the language, and uses the browser language as fallback. If the browser language isn't present as language file, it'll just grab the `en`, like it did before.

The language for this can be set through an attribute on media-controller like in the example. So: 
```html
<media-controller language="fr"></media-controller>
```

Now this can be managed outside media-chrome through a cookie, interface element or whatsoever.

### DIFF
--- a/src/js/media-controller.ts
+++ b/src/js/media-controller.ts
@@ -33,6 +33,7 @@ import {
 } from './utils/element-utils.js';
 import { createMediaStore, MediaStore } from './media-store/media-store.js';
 import { CustomElement } from './utils/CustomElement.js';
+import { i18n } from './utils/i18n.js';
 
 const ButtonPressedKeys = [
   'ArrowLeft',
@@ -61,6 +62,7 @@ export const Attributes = {
   NO_DEFAULT_STORE: 'nodefaultstore',
   KEYBOARD_FORWARD_SEEK_OFFSET: 'keyboardforwardseekoffset',
   KEYBOARD_BACKWARD_SEEK_OFFSET: 'keyboardbackwardseekoffset',
+  LANGUAGE: 'language',
 };
 
 /**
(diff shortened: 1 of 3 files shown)


src/js/utils/i18n.ts
```diff
@@ -11,13 +11,39 @@ export const addTranslation = (
   translationsLanguages[langCode] = languageDictionary;
 };
 
-const getBrowserLanguage = () => globalThis.navigator?.language.split('-')[0] || 'en';
+export class I18n {
```
luwes, line 14: is there an advantage to using a singleton instead of a simple variable in the module?

an ESM module is only imported and executed once so using a simple variable to hold the state is the same I think with less code.

davidvanleeuwen: It's mostly for my own clarity, but I understand that this is not a pattern used within media-chrome, so I hold my opinions loosely 😄 So I'm happy to convert it to a simple variable.

luwes: yes, this feels more fitting. thank you. the bundle size is also starting to increase so we have try to squeeze every byte out of it 😄

luwes:
one more little change and I think we could merge this. 

I did a poll internally for the attribute naming. `lang` vs `language`.
most folks preferred the `lang` naming since it's the same than `<html lang="">`.

It can be set on each HTML element and doesn't cause side-effects as far as I know.
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang#inheritance

src/js/media-controller.ts
```diff
@@ -273,6 +276,14 @@ class MediaController extends MediaContainer {
     setBooleanAttr(this, Attributes.NO_DEFAULT_STORE, value);
   }
 
+  get language(): string | undefined {
```
luwes, line 279:
```suggestion
  get lang(): string | undefined {
```

src/js/media-controller.ts
```diff
@@ -273,6 +276,14 @@ class MediaController extends MediaContainer {
...
+    return getStringAttr(this, Attributes.LANGUAGE);
+  }
+
+  set language(value: string | undefined) {
```
luwes, line 283:
```suggestion
  set lang(value: string | undefined) {
```

src/js/media-controller.ts
davidvanleeuwen: Not the biggest fan of making function names obscure, I just like things to be more explicit. Anyway, those loose opinions.

PR #1108, opened 2025-03-27
fix: simplify exports, resolve tests. fix lang prop type issue

luwes:
we were seeing some strange behavior with resolving of paths of the `/lang/x.js` path.
did a quick spike on this, simplified our exports object and added some tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,12 +41,6 @@
       "require": "./dist/cjs/react/menu/index.js",
       "default": "./dist/react/menu/index.js"
     },
-    "./react/media-store": {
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -41,12 +41,6 @@
       "require": "./dist/cjs/react/menu/index.js",
       "default": "./dist/react/menu/index.js"
     },
-    "./react/media-store": {
-      "types": "./dist/react/media-store.d.ts",
-      "import": "./dist/react/media-store.js",
-      "require": "./dist/cjs/react/index.js",
```
luwes, line 47: also a bug here 🙈

PR #1122, opened 2025-04-11
Discourage external access to example JS

decepulis:
We'd like to encourage users to access media chrome through npm or a CDN like jsdeliver


luwes (approved):

luwes (approved):
👍

PR #1123, opened 2025-04-14
Add visual feedback in players with autoplay

### LINKED ISSUE #1078: player with autoplay not giving visual feedback
Thanks for creating media-chrome!

I have built a custom player which should autoplay videos on page-load. (This is not a permission issue.) 
Media-chrome currently loads metadata without UI feedback and starts playing as soon as metadata has loaded.
I would expect the play-button to change state right away and display the loading-indicator while metadata is loading. At the moment the user does not get feedback, that the player is loading data in the background.

If I call `.play()` imperatively on pageload I get `Uncaught (in promise) AbortError: The play() request was interrupted by a new load request.` error. I assume the way media-chrome is designed the `media-controller` needs to "discover" the video element first. So I ideally would listen for some kind of source-changed event and then call `.play()` to get visual feedback right away.

Pintorado:
Closes #1078 

### Current behavior
When using a player with autoplay enabled, the play button remains in a paused state until the metadata is loaded and playback begins, so there is no visible indication that the video is preparing to play which leads to a confusing user experience.
## Changes
To provide immediate visual feedback, this PR simulates playback and loading states until the actual play event fires. 
- Added `#isAutoplay = false` to track simulated autoplay state.
- In connectedCallback: If the media element has the autoplay attribute => Set `mediaPaused = false` and `mediaLoading = true` to simulate playback. Listen for the actual play event to end the simulation and update real state.
- Updated propagateMediaState helper: Added `isAutoplay = false` param and skipped setting `mediaPaused` when `isAutoplay` is true to avoid misleading state changes before playback starts (when simulating autoplay).

### DIFF
--- a/src/js/media-controller.ts
+++ b/src/js/media-controller.ts
@@ -100,6 +100,7 @@ class MediaController extends MediaContainer {
   mediaStateReceivers: HTMLElement[] = [];
   associatedElementSubscriptions: Map<HTMLElement, () => void> = new Map();
 
+  #isAutoplay = false;
   #hotKeys = new AttributeTokenList(this, Attributes.HOTKEYS);
   #fullscreenElement: HTMLElement;
   #mediaStore: MediaStore;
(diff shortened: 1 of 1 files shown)


luwes:
I'm not sure about this approach, it adds quite a bit of code. 

Are we setting mediapaused when the media element autoplay is enabled? seems like a bug.
Can this fix be done by the user adding `medialoading` manually to the media-loading-indicator?

Pintorado:
This approach simulates a playing state when autoplay is enabled and while the media is still loading, it sets `mediapaused` to false so that the loading indicator is visible in this state. I agree, I'm not entirely sure about this, I was testing and thinking about other approaches while this was still in draft.

Adding `medialoading` manually doesn't seem to solve this, as `mediapaused` needs to be false for the loader to be visible, and for some players like HLS and DASH this attribute is true until the media is fully loaded, so it's a bit tricky.

Issue #1128, opened 2025-04-25
Is there a way to translate the default settings menu items and labels?


luwes:
### Discussed in https://github.com/muxinc/media-chrome/discussions/1067

<div type='discussions-op-text'>

<sup>Originally posted by **akkie** February  3, 2025</sup>
Hi, I currently search for a way to translate the default values like "off" or "auto" in the settings menu. Most things can be replaced via slots, but I have not found a way to do it for these values.

![image](https://github.com/user-attachments/assets/8a39182c-49c8-45f7-9852-2ca53bce4d70)

Cheers,
Christian

</div>

PR #1130, opened 2025-04-29
Add i18n to settings menu defaults

ronalduQualabs:
As requested in [this issue](https://github.com/muxinc/media-chrome/issues/1128), the labels for “auto” and “off” have been added to the language files so they can now be translated if desired.


luwes (approved):
LGTM
some formatting issues, double quote instead of single quote

PR #1131, opened 2025-05-01
Check that language exists before splitting

wesbos:
This code assumes that if there is a navigator, that `navigator.language` exists. Ran into issues server rendering on Cloudflare since they don't have this property.

I've made a report on their end, but this should be checked so we can safely fallback to `en` 

https://github.com/cloudflare/workerd/issues/4067

### DIFF
--- a/src/js/utils/i18n.ts
+++ b/src/js/utils/i18n.ts
@@ -4,7 +4,7 @@ const translationsLanguages = {
   en: En,
 };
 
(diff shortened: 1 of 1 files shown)


davekiss (approved):

luwes (approved):
thanks, lgtm!

wesbos:
thanks everyone!

PR #1134, opened 2025-05-06
Add Chinese (Simplified and Traditional) translations

gnehs:
This pull request adds support for Chinese translations:

* `zh-TW`: Traditional Chinese (commonly used in Taiwan)
* `zh-CN`: Simplified Chinese (commonly used in Mainland China)

The translations include all key UI elements and error messages related to media playback.
Let me know if any revisions are needed.

PR #1136, opened 2025-05-07
Add support for language variants in i18n (e.g., en-AU fallback to en if doesn't exists)

ronalduQualabs:
This PR enhances the i18n system to support language variants (e.g., en-AU, fr-CA, zh-TW) with proper fallback logic. 

This comes as a solution for accepting this [PR](https://github.com/muxinc/media-chrome/pull/1134) which includes two variants of Chinese.


src/js/utils/i18n.ts
```diff
@@ -1,31 +1,37 @@
...
   );
 };
+
+export const t = (
+  key: TranslateKeys,
+  vars: Record<string, string | number> = {}
+): string =>
+  resolveTranslation(key).replace(/\{(\w+)\}/g, (_, v) =>
+    v in vars ? String(vars[v]) : `{${v}}`
+  );
```
luwes, line 37:
```suggestion
  );

```

always end with an empty newline. should be a text editor setting that is done when saving

luwes (approved):
LGTM 👍

src/js/utils/i18n.ts
```diff
@@ -1,31 +1,37 @@
...
+  translations[lang] = languageDictionary;
 };
 
-export const t = (
-  key: TranslateKeys,
-  variables: Record<string, string | number> = {}
-) => {
-  const result = translationsLanguages[currentLanguage]?.[key] || En[key];
+const resolveTranslation = (key: TranslateKeys): string => {
+  const [base] = currentLang.split('-');
```
luwes, line 21:
```suggestion
  const [base] = currentLang?.split('-');
```

luwes (changes requested):

PR #1140, opened 2025-05-13
fix: React rates array

### LINKED ISSUE #1120: MediaPlaybackRateMenu rates property should be a string in React instead of array of numbers or media-chrome will not use custom rates
If I use the rates property as it should be by type definitions, then rates will stay as a default value.

In code:

Image

How they appear (not expected):

Image

If instead I will use a string that will be treated by TS as a any type, then it works

In code:

Image

How they appear (expected):

Image

Media-Chrome version: 4.9.0

luwes:
fix #1120
closes #1139

### DIFF
--- a/scripts/react/build.js
+++ b/scripts/react/build.js
@@ -38,6 +38,13 @@ export const ${ReactComponentName} = createComponent({
   tagName: "${elementName}",
   elementClass: Modules.${ReactComponentName},
   react: React,
+  toAttributeValue: (propValue) => {
+    if (typeof propValue === 'boolean') return propValue ? '' : undefined;
+    if (typeof propValue === 'function') return undefined;
+    if (Array.isArray(propValue)) return propValue.join(' ');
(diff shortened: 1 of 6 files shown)


scripts/react/build.js
```diff
@@ -38,6 +38,14 @@ export const ${ReactComponentName} = createComponent({
   tagName: "${elementName}",
   elementClass: Modules.${ReactComponentName},
   react: React,
+  toAttributeValue: (propValue) => {
+    if (typeof propValue === 'boolean') return propValue ? '' : undefined;
+    if (typeof propValue === 'function') return undefined;
+    const isPrimitive = (v) => typeof v === 'string' || typeof v === 'number';
```
cjpillsbury, line 44: nitpick(blocking): minimally add `boolean` here as well

cjpillsbury (approved):
pre-approved with one nitpick

PR #1142, opened 2025-05-15
feat: use getTemplateHTML (React SSR support)

### LINKED ISSUE #1143: React wrappers don't have `suppressHydrationWarning` by default anymore
Got removed here https://github.com/muxinc/media-chrome/pull/1068

Probably want to put those back because many Media Chrome elements "sprout" attributes when upgraded. 
Things like `role`, `tabindex`, `mediaxstate`...

luwes:
Implementation of https://github.com/muxinc/media-chrome/discussions/1127
fix #1143 
Test page: https://media-chrome-demo-nextjs-git-gettemplatehtml-mux.vercel.app/react-wrappers

This change introduces a new static method called `getTemplateHTML(attrs, props)` on all the Media Chrome elements.
The pattern it replaces is roughly:

```js
template.innerHTML = `<style>...</style><slot></slot>`;
...
this.shadowRoot.append(template.content.cloneNode(true));
```

with

```js
static getTemplateHTML(attrs, props) {
	return `<style>...</style><slot></slot>`;
}
...
const attrs = namedNodeMapToObject(this.attributes);
this.shadowRoot.innerHTML = this.constructor.getTemplateHTML(attrs, {});
```

This change makes it possible to easily create a declarative shadow dom (DSD) on the server with just the classes of the custom elements. See https://github.com/muxinc/ce-la-react/blob/main/src/ce-la-react.ts#L305-L323

The result is that our React wrappers can now server side render the internal HTML of Media Chrome's elements speeding up rendering of the controls.

Looking at the code above you might have noticed the template is not cloned any longer. That is a small performance hit in JS / DOM processing if the app were to re-create the controls over and over but I'm quite sure in real world it is negligible. And we could easily add template caching if needed.


src/js/media-chrome-button.ts
```diff
@@ -149,39 +168,18 @@ class MediaChromeButton extends globalThis.HTMLElement {
...
-        this.tooltipContent = options.tooltipContent;
-      }
-
-      this.nativeEl.appendChild(slotTemplate.content.cloneNode(true));
+      const attrs = namedNodeMapToObject(this.attributes);
 
-      this.shadowRoot.appendChild(buttonHTML);
+      this.shadowRoot.innerHTML = /*html*/ `
+        ${(this.constructor as typeof MediaChromeButton).getTemplateHTML(attrs)}
+      `;
```
luwes, line 182: see how this got cleaned up ^^

src/js/media-chrome-button.ts
```diff
@@ -149,39 +171,18 @@ class MediaChromeButton extends globalThis.HTMLElement {
...
-      }
+      this.attachShadow((this.constructor as typeof MediaChromeButton).shadowRootOptions);
 
-      this.nativeEl.appendChild(slotTemplate.content.cloneNode(true));
+      const attrs = namedNodeMapToObject(this.attributes);
 
-      this.shadowRoot.appendChild(buttonHTML);
+      this.shadowRoot.innerHTML = /*html*/ `
+        ${(this.constructor as typeof MediaChromeButton).getTemplateHTML(attrs)}
+      `;
```
luwes, line 185: look how this got cleaned up ^^

src/js/media-chrome-button.ts
```diff
@@ -4,105 +4,124 @@ import {
...
+
+    <slot name="tooltip">
+      <media-tooltip part="tooltip" aria-hidden="true">
+        <template shadowrootmode="${MediaTooltip.shadowRootOptions.mode}">
+          ${MediaTooltip.getTemplateHTML({})}
+        </template>
+        <slot name="tooltip-content">
+          ${this.getTooltipContentHTML(_attrs)}
+        </slot>
+      </media-tooltip>
```
luwes, line 111: @cjpillsbury nested example

mmcc:
@claude what do you think of this pr (I have no idea what I'm doing and trying to just test out the new action)

Issue #1126, opened 2025-04-23
Programmatically set translation dictionary


raphaelokon:
Hi there.

I noticed that when the `media-controller` is initialized before a given language (TranslateDictionary) is available or dynamically changed, e.g. when calling `setLanguage` and `setTranslation` in a wrapping component’s  `connectedCallback`, the labels are not displayed in the newly set language. This is unfortunate as our wrapping component does not know which language will be used and cannot guarantee an import of the language module prior iinitialization.

Is there a way to (re)set the translation dictionary dynamically via an API instead on relying on import order?

Cheers

ronald-urbina:
Hi @raphaelokon, Thanks so much for the suggestion to improve our translation system. We really appreciate the feedback!

We’re happy to share that in the next release of Media Chrome, language files can now be imported at any time. Simply set the lang attribute on the <media-controller> to your desired language code (e.g. lang="es"), and the UI will update automatically to reflect the new language.

Let us know if you run into any issues or have further suggestions.

PR #1152, opened 2025-06-11
feat: dynamic label update on lang attribute change i18n

ronalduQualabs:
## Problem

Currently, components render aria-label and tooltip text only once during construction. If a new language is added or the lang attribute is set on the after components are initialized, these labels do not update. The current implementation only supports loading additional languages before the initial render, limiting support for dynamic internationalization.

## Solution
- Added a mutation observer to detect changes to the lang attribute on the state mediator
- add the langmedia attr to buttons 
- Triggered updates to aria-label and tooltip values when the language changes.
- Enabled support for adding new languages after the media controller is created, ensuring components re-evaluate their labels dynamically.

This resolves https://github.com/muxinc/media-chrome/issues/1126 and improves support for dynamic i18n workflows.

**Usage**

HTML

Option 1: Auto-detect browser language
	•	Import the language file before loading Media Chrome.
	•	Media Chrome will detect the browser’s language and apply the matching loaded language automatically.

Option 2: Manually specify language
	•	Import the desired language file before loading Media Chrome.
	•	Set the lang attribute on <media-controller> to specify which language to use.

Option 3: Load language dynamically
	•	Dynamically load the language file at runtime (e.g., with a <script> tag or dynamic import).
	•	Set the lang attribute on <media-controller> to match the loaded language.

React
	•	Import the language file directly in your component or page.
	•	Set the language via the lang attribute on <media-controller>, either statically or dynamically in your component logic.


cjpillsbury:
@ronalduQualabs Oh one place where this will plausibly break is when using the media store server side. You should be able to test that locally using the nextjs demo and testing the different pages there.

ronald-urbina:
> Haven't deep dived into smoke testing/permutations, but this feels like a reasonable solution to me without a larger architectural refactor (which I think we should do longer term, but I think it's fair to treat as out of scope for this effort). Defer to @luwes for final approval.

I added some usage instructions in the description to help with testing.
This has been tested in both Next.js example and plain HTML, and it works great in both cases.

src/js/media-store/media-store.ts
```diff
@@ -527,7 +527,7 @@ export const createMediaStore = ({
           // responsible for managing/monitoring `defaultSubtitles` in the `defaultStateMediator`. (CJP)
           stateOwners.options[optionName] = optionValue;
         });
-        // updateStateFromFacade();
+        updateStateFromFacade();
```
luwes, line 530: why was this needed?

luwes (changes requested):
Remove the `lang` property and refactor the buttons to change the tooltip contents via the already existing getTooltipContentHTML method in media-chrome-button. Saves bytes and prevents the repeating of the labels which could become a source of bugs which is the case with proper casing in labels e.g. pip button

src/js/media-store/media-store.ts
cjpillsbury: Yeah I think we need to take a step back here. Per out of band convo with @ronalduQualabs, this impacts other state change behaviors in the media store and also explicitly contradicts the code comment above. Will try to help with a strategy/rework here.

TL;DR answer to your question, @luwes - `mediaLang` is modeled as an "option". Options aren't designed to have "live" dynamic updates in the way that other modeled state is. We want `mediaLang` to have "live" dynamic updates. This does that (but is probably a "left turn instead of a right" in the solution).

src/js/media-chrome-button.ts
```diff
@@ -266,6 +270,9 @@ class MediaChromeButton extends globalThis.HTMLElement {
     ) {
       this.tooltipEl.placement = newValue;
     }
+    else if (attrName === MediaUIAttributes.MEDIA_LANG) {
+      this.shadowRoot.querySelector('slot[name="tooltip-content"]').innerHTML = (this.constructor as typeof MediaChromeButton).getTooltipContentHTML();
+    }
```
luwes, line 275: 👍

src/js/media-live-button.ts
```diff
@@ -95,6 +99,7 @@ class MediaLiveButton extends MediaChromeButton {
   ): void {
     super.attributeChangedCallback(attrName, oldValue, newValue);
     updateAriaAttributes(this);
+    console.log("chang woo")
```
luwes, line 102: remove console log

src/js/media-play-button.ts
```diff
@@ -40,9 +40,14 @@ function getTooltipContentHTML() {
...
 
-const updateAriaLabel = (el: MediaPlayButton) => {
+const updateAriaLabelTooltip = (el: MediaPlayButton) => {
   const label = el.mediaPaused ? t('play') : t('pause');
   el.setAttribute('aria-label', label);
+
+  const playSlot = el.shadowRoot?.querySelector('slot[name="tooltip-play"]');
+  const pauseSlot = el.shadowRoot?.querySelector('slot[name="tooltip-pause"]');
+  if (playSlot) playSlot.textContent = t('Play');
+  if (pauseSlot) pauseSlot.textContent = t('Pause');
```
luwes, line 50: could this be removed?

src/js/media-preview-chapter-display.ts
```diff
@@ -30,13 +32,19 @@ class MediaPreviewChapterDisplay extends MediaTextDisplay {
...
       // Only update if it changed, preview events are called a few times per second.
       if (newValue !== oldValue && newValue != null) {
         this.#slot.textContent = newValue;
 
         if (newValue !== '') {
-          this.setAttribute('aria-valuetext', `chapter: ${newValue}`);
+          const ariaValueText = t('chapter', {
+            chapterName: newValue,
+          });
+          this.setAttribute('aria-valuetext', ariaValueText);
```
luwes, line 47: great you added this one! 
the text should be `'chapter: {chapterName}'` I think

src/js/menu/media-audio-track-menu-button.ts
```diff
@@ -50,6 +56,18 @@ class MediaAudioTrackMenuButton extends MediaChromeMenuButton {
     this.setAttribute('aria-label', t('Audio'));
```
luwes, line 56: use the `updateAriaLabel` here then too?

src/js/media-live-button.ts
```diff
@@ -80,6 +83,7 @@ class MediaLiveButton extends MediaChromeButton {
       ...super.observedAttributes,
       MEDIA_TIME_IS_LIVE,
       MEDIA_PAUSED,
+      MediaUIAttributes.MEDIA_LANG
```
luwes, line 86: the constants above this line are destructured above, might as well do the same for `MediaUIAttributes.MEDIA_LANG`

src/js/media-preview-chapter-display.ts
ronalduQualabs: done

src/js/menu/media-audio-track-menu-button.ts
ronalduQualabs: done

src/js/media-play-button.ts
ronalduQualabs: done

src/js/menu/media-captions-menu-button.ts
```diff
@@ -48,6 +48,7 @@ function getTooltipContentHTML() {
 
 const updateAriaChecked = (el: HTMLElement): void => {
   el.setAttribute('aria-checked', areSubsOn(el).toString());
+  el.setAttribute('aria-label', t('closed captions'));
```
luwes, line 51: this shouldn't go in here I think

src/js/menu/media-captions-menu-button.ts
```diff
@@ -74,8 +75,6 @@ class MediaCaptionsMenuButton extends MediaChromeMenuButton {
 
   connectedCallback(): void {
     super.connectedCallback();
-
-    this.setAttribute('aria-label', t('closed captions'));
```
luwes, line 78: does this need to be translated on lang change?

PR #1166, opened 2025-07-16
feat: Add keyboard arrow controls for volume when not on the volume range

### LINKED ISSUE #957: Add keyboard arrow controls for volume when not on the volume range
At first I thought we might add arrow controls to the mute button to control volume, because it felt odd that didn't work, but looking at the youtube player it looks like they just capture any up and down arrow usage as volume control.

I think we should do something similar. We already do this for left and right arrows, skipping ahead and back.

We should also make the left and right arrows work for volume control when the volume range is specifically selected. This is a range-level detail. The native HTML range will respond to either up/down or left/right arrows. The media chrome range only responds to up/down for some reason. I'm sure there's a reason we'll need to dig up.

Pintorado:
Closes #957 

## Changes
- Only prevent default when not focused on a `media-volume-range` element - this allows using `ArrowLeft` & `ArrowRight` keys to control volume when volume range is focused
- Added `ArrowUp` and `ArrowDown` cases to control volume without the volume range being focused

## Behavior
- Up & Down arrows work for volume control when the volume range is NOT selected (as we do for Left & Right arrows for the time range)
- Left & Right arrows work for volume control when the volume range is specifically selected

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+# [4.14.0](https://github.com/muxinc/media-chrome/compare/v4.13.1...v4.14.0) (2025-10-01)
+
+
+### Features
+
+* support multiple bitrates in rendition menu ([#1209](https://github.com/muxinc/media-chrome/issues/1209)) ([3c2ca97](https://github.com/muxinc/media-chrome/commit/3c2ca9755d337523c1501150e2f2bcc989a507c4))
+
+
+
 ## [4.13.1](https://github.com/muxinc/media-chrome/compare/v4.13.0...v4.13.1) (2025-09-03)
 
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "media-chrome",
-  "version": "4.13.1",
+  "version": "4.14.0",
   "lockfileVersion": 3,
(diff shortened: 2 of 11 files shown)


cjpillsbury (changes requested):
This is a good first pass without architectural context, but we'll want to fit this into our keyboard shortcut handler architecture, found here: 
https://github.com/muxinc/media-chrome/blob/main/src/js/media-controller.ts#L578-L691
This also needs to be appropriately accounted for in https://www.media-chrome.org/docs/en/keyboard-shortcuts#article
Docs will also need to be updated accordingly.
This may not be an exhaustive list (cc @luwes)

Pintorado:
Updated docs and tested. I used the `hotkeys` property to add `noarrowup` or `noarrowdown` and it disables them correctly. The `nohotkeys` property didn't work but it doesn't work in main either, there's [this issue](https://github.com/muxinc/elements/issues/879) about `nohotkeys` not working that hasn't been solved yet (in fact today I was taking a look at it so I can continue investigating this bug)

Not sure if it's related though, since it's from elements, but it's weird that both `nohotkeys` properties aren't working.

spuppo-mux:
Tested this with and without the changes and looks good! 👍

Issue #187, opened 2022-03-16
autohide broken with new poster layer


gkatsev:
Going to https://media-chrome.mux.dev/examples/basic.html, clicking play, and then hovering over the video should auto hide after 2 seconds. However, it does not.
If I set `display:none` n the poster-layer, this behavior is returned.

user-inactive only gets set when mousing out of the player area instead of also after a timeout.

gkatsev:
This is also related to #116 and once this is fixed potentially can close out #116.

cjpillsbury:
Fixed by #191

PR #1168, opened 2025-07-21
fix:: switch "template" attribute to string to resolve React and Next.js type issues

ronald-urbina:
This PR addresses the following issues:

- **[Issue #187](https://github.com/muxinc/player.style/issues/187)**:

Updated the "template" attribute to accept only a string (template ID or URL), resolving the actual HTMLTemplateElement internally. This avoids type errors in React and Next.js when using <media-theme template="...">.

### DIFF
--- a/src/js/media-theme-element.ts
+++ b/src/js/media-theme-element.ts
@@ -135,7 +135,9 @@ export class MediaThemeElement extends globalThis.HTMLElement {
   }
 
   get template(): string | HTMLTemplateElement | null {
-    return this.#template ?? (this.constructor as typeof MediaThemeElement).template;
+    return (
+      this.#template ?? (this.constructor as typeof MediaThemeElement).template
+    );
   }
 
   set template(value: string | HTMLTemplateElement | null) {
@@ -150,8 +152,8 @@ export class MediaThemeElement extends globalThis.HTMLElement {
       this.#template = value;
       this.#prevTemplateId = null;
       this.createRenderer();
+    }
   }
-}
 
   get props() {
     const observedAttributes = [
@@ -198,8 +200,6 @@ export class MediaThemeElement extends globalThis.HTMLElement {
   }
 
   #updateTemplate(): void {
-    if (this.#template) return;
-
     const templateId = this.getAttribute('template');
     if (!templateId || templateId === this.#prevTemplateId) return;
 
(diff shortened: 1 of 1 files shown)


luwes (changes requested):
the template prop has to support strings IN ADDITION to the other types.

these are valid too
`this.#template ?? (this.constructor as typeof MediaThemeElement).template;`

src/js/media-theme-element.ts
```diff
@@ -134,15 +134,24 @@ export class MediaThemeElement extends globalThis.HTMLElement {
     return this.renderRoot.querySelector('media-controller');
   }
 
-  get template() {
-    return this.#template ?? (this.constructor as typeof MediaThemeElement).template;
+  get template(): string | HTMLTemplateElement | null {
+    return this.getAttribute('template');
```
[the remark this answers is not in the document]
ronalduQualabs: reverted

src/js/media-theme-element.ts
```diff
@@ -134,15 +134,24 @@ export class MediaThemeElement extends globalThis.HTMLElement {
     return this.renderRoot.querySelector('media-controller');
   }
 
-  get template() {
+  get template(): string | HTMLTemplateElement | null {
     return this.#template ?? (this.constructor as typeof MediaThemeElement).template;
```
luwes, line 138:
```suggestion
    return this.getAttribute('template') ?? this.#template ?? (this.constructor as typeof MediaThemeElement).template;
```

src/js/media-theme-element.ts
```diff
@@ -189,26 +198,26 @@ export class MediaThemeElement extends globalThis.HTMLElement {
   }
 
   #updateTemplate(): void {
+    if (this.#template) return;
```
luwes, line 201: why is this added? 
what if one first sets a node via template prop and after it sets a string?

src/js/media-theme-element.ts
```diff
@@ -222,11 +231,11 @@ export class MediaThemeElement extends globalThis.HTMLElement {
...
 
   createRenderer(): void {
-    if (this.template && this.template !== this.#prevTemplate) {
-      this.#prevTemplate = this.template;
+    if (this.#template && this.#template !== this.#prevTemplate) {
+      this.#prevTemplate = this.#template;
 
       this.renderer = new TemplateInstance(
-        this.template,
+        this.#template,
```
luwes, line 238: this needs to be reverted because it might access 
`(this.constructor as typeof MediaThemeElement).template;`

luwes (approved):
LGTM 👍

PR #1170, opened 2025-07-22
Fix: `nohotkeys` prop disables all shortcuts

### LINKED ISSUE #1169: `nohotkeys` prop does not turn off shortcuts
As it says in the docs, adding this prop to the media controller should turn off _all_ keyboard shortcuts, but it isn't working.

Pintorado:
Closes #1169 

## Changes
Conditionally enable or disable hotkeys based on the presence of the `NO_HOTKEYS` attribute.


luwes (approved):
LGTM 👍

Issue #1100, opened 2025-03-21
Ability to amend the transition from full settings menu to a sub-menu


SteGreig:
Hey 👋  I am trying to amend the transition from the full settings menu to the individual speed/quality menus.

i.e. from this:
<img width="276" alt="Image" src="https://github.com/user-attachments/assets/9143aae4-3611-428d-97cf-235a18f297c1" />

to this:
<img width="264" alt="Image" src="https://github.com/user-attachments/assets/53b5f613-b1c6-4a8a-9e42-88a2e4af20ad" />

I can see the way this works is there is an element within the shadow DOM `<div id="container" class="has-expanded"` which has the following CSS applied:

```
#container {
    transition: transform .2s ease-out;
    transform: translate(0, 0);
}

#container.has-expanded {
    transition: transform .2s ease-in;
    transform: translate(-100%, 0);
}
```

<img width="504" alt="Image" src="https://github.com/user-attachments/assets/037690ce-0c6f-4daa-b642-b225dfbb1126" />

I think I'm right in saying there is currently no way to amend the transition (or any of the styles) on this `#container` div? I believe the only way to target it would be if it had a `part` attribute defined - is that right? Alternatively, you could add a CSS variable for the transition, but the `part` approach would allow for greater customisation.

Is this something that would be easy to add?

And to speak more generally, I have found some aspects of the media player to be difficult to style - is there a reason not to add a `part` attribute to basically all of the markup in the shadow DOM so that it can be targeted for styling if needed?

Thanks!

ronald-urbina:
Hi @SteGreig Thanks for bringing this up! I’m happy to say that support for customizing the transition (e.g., via part attributes) has been implemented. You can expect this functionality to be available in the next release of media-chrome.

PR #1172, opened 2025-07-29
feat:: expose <div> container via part so host DOM styles can target it

ronald-urbina:
This PR adds the part="container" attribute to the internal container <div> in `media-chrome-menu`, allowing users to style it from the light DOM using the ::part(container) selector. This enables better theme customization and external control over layout and appearance.

Fixes [#1100](https://github.com/muxinc/media-chrome/issues/1100).

Usage: 

`media-settings-menu::part(container){
   // css
}`

### DIFF
--- a/src/js/menu/media-chrome-menu.ts
+++ b/src/js/menu/media-chrome-menu.ts
@@ -136,7 +136,7 @@ function getTemplateHTML(_attrs: Record<string, string>) {
         right: -100%;
       }
 
-      :host::part(container) {
+      #container {
         display: flex;
         flex-direction: column;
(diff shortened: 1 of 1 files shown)


src/js/menu/media-chrome-menu.ts
```diff
@@ -136,15 +136,15 @@ function getTemplateHTML(_attrs: Record<string, string>) {
         right: -100%;
       }
 
-      #container {
+      :host::part(container) {
```
luwes, line 139: this change is not needed I think

luwes (approved):

Issue #1173, opened 2025-07-29
Tap on iOS Play/Pause video


ronalduQualabs:
On iOS the Tap event over the video element cause play/pause the media, different from android/mobile expected that should be show the controls.

ronald-urbina:
Solved in [Pull 1174](https://github.com/muxinc/media-chrome/pull/1174)

PR #1174, opened 2025-07-30
Ensure touch taps on iOS are handled correctly as pointerType "touch"

ronalduQualabs:
Problem:
On iOS, a tap triggers a click event that’s interpreted as coming from a "mouse". Our existing click handler treats that as a “mouse” event and play/pause media.

What I did:
Stored the real event.pointerType during the `pointerdown` event (which correctly reports "touch" on iOS), and then referenced that stored value in the click handler.

Now touch taps on iOS do not trigger play/pause, while standard mouse clicks continue working as play/pause.

This solves [1173](https://github.com/muxinc/media-chrome/issues/1173)

### DIFF
--- a/src/js/media-gesture-receiver.ts
+++ b/src/js/media-gesture-receiver.ts
@@ -48,6 +48,7 @@ class MediaGestureReceiver extends globalThis.HTMLElement {
   }
 
   _pointerType: string;
+  _clearPointerTypeTimeout: any | undefined;
 
   constructor() {
     super();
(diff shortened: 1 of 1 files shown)


src/js/media-gesture-receiver.ts
```diff
@@ -112,6 +113,11 @@ class MediaGestureReceiver extends globalThis.HTMLElement {
       // Since not all browsers have updated to be spec compliant, where 'click' events should be PointerEvents,
       // we can use use 'pointerdown' to reliably determine the pointer type. (CJP).
       this._pointerType = event.pointerType;
+      // Optional timeout clear — in case click is delayed
+      clearTimeout(this._clearPointerTypeTimeout);
+      this._clearPointerTypeTimeout = setTimeout(() => {
+        this._pointerType = undefined;
+      }, 500);
```
luwes, line 120:
```suggestion
      // Optional timeout clear — in case click is delayed
      clearTimeout(this._clearPointerTypeTimeout);
      this._clearPointerTypeTimeout = setTimeout(() => {
        this._pointerType = undefined;
      }, 500);
```

this is not needed I think. it's also too random

src/js/media-gesture-receiver.ts
```diff
@@ -48,6 +48,7 @@ class MediaGestureReceiver extends globalThis.HTMLElement {
   }
 
   _pointerType: string;
+  _clearPointerTypeTimeout: any | undefined;
```
luwes, line 51:
```suggestion
```

PR #1195, opened 2025-08-27
Feat: Prevent storing local storage values if not used

### LINKED ISSUE #1194: Feat: Prevent updating localStorage when it's not considered
When adding the `muted` attribute to the media, the `media-chrome-pref-muted` value still gets updated in localStorage when interacting with the mute state.

This behavior is not very intuitive, since the stored value is not actually taken into account when `muted` is present (as the attribute overrides any value from localStorage).

Pintorado:
Closes #1194 

This PR prevents storing local storage values in the following cases:
- When the `muted` attribute is present on the media element
- When the `nomutedpref` attribute is present on the media-controller
- When the `novolumepref` attribute is present on the media-controller

## Changes
- Added `nomutedpref` attribute to the media controller
- Prevent storing muted preference if `muted` or `nomutedpref` are present
- Prevent storing volume preference if `muted` or `novolumepref` are present

### DIFF
--- a/src/js/media-controller.ts
+++ b/src/js/media-controller.ts
@@ -59,6 +59,7 @@ export const Attributes = {
   NO_AUTO_SEEK_TO_LIVE: 'noautoseektolive',
   NO_HOTKEYS: 'nohotkeys',
   NO_VOLUME_PREF: 'novolumepref',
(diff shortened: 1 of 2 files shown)


luwes (approved):
LGTM!

Issue #1180, opened 2025-08-13
Thumb in time range not smoth when fast forward.


tymon3568:
When using the mouse to fast forward the video, the dot (thumb) is not smooth, it does not stay at the forward position, but jerks back to the playing position and then jumps to the forward position.
In youtube, vidstack, plyr,... the thumb ( i mean the dot) move smoothy.
Hope team improve this.

tymon3568:
- Device Specifications: archcraft xfce
- Browser Information: chrome newest version 
- Sample: https://youtu.be/1K1Pz1Avqv8 ( in this video you will see the difference between media chrome and plyr as i thumb forward and backward the video.  The dot (thumb) is not smooth, it does not stay at the forward position, but jerks back to the playing position and then jumps to the forward position)

ronalduQualabs:
Hi @tymon3568 Thanks for the feedback and for taking the time to send the video! We’ve worked on a fix, and the solution will be included in the next Media Chrome release in the coming days.

tymon3568:
Thank you @ronalduQualabs

PR #1204, opened 2025-09-03
Fix race condition between input and pointermove in <media-time-range>

ronald-urbina:
Resolve https://github.com/muxinc/media-chrome/issues/1180

This PR resolves a race condition in <media-time-range> where the `input` event would trigger, but a `pointermove` update could fire immediately afterward and override the expected seek position. This caused the thumb to glitch back before settling at the clicked position.

### Changes
- Introduced `#isPointerDown` flag to prevent `updateBar` from running during the conflict window.
- Updated `handleEvent` to correctly manage the suppression logic when `input` and `pointermove` events overlap.
- Ensures the `input` event’s seek update always wins, preventing visual glitches.

### Result
Clicking or dragging the timeline now updates smoothly without the thumb jumping back briefly.

### DIFF
--- a/src/js/media-time-range.ts
+++ b/src/js/media-time-range.ts
@@ -429,7 +429,7 @@ class MediaTimeRange extends MediaChromeRange {
   #boxPaddingLeft: number;
   #boxPaddingRight: number;
   #mediaChaptersCues;
-  #isSeekInProgress: boolean;
+  #isPointerDown: boolean;
 
   constructor() {
     super();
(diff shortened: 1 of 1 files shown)


luwes (approved):
seems to work! the naming is a bit misleading because the seeking can still be happening after pointer up but not a blocker

ronald-urbina:
> seems to work! the naming is a bit misleading because the seeking can still be happening after pointer up but not a blocker

@luwes I updated name into #isPointerDown

PR #1209, opened 2025-09-30
feat: support multiple bitrates in rendition menu

mihar-22:
This PR adds support for multiple bitrates in the rendition menu by:

1. Showing the bitrate when there are multiple renditions with the same height.
2. Sorting renditions based on both height and bitrate (if any have the same height).

---

The way we track the current rendition needs to be revised to support auto and still knowing accurately which rendition is selected. They're conflated under `mediaRenditionSelected` which can only be `auto` or a rendition. 

## 🖼️ Screenshots

<details>
  <summary>Settings menu item without bitrate</summary>
   
CleanShot 2025-09-30 at 21 41 08@2x

</details>

<details>
  <summary>Settings menu item with bitrate</summary>
   
CleanShot 2025-09-30 at 21 04 51@2x

</details>

<details>
  <summary>Rendition menu with bitrates</summary>
  
CleanShot 2025-09-30 at 21 53 23@2x

</details>

### DIFF
--- a/src/js/menu/media-rendition-menu.ts
+++ b/src/js/menu/media-rendition-menu.ts
@@ -14,7 +14,12 @@ import {
   createIndicator,
 } from './media-chrome-menu.js';
 import { Rendition } from '../media-store/state-mediator.js';
-import {t} from '../utils/i18n.js';
+import { t } from '../utils/i18n.js';
+
+export type FormatRenditionOptions = {
+  showBitrate?: boolean;
+};
+
 /**
  * @extends {MediaChromeMenu}
  *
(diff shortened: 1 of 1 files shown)


src/js/menu/media-rendition-menu.ts
```diff
@@ -36,6 +41,37 @@ class MediaRenditionMenu extends MediaChromeMenu {
...
+    }
+
+    return this.formatMenuItemText(renditionText, rendition);
+  }
+
+  static compareRendition(a: Rendition, b: Rendition): number {
+    return b.height === a.height
+      ? (b.bitrate ?? 0) - (a.bitrate ?? 0)
+      : b.height - a.height;
+  }
```
mihar-22, line 73: Formats in descending order by height then bitrate. It can be overridden for custom sorting.

PR #1214, opened 2025-10-10
fix: Use CSS variable for focus shadow in media button

jarred:
Use CSS variable (if present) for box-shadow


ronald-urbina:
@jarred Thanks so much for this small improvement and for taking the time to collaborate! Really appreciate it.

Issue #1228, opened 2025-10-30
defaultsubtitles always forces captions ON in Safari, even when disabled


mudit2108:
Hi team 👋,

I noticed an issue with the defaultsubtitles parameter in media-chrome.

**Issue**

When defaultsubtitles=false is used, captions/subtitles appropriately turn off by default in Chromium-based browsers when the playback starts.

However, on Safari (observed on some versions), subtitles remain on even if defaultsubtitles=false. On next playback/load, subtitles automatically turn on again.

Expected Behavior

With defaultsubtitles=false → subtitles disabled on load

I have tested on multiple safari versions it works fine on some. 
Safari 26.0.1 is one which has the issue

<img width="2662" height="1468" alt="Image" src="https://github.com/user-attachments/assets/f5f9c486-3ff0-4a52-9050-3117fa2f2ca3" />

ignasio-mux:
Thanks for the report! One quick question to help us narrow this down: does the captions track you’re using have the default attribute set on it?

mudit2108:
hey @ignasio-mux 

Sorry for the late response on this, no we don't set the default attribute on the track

ignasio-mux:
Hi @mudit2108, sorry for the delay. Thanks for the clarification. One more question, are you using the React component MediaController or the web component media-controller?

ignasio-mux:
Hi @mudit2108, just following up on this.
If you’re using the React MediaController component, this issue should now be resolved in the latest release.

Whenever you get a chance, let us know if updating fixes the problem on your side.

mudit2108:
Hey @ignasio-mux 

Apologies on the late ping, we are using the React MediaController component. 
Let me try updating the library and report back.

Thanks for taking it up.

ignasio-mux:
Hi @mudit2108, just checking in.
Were you able to reproduce this in the latest release?

ignasio-mux:
Hi @mudit2108, have you had a chance to look at this?
If we don’t hear back soon, we’ll go ahead and close the issue, but please feel free to reopen it anytime with more details.

PR #1230, opened 2025-11-05
fix defaultsubtitles value issue

akbansa:
Fix for #1228


luwes:
`false` string attribute values are not really the way to do it in html. The absence of an attribute is a false value.

PR #1231, opened 2025-11-05
feat: Add support for hotkeys

### LINKED ISSUE #1171: Add support for hotkeys
Add support for additional support keys, similar to YouTube.

Additional keys:
Volume up: `↑`
Volume up: `↓`
Seek Forward: `l`
Seek Backward: `j`
Playback speed up: `>`
Playback speed down: `<`
Show shortcut overlay: `Shift + /`

It will make the player more accessible and familiar to users.

Pintorado:
Closes #1171 

### Added hotkeys:
- `l` : Seek forward 10s
- `j` : Seek backward 10s
- `>` : Increase playback rate
- `<` : Decrease playback rate
- `Shift + /` : Show keyboard shortcuts

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+## [4.15.1](https://github.com/muxinc/media-chrome/compare/v4.15.0...v4.15.1) (2025-10-30)
+
+
+### Bug Fixes
+
+* muted pref covers border cases ([#1224](https://github.com/muxinc/media-chrome/issues/1224)) ([6366f72](https://github.com/muxinc/media-chrome/commit/6366f72babf076adefc8feb8c6a20d8d5717ba7e)), closes [#1226](https://github.com/muxinc/media-chrome/issues/1226)
+
+
+
 # [4.15.0](https://github.com/muxinc/media-chrome/compare/v4.14.0...v4.15.0) (2025-10-27)
 
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "media-chrome",
-  "version": "4.15.0",
+  "version": "4.15.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "media-chrome",
-      "version": "4.15.0",
+      "version": "4.15.1",
       "license": "MIT",
       "dependencies": {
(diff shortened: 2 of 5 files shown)


src/js/media-controller.ts
```diff
@@ -46,30 +47,37 @@ const ButtonPressedKeys = [
...
   NO_SUBTITLES_LANG_PREF: 'nosubtitleslangpref',
-  NO_DEFAULT_STORE: 'nodefaultstore',
-  KEYBOARD_FORWARD_SEEK_OFFSET: 'keyboardforwardseekoffset',
-  KEYBOARD_BACKWARD_SEEK_OFFSET: 'keyboardbackwardseekoffset',
-  KEYBOARD_UP_VOLUME_STEP: 'keyboardupvolumestep',
-  KEYBOARD_DOWN_VOLUME_STEP: 'keyboarddownvolumestep',
-  LANG: 'lang',
+  NO_VOLUME_PREF: 'novolumepref',
+  SEEK_TO_LIVE_OFFSET: 'seektoliveoffset',
 };
```
Pintorado: Don't pay attention to this, I just reordered the attributes

spuppo-mux:
Adding a screenshot example for future reference.

image

spuppo-mux:
Just as an opinion, since I'm not a UX expert, it may be good to add another entry saying something like `[Esc] Close this menu`, also seems intuitive to be able to close the menu also with `shift + /`.

src/js/media-controller.ts
```diff
@@ -759,10 +788,71 @@ class MediaController extends MediaContainer {
...
+          detail,
+        });
+        this.dispatchEvent(evt);
+        break;
+      }
+
+      case '>': {
+        const playbackRate = this.mediaStore.getState().mediaPlaybackRate ?? 1;
+        detail = Math.min(
+          playbackRate + DEFAULT_PLAYBACK_RATE_STEP,
```
spuppo-mux: We should use .toFixed(2) here to avoid precision errors.
image
I think this happens particularly because clicking the button takes a 0.2 step, so then when adding with the '>' (and because of JS precision errors) we get weird numbers.
image

spuppo-mux: Also, if this is a settable attribute, we should get it here

PR #1232, opened 2025-11-08
[a11y and i18n] Translate time phrases

manuhabitela:
:wave: Hey there,

Thanks for media-chrome.

This is a tiny suggestion to improve the experience of non-English speaking users of assistive technologies like screen readers.

The duration labels meant for assistive tech were not translated, rendering for example "1 hour remaining" when using the French locale. Now all necessary strings are passed through the i18n translation function.

I thought of using [Intl.DurationFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DurationFormat) but since browser support is very recent I figured it was not that great of an idea for now.

**:warning: only the English and French strings are legit. I can't vouch for other locales where AI did its thing.**

Sorry if this is not complete enough (like maybe missing tests or anything else), I'm just discovering the code.

Cheers!

### DIFF
--- a/src/js/lang/en.ts
+++ b/src/js/lang/en.ts
@@ -65,6 +65,15 @@ export const En = {
     'An unsupported error occurred. The server or network failed, or your browser does not support this format.',
   'The media is encrypted and there are no keys to decrypt it.':
     'The media is encrypted and there are no keys to decrypt it.',
+  hour: 'hour',
+  hours: 'hours',
+  minute: 'minute',
+  minutes: 'minutes',
+  second: 'second',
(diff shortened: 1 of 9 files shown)


manuhabitela:
Hey there,

Do I have to do something specific that I missed to get an answer here? :sweat_smile: Thanks

ronald-urbina:
@manuhabitela Thank you very much for taking the time to look into the code and collaborate on this missing part of media-chrome's internationalization. We greatly appreciate it.

manuhabitela:
Thanks for merging!

PR #1235, opened 2025-11-11
feat: expose decrease threshold as an attribute in media-time-range

zivac:
Addresses this issue: https://github.com/muxinc/media-chrome/discussions/1234

### DIFF
--- a/src/js/media-time-range.ts
+++ b/src/js/media-time-range.ts
@@ -515,6 +515,12 @@ class MediaTimeRange extends MediaChromeRange {
       this.mediaChaptersCues = this.#mediaChaptersCues;
       this.updateBar();
     }
+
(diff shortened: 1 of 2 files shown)


luwes:
Thanks for the contribution but I'm hesitant to add this new config option as a public API.
Could this be solved internally in some way? 

Is the video in question very long? Maybe this 0.3 value should change based on the length of the video?

zivac:
The issue is more pronounced on longer videos (i.e. over 40 minutes) but also visible on shorter videos (i.e. about 10 minutes). We solved the problem temporarily with an npm patch that completely removes the 0.3 value. Maybe instead of exposing the number value, you could expose a boolean to enable/disable it?

PR #1261, opened 2026-01-16
fix: Patch Memory Leaks

### LINKED ISSUE #1235: Bug: Mux Player memory not released when unmounted
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Which Mux Elements/Packages does this apply to? Select all that apply

mux-player-react

### Which browsers are you using?

Chrome

### Which operating systems are you using?

macOS

### Description

`<mux-video>` and `<mux-player>` elements are not released from javascript heap when the React MuxPlayer component is unmounted.

### Reduced test case

https://github.com/musatov/mux-video-memoryleak

### Steps to reproduce

1. Open the home page with a couple of links to video's pages
2. Open the dev tools and make s heap snapshot - make sure there is no mux-video nor mux-player objects in the heap
3. Navigate to the single video page
4. Navigate back to the home page
5. Force garbage collection
6. Make snapshot again
7. There is mux-video and mux-player objects in the detached state in the heap.

https://github.com/user-attachments/assets/fdece2bc-a26e-4510-a724-39cac75b9ec1

### Current Behavior

The objects related to unmounted react MuxPlayer component are in the heap. 

Image

### Expected Behavior

`<mux-video>` and `<mux-player>` elements are cleaned up.

### Errors

_No response_

### What version of the package are you using?

3.8.0

spuppo-mux:
This PR addresses some memory leak issues across many of media-chrome components. Relates to https://github.com/muxinc/elements/issues/1235

The main idea in these changes is to clean references on component disconnect. It includes changes like removing added event listeners, disconnecting MutationObservers, and making sure we keep reference to the functions passed as callbacks to these types of objects.

It also includes some minor fixes like: adding typing; moving some logic from the components constructor to connectedCallback, for example adding event listeners or reinitializing stuff that is removed on disconnect; adding guards so that listeners are not added more than once.

## Testing
To test this I added the `examples/vanilla/memory-leak-tester.html` file which allows mounting and unmounting a `<media-controller>` element. 

The process consisted on mounting and unmounting the component and taking a heap snapshot to verify no leaks are present. To track memory leaks you can check the element's retainer tree to try and find who is holding the reference. It's highly recommended to use a non minified version of the code when testing. 

## Some extra details
1. Hanging `<video>` tag
When debugging media chrome, even if it's not specifically a component being leaked, we may find a Detached video tag

image

It should be an empty `<video>` tag, which relates to `testMediaEl` used in `src/js/utils/platform-tests.ts` as we can see from it's retainers. It is purposefully maintained in memory to not create several ones when checking some browser compatibilities.

Meaning that a "clean" heap snapshot looks like this:
image

2. Another video tag is held by a console warning about aria-hidden, which in turn holds the media controller:

```
Blocked aria-hidden on an element because its descendant retained focus. The focus must not be hidden from assistive technology users. Avoid using aria-hidden on a focused element or its ancestor. Consider using the inert attribute instead, which will also prevent focus. For more details, see the aria-hidden section of the WAI-ARIA specification at https://w3c.github.io/aria/#aria-hidden.
Element with focus: <video slot=media>
Ancestor with aria-hidden: 
<video slot=media> 
<video slot=​"media" playsinline src=​"https:​/​/​stream.mux.com/​A3VXy02VoUinw01pwyomEO3bHnG4P32xzV7u1j1FSzjNg/​high.mp4" aria-hidden=​"true" tabindex=​"-1">​</video>
```
To remove this I added the `inert` attribute on `media-container.ts:414`, but this makes the element non clickable, which may not be desired. Alternatively, we can remove the aria-hidden attribute, but I understand this was added to address accessibility issues in a previous PR. https://github.com/muxinc/elements/issues/1040
On media-controller:510 (mediaSetCallback), we set media tabIndex to -1, meaning it's not keyboard accessible.

3. Also, when debugging, sometimes Chrome will cache some elements (holding reference to the whole tree of components), which looks like a memory leak. Generally, if `<symbol Window#DocumentCachedAccessor>inWindow (global*) /` appears as one of the first retainers, it means that this is happening.
image
One way to check this is not actually a leak is to mount the component again, play the video, and unmount it. Now on another heap snapshot we should no longer see these elements.

### DIFF
--- a/examples/nextjs-with-typescript/next-env.d.ts
+++ b/examples/nextjs-with-typescript/next-env.d.ts
@@ -1,6 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
-/// <reference types="next/navigation-types/compat/navigation" />
+import "./.next/dev/types/routes.d.ts";
 
 // NOTE: This file should not be edited
 // see https://nextjs.org/docs/app/api-reference/config/typescript for more information.
--- a/examples/nextjs-with-typescript/package-lock.json
+++ b/examples/nextjs-with-typescript/package-lock.json
@@ -14,7 +14,7 @@
         "@mui/material": "^7.0.2",
         "@mux/mux-video": "^0.17.5",
         "media-chrome": "file:../../",
-        "next": "15.3.8",
+        "next": "~16.1.1",
         "react": "19.2.2",
         "react-dom": "19.2.2"
       },
@@ -31,7 +31,7 @@
       }
     },
     "../..": {
-      "version": "4.17.0",
+      "version": "4.17.2",
       "license": "MIT",
       "dependencies": {
         "ce-la-react": "^0.3.2"
@@ -53,6 +53,7 @@
         "react": "19.2.2",
         "resolve.exports": "^2.0.3",
         "rimraf": "^3.0.2",
+        "shx": "^0.4.0",
         "sinon": "^14.0.1",
         "typescript": "^5.5.2",
         "typescript-eslint": "^7.14.1",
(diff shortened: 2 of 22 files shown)


luwes (changes requested):
Thanks for all the fixes. 

Are you sure these are all necessary? 
Could you try to make the least amount of changes that are needed to solve the problem, could be that it's needed. Just double checking.

luwes (approved):
LGTM thank you!

Issue #1307, opened 2026-08-13
Seeking into an unbuffered range leaves time range and time display at the old position until the seek completes


DiFuks:
### Summary

When a seek targets a position that is not buffered yet, `media-time-range` and `media-time-display` keep showing the **pre-seek** position for the whole stall, and only jump to the new position once the seek completes. With a burst of seek requests (e.g. holding/repeating a keyboard hotkey) nothing in the UI moves at all until the last seek finishes, which reads as a frozen player.

### Steps to reproduce

1. Play an HLS VOD (mine: 1:38:30) and let it buffer, e.g. up to 2610 s.
2. While playing, seek to a position past the buffered range — a single `mediaseekrequest` with `detail = 2780`, or repeated `+5 s` hotkey presses.
3. Watch `media-time-range` / `media-time-display` during the buffering stall.

**Expected:** the UI reflects the requested position immediately; the loading indicator communicates the stall.
**Actual:** both stay at 2480.9 s until the seek resolves.

Measured event log (media-chrome 4.19.2, hls.js 1.6.15, Chrome 150, macOS) — columns are `event | media.currentTime | readyState | timeRange bar position (s) | medialoading`:

```
seeking            2780.89  1  2480.9  0
A:medialoading     2780.89  1  2480.9  1
waiting            2780.89  1  2480.9  1
… 824 ms, no state change, bar frozen at 2480.9 …
A:mediacurrenttime 2780.90  4  2780.9  1
timeupdate         2780.90  4  2780.9  1
seeked             2780.90  4  2780.9  0
```

824 ms on a fast desktop connection; on mobile/throttled it lasts as long as the fragment takes to load.

### Cause

`media.currentTime` already returns the new position as soon as the seek starts (visible in the log above at the `seeking` event), but nothing propagates it into the store:

- `state-mediator.js` — `mediaCurrentTime` is only recomputed on `mediaEvents: ["timeupdate", "loadedmetadata"]`, and per spec `timeupdate` is not fired while a seek is pending.
- `request-map.js` — the seek request sets the value but returns nothing, so `updateState(undefined)` is a no-op and there is no optimistic update:

```js
[MediaUIEvents.MEDIA_SEEK_REQUEST](stateMediator, stateOwners, { detail }) {
  const key = "mediaCurrentTime";
  const value = detail;
  stateMediator[key].set(value, stateOwners);
},
```

### Suggested fix

Return the delta so the UI reflects the requested position right away:

```js
[...]

PR #1315, opened 2026-09-12
fix: snap media-time-range when seeking back less than 3% of duration

### LINKED ISSUE #1306: media-time-range does not move backward for seeks smaller than 3% of duration
### Summary

`media-time-range` never moves backward when the seek is smaller than 3% of the media duration. The bar freezes at the pre-seek position and only starts moving again once playback catches up with it, so the freeze lasts exactly as long as the amount that was seeked back.

### Steps to reproduce

1. Play a long VOD (mine: 1:38:30, `duration = 5909.6`).
2. While playing, seek backward by 20 s (arrow-key hotkey, `media-seek-backward-button`, or a `mediaseekrequest` with `detail = currentTime - 20`).
3. Watch `media-time-range`.

**Expected:** the bar jumps back to the new position.
**Actual:** the bar stays where it was for the next 20 s of playback, then resumes normally.

Measured on the page (Chrome 150, VOD via hls.js 1.6, `media-chrome` 4.19.2), seek from 2301 s to 2281 s:

| | value |
|---|---|
| `mediacurrenttime` attribute | 2281.01 → 2281.22 → 2281.49 … (correct) |
| `timeRange.range.valueAsNumber` | 0.38936 (= 2301 s) for 20 s, unchanged |

So state propagation is fine — only the range animation refuses to move.

### Cause

`dist/utils/range-animation.js`:

```js
update({ start, duration, playbackRate }) {
  const increase = start - this.#range.valueAsNumber;
  const durationDelta = Math.abs(duration - this.duration);
  if (increase > 0 || increase < -0.03 || durationDelta >= 0.5) {
    this.callback(start);
  }
  ...
}
```

`start` is normalized to `0..1`, so the `-0.03` guard means **3% of the media duration**, not a fixed amount of time:

| duration | backward seeks that never move the bar |
|---|---|
| 5 min | < 9 s |
| 1 h 38 min | < 177 s |
| 3 h | < 324 s |

When the callback is skipped, `#animate` takes the `increase <= 0` branch and decays `#lastRangeIncrease` by `0.995` per frame, so the bar visually stands still instead of following the media.

### Suggested fix

The guard looks intended to absorb the sub-second lag between `timeupdate` values and the position the animation has already advanced to — that intent is time-based, but the constant is expressed in fractions of duration. Something like:

```js
if (increase > 0 || increase < -0.5 / duration || durationDelta >= 0.5) {
```

keeps the anti-jitter behaviour (0.5 s of tolerance) without scaling it with the media length.

Happy to send a PR if you agree with the approach.

### Possibly related, separate

For seeks into an unbuffered range, the bar and `media-time-display` also stay at the **old** position for the whole buffering stall (measured: ~0.8 s on a fast connection, arbitrarily long on a slow one). `mediaCurrentTime` in the state mediator only updates on `timeupdate`/`loadedmetadata`, and the browser does not fire `timeupdate` until the seek completes, while `MEDIA_SEEK_REQUEST` in `request-map.js` sets `media.currentTime` without returning an optimistic state delta. Returning `{ mediaCurrentTime: value }` there would make the UI reflect the requested position immediately. Let me know if you'd like that as its own issue.

### Environment

- media-chrome 4.19.2 (code is identical in 4.19.3-canary.1)
- Chrome 150, macOS
- hls.js 1.6.15, VOD stream

I searched open and closed issues for this and did not find a duplicate — apologies if I missed one.

cpruijsen:
## Summary

The snap guard is now 0.5 seconds of media time (`increase < -0.5 / duration`), the same scale as the existing `durationDelta >= 0.5` check in that `if`. Sub-second playhead jitter is still ignored.

`media-time-range` feeds the animator a 0..1 position. The old guard was `increase < -0.03`, which is 3% of duration, not 0.03 seconds. A 20 s backward seek on a ~1 h 38 min VOD (`20 / 5909.6 ≈ 0.34%`) never called the range callback, so the thumb stayed put until playback caught up.

Fixes #1306

Reporter's measurement (Chrome 150, media-chrome 4.19.2, seek 2301 to 2281): `mediacurrenttime` advanced correctly while `timeRange.range.valueAsNumber` stayed at `0.38936` for 20 s.

## Decision

The threshold is `increase < -0.5 / duration`, the formula on the issue. The alternative is `Math.max(-0.03, -0.5 / duration)` so short clips keep a 3% jitter floor, or snap on every negative `increase`. `-0.5 / duration` is time-based like the Safari/`timeupdate` lag the guard was meant to absorb, and it reuses the 0.5 s already used for duration hops in the same condition. On a 10 s clip the new threshold ignores a 0.4 s decrease that the old 3% floor would have snapped. Happy to switch to the hybrid or another tolerance.

The unbuffered-seek stall (no optimistic `mediaCurrentTime` from `MEDIA_SEEK_REQUEST`) is #1307 and is not in this change.

## Test plan

- [x] Unit test: 20 s backward seek on duration `5909.6` invokes the range callback (fails with `-0.03`, passes here).
- [x] Unit test: 0.2 s playhead decrease still does not snap.
- [ ] Optional: on a VOD longer than ~15 minutes, seek back 10-20 s while playing and confirm the thumb jumps immediately.

---

> [!NOTE]
> **Low Risk**
> Localized change to seek-bar animation filtering plus new unit tests; no auth, data, or API surface.
> 
> **Overview**
> Fixes the seek bar thumb not jumping when users seek backward on long videos (#1306).
> 
> **`RangeAnimation.update()`** now treats backward playhead moves as “real” when they exceed **0.5 seconds of media time** (`increase < -0.5 / duration`), instead of a fixed **3%** normalized decrease (`-0.03`). Small backward jitter (e.g. 0.2 s) is still ignored, aligned with the existing **0.5 s** `durationDelta` guard in the same condition.
> 
> Adds unit tests for a ~98 min VOD: a **20 s** backward seek must update the range; a **0.2 s** decrease must not.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 66294ac11d525de7cc7c88d30969aaded73147f7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


javierarams:
We'll take a look at this with the team and get back to you soon. Thanks for the fix!

spuppo-mux (approved):
Tested on Chrome and Safari and lgtm! Thanks!

